PR #362, opened 2021-02-24
inject fsgroup in grafana-dev

marcolan018:
https://github.com/open-cluster-management/backlog/issues/9699


anthonyfister:
2.2.1 will only be a container refresh.
Please target this to ACM 2.2.2.

songleo:
we just need to fix it in upstream. and it is not bundled into the release images.

PR #378, opened 2021-03-05
delete invalid manifestwork during reconcile

marcolan018:


controllers/placementrule/placementrule_controller.go
```diff
@@ -151,6 +151,14 @@ func (r *PlacementRuleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		staleAddons = append(staleAddons, addon.Namespace)
 	}
 	for _, work := range workList.Items {
+		if work.Name != work.Namespace+operatorWorkNameSuffix &&
+			work.Name != work.Namespace+resWorkNameSuffix {
```
clyang82, line 155: it seem the logic will remove `klusterlet-*` manifestwork as well

clyang82: I think we should delete the previous endpoint-*-manifestwork prior to creating the new one.

marcolan018: @clyang82 
the changes are not ready and I have not tested it in a upgrade scenario. Seems I created the pr by mistake...
I will request your review later once I completed all works on this.

For your concern/questions above:
the workList is manifestworks with label owner: multicluster-observability-operator, so will not affect other manifestworks.
imo, it doesn't matter to delete old manifestworks before/after new manifestworks created. Since work agent will guarantee the state in managed cluster as same as the new manifestworks finally.

PR #386, opened 2021-03-10
switch to use clusterrole to replace roles in each cluster namespace

marcolan018:
https://github.com/open-cluster-management/backlog/issues/10004


clyang82:
Do we need to special handle the upgrade case - to remove the role in managedcluster namespace?

marcolan018:
@clyang82 
yes, I will add the code to handle the upgrade in another pr

PR #402, opened 2021-03-19
delete deprecated roles/manifestworks in 2.2

marcolan018:
1. clean old manifestworks
2. clean old roles: https://github.com/open-cluster-management/backlog/issues/10004

### DIFF
--- a/controllers/placementrule/manifestwork.go
+++ b/controllers/placementrule/manifestwork.go
@@ -34,6 +34,22 @@ type MetricsAllowlist struct {
 	MatchList []string `yaml:"matches"`
 }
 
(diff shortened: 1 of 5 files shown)


clyang82:
@marcolan018 I did not follow the comment: `imo, it doesn't matter to delete old manifestworks before/after new manifestworks created. Since work agent will guarantee the state in managed cluster as same as the new manifestworks finally`. for instance, upgrade from 2.2, the old manifestwork is there, the endpoint-operator is in managed cluster. if the new manifestwork created, do we have conflicted between old and new?

marcolan018:
@clyang82 
there will be no conflict
the resource with the same name in new manifestwork will not be created until the old manifestwork deleted.

PR #408, opened 2021-03-24
handle  null observabilityaddonspec

marcolan018:
also fix the error when listing deprecated roles

https://github.com/open-cluster-management/backlog/issues/10460


clyang82:
Just let you know. the `observabilityAddonSpec` is required in v1beta2.

PR #425, opened 2021-03-27
add permission for observatorium sa for crud of poddisruptionbudgets

marcolan018:
https://github.com/open-cluster-management/backlog/issues/10847


clyang82:
Good catch 👍 
/lgtm

PR #426, opened 2021-03-29
Upgrade obs to adopt new feature

clyang82:
Upgrade obs to adopt new feature - fetch image from obs cr.


songleo (approved):

songleo (approved):

clyang82:
the failure case we need to update e2e to reflect the observatorium cr structure change. Will submit a separated CR in e2e tests. forcefully merge it.

PR #431, opened 2021-03-31
Tackle replicas

clyang82:
https://github.com/open-cluster-management/backlog/issues/8803

### DIFF
--- a/controllers/multiclusterobservability/multiclusterobservability_controller.go
+++ b/controllers/multiclusterobservability/multiclusterobservability_controller.go
@@ -345,6 +345,44 @@ func (r *MultiClusterObservabilityReconciler) SetupWithManager(mgr ctrl.Manager)
 		return err
 	}
 
+	// Watch for changes to secondary resource Deployment and requeue the owner Observatorium
+	err = c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &observatoriumv1alpha1.Observatorium{},
+	}, predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return updateObservatoriumReplicas(e.Object, nil, "Deployment")
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return updateObservatoriumReplicas(e.ObjectNew, e.ObjectOld, "Deployment")
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to secondary resource statefulSet and requeue the owner Observatorium
+	err = c.Watch(&source.Kind{Type: &appsv1.StatefulSet{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &observatoriumv1alpha1.Observatorium{},
+	}, predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return updateObservatoriumReplicas(e.Object, nil, "StatefulSet")
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return updateObservatoriumReplicas(e.ObjectNew, e.ObjectOld, "StatefulSet")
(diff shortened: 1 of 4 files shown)


controllers/multiclusterobservability/observatorium.go
```diff
@@ -33,8 +34,48 @@ const (
...
+	ObservatoriumAPI     = "observatorium-api"
+	ThanosQuery          = "thanos-query"
+	ThanosQueryFrontend  = "thanos-query-frontend"
+	ThanosRule           = "thanos-rule"
+	ThanosReceive        = "thanos-receive-default"
+	ThanosStoreMemcached = "thanos-store-memcached"
+)
+
+var (
+	replicas1      int32 = 1
```
songleo, line 47: seems we do not use this var.

clyang82: It is designed to be used by compact. Let me remove local variable in compact.

controllers/multiclusterobservability/observatorium.go
```diff
@@ -412,7 +453,7 @@ func newStoreSpec(mco *mcov1beta2.MultiClusterObservability, scSelected string)
 	storeSpec.VolumeClaimTemplate = newVolumeClaimTemplate(
 		mco.Spec.StorageConfig.StoreStorageSize,
 		scSelected)
-	storeSpec.Shards = util.GetReplicaCount("StatefulSet")
+	storeSpec.Shards = &replicas3
```
songleo, line 456: I do not understand why `storeSpec.Shards` value is a fixed value `replicas3 `.

controllers/multiclusterobservability/observatorium.go
```diff
@@ -33,8 +34,48 @@ const (
...
 )
 
+func GetObservatoriumComponentReplicas(componentName string) *int32 {
+	return thanosReplicas[componentName]
+}
+
+func SetObservatoriumComponentReplicas(componentName string, replicas *int32) {
+	for k := range thanosReplicas {
+		// observability-thanos-query-frontend contains thanos-query-frontend and thanos-query
+		if strings.Contains(componentName, ThanosQueryFrontend) {
```
songleo, line 68: why we use this  func `strings.Contains` to check the `componentName`, can not use this like `if k == componentName` to check it.

controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -345,6 +345,44 @@ func (r *MultiClusterObservabilityReconciler) SetupWithManager(mgr ctrl.Manager)
 		return err
 	}
 
+	// Watch for changes to secondary resource Deployment and requeue the owner Observatorium
```
marcolan018, line 348: don't understand the watch function here fully.
The purpose for the watch is to update the thanosReplica? why need to trigger reconcile?

controllers/multiclusterobservability/observatorium.go
clyang82: Replicas are for same data (HA). Each shard has different data, so we need different statefulset. We need to do more tests to see if we need to have more shards.

controllers/multiclusterobservability/observatorium.go
clyang82: `observability-thanos-query-frontend` contains `thanos-query-frontend` and `thanos-query` so check `thanos-query-frontend` firstly. if contained, then return to avoid make changes on `thanos-query`

controllers/multiclusterobservability/multiclusterobservability_controller.go
clyang82: the watch just update the `thanosReplicas` in memory. still reply on the reconcile to update the observatorium CR. normally, we just need to watch for update. watch create is for deleting the mco-o case.

songleo (approved):

marcolan018 (approved):
lgtm

clyang82:
@songleo You are correct. using `contains` may introduce some problems. I found that the `observability-thanos-query-frontend` is set to 1 due to `observability-thanos-query-frontend-memcached` contains `observability-thanos-query-frontend`.  I have changed it to use `equal`

songleo (approved):

PR #436, opened 2021-04-01
Support configure replicas for alertmanager/grafana/rbac-query-proxy

clyang82:
https://github.com/open-cluster-management/backlog/issues/8803


songleo:
LGTM,  maybe we need to update e2e and doc due to we change the deployment or statefulset name, right?

clyang82:
> LGTM, maybe we need to update e2e and doc due to we change the deployment or statefulset name, right?

Yes. already updated.

PR #437, opened 2021-04-01
 support roks

marcolan018:
https://github.com/open-cluster-management/backlog/issues/11120


clyang82:
@marcolan018 How does klusterlet support roks? Do we follow the same way? Thanks.

marcolan018:
@clyang82 
klusterlet supports roks from 2.2, and in fact, I almost reuse their code.

clyang82:
> @clyang82
> klusterlet supports roks from 2.2, and in fact, I almost reuse their code.

I saw you have logged an issue - https://github.com/open-cluster-management/backlog/issues/11125 against 2.1. since klusterlet supports roks from 2.2, we cannot support it in 2.1 separately. right?

marcolan018:
@clyang82 that's a typo. I changed to 2.2

songleo (approved):

PR #441, opened 2021-04-06
manage the storageversionmigration resource for api migration.

morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/11096


controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -103,6 +103,14 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, err
 	} else if isTerminating {
 		reqLogger.Info("MCO instance is in Terminating status, skip the reconcile")
+		svmCrdExists, err := util.CheckCRDExist(r.CrdClient, config.StorageVersionMigrationCrdName)
```
clyang82, line 106: Can we add ownerreference for this CR so that we do not need to special handle the deletion case?

controllers/multiclusterobservability/multiclusterobservability_controller_test.go
```diff
@@ -30,6 +30,7 @@ import (
 	utilpointer "k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	migrationv1alpha1 "sigs.k8s.io/kube-storage-version-migrator/pkg/apis/migration/v1alpha1"
```
clyang82, line 33: reference to this one or reference to the openshift forked repo?

controllers/multiclusterobservability/multiclusterobservability_controller.go
morvencao: Actually I tried that, I got error of `can't add owner ownerreference to cluster-scope resource` because `StorageVersionMigration` is cluster-scoped resource.
It's weird because MCO CR is also cluster scoped resource, it should be working.

Now I have to add logic to delete the `StorageVersionMigration` resource after MCO CR is deleted, we can't depend on the kubernetes GC to delete it because we can't add ownerreference.

controllers/multiclusterobservability/multiclusterobservability_controller_test.go
morvencao: Good point, the import path for forked repo is not changed, so we have to replace the repo in `go.mod`

```
replace (
	sigs.k8s.io/kube-storage-version-migrator => github.com/openshift/kubernetes-kube-storage-version-migrator v0.0.3-0.20210302135122-481bd04dbc78
)
```

see: https://github.com/morvencao/multicluster-observability-operator/blob/br_create_storage_version_migration/go.mod#L39

PR #460, opened 2021-04-14
update dashes, fix workload dash metric bug

RobDolares:


manifests/base/grafana/dash-k8s-apiserver.yaml
```diff
@@ -977,14 +981,16 @@ data:
...
+        "kubernetes-mixin"
+      ],
       "templating": {
         "list": [
           {
             "current": {
               "selected": false,
-              "text": "Observatorium",
-              "value": "Observatorium"
+              "text": "Prometheus",
```
marcolan018, line 992: why change from Observatorium to Prometheus?

manifests/base/grafana/dash-k8s-compute-resources-pod.yaml
```diff
@@ -1914,8 +1916,8 @@ data:
             "allValue": null,
             "current": {
               "selected": false,
-              "text": "cert-manager-webhook-559cdf74cf-2hml5",
-              "value": "cert-manager-webhook-559cdf74cf-2hml5"
+              "text": "acm-custom-registry-d4f6cfc6d-f4trs",
```
marcolan018, line 1919: should we keep text/value as empty string in dashboard yaml?

manifests/base/grafana/dash-k8s-apiserver.yaml
RobDolares: Probably changed by default from development instance.

manifests/base/grafana/dash-k8s-compute-resources-pod.yaml
RobDolares: I think so yes - likely exists in all of the current dashboards - will update.

PR #470, opened 2021-04-20
use only one manifestwork in observability

marcolan018:
https://github.com/open-cluster-management/backlog/issues/11704


clyang82:
How many resources we have in `manifestwork` right now?

/lgtm

marcolan018:
@clyang82 
depends on different env, maybe 10~12

PR #471, opened 2021-04-21
add microdnf update

songleo:
https://github.com/open-cluster-management/backlog/issues/11551


songleo:
@morvencao FYI

PR #481, opened 2021-04-25
cert rotation

marcolan018:
https://github.com/open-cluster-management/backlog/issues/11157

support certificate rotation, not include CA rotation, which will be included in another pr

### DIFF
--- a/pkg/certificates/cert_controller.go
+++ b/pkg/certificates/cert_controller.go
@@ -1,0 +1,155 @@
+// Copyright (c) 2021 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package certificates
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"os"
+	"reflect"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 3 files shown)


pkg/certificates/cert_controller.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	watchlist := cache.NewListWatchFromClient(kubeClient.CoreV1().RESTClient(), "secrets", config.GetDefaultNamespace(),
+		fields.Everything())
+	_, controller := cache.NewInformer(
+		watchlist,
+		&v1.Secret{},
+		time.Minute*60,
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
+				restartPods(*kubeClient, *obj.(*v1.Secret))
```
clyang82, line 56: the rbac-query-proxy and  observatorium-api  will be restarted for the first time when the secrets are created. is it  desired behaviour?

pkg/certificates/cert_controller.go
```diff
@@ -0,0 +1,155 @@
...
+			},
+
+			DeleteFunc: func(obj interface{}) {
+			},
```
clyang82, line 60: I think we should restart the pod if the secrets are deleted, because the secrets will be recreated. WDYT?

marcolan018: the restart will happen in Add event when the certificates recreated after deletion

pkg/certificates/cert_controller.go
marcolan018: in function updateDeployLabel, the restart label will only be injected when secret is created after deployment. So for first time,  there will be no restart.

clyang82: 👍 You can follow the same way of cert-manager does.

PR #484, opened 2021-04-26
reduce metrics allowlist

haoqing0110:
https://github.com/open-cluster-management/backlog/issues/11680
Go through grafana dashboard, find below metrics are not used. 
@bjoydeep @marcolan018 could you help review and confirm if we can remove these metrics?


marcolan018:
@haoqing0110 
what's the criteria to remove the metrics from the allow list?

haoqing0110:
@marcolan018 just go through the metrics_allowlist.yaml, grep each in grafana dashboard folder `./manifests/base/grafana`, if not find, treat it as not used.

RobDolares:
Please hold on this merge.  I recently added metrics to fix dashboard discrepancies with underlying ocp versions 4.6, 4.7, and 4.8.  So they are not in updated dashboards yet.

clyang82:
/hold
We need to review them:
1. how many time-series for those metrics?
2. it is breaking change if the user already use those metrics to build their dashboard. we need to consider to have document ask the user to add them into custom_allowlist

Let us think over carefully.

PR #491, opened 2021-04-27
fix incorrect name.

morvencao:


clyang82:
@morvencao Do you need to update `bundles` folder?

morvencao:
@clyang82 I have executed `make bundle -f Makefile.prow` command, the reason why you see the bundle is not updated because we just change the file name, the content is not updated, so the bundle generated is also not changed.

PR #493, opened 2021-04-28
update ruleSpec.AlertmanagersURL to ruleSpec.AlertmanagerURLs

clyang82:
right now, we cannot specify the alertmanager url in rule because the name is changed in upstream.


morvencao (commented):
/lgtm

songleo (approved):

PR #500, opened 2021-04-29
Add resource requirement  for metrics-collector

songleo:
- Add resource requirement  for metrics-collector and fix https://github.com/open-cluster-management/backlog/issues/5764 
- Add resource requirement  for endpoint-monitoring-operator


songleo:
need to update doc and api doc

marcolan018:
also need to update here:
https://github.com/open-cluster-management/multicluster-observability-operator/blob/3a97707319781bd40289dc36678448f36fdd457c/controllers/placementrule/manifestwork.go#L314

clyang82:
@songleo Where is the resources from?

PR #501, opened 2021-04-30
Filter by job=apiserver

clyang82:
fixed https://github.com/open-cluster-management/backlog/issues/11681


clyang82:
From the log, we can see that the failure is due to apply `v1beta1` CR. We met this same error several times. @morvencao comments?

PR #506, opened 2021-05-03
add metrics for scheduler dashboard

RobDolares:

related Issue: https://github.com/open-cluster-management/backlog/issues/12023


marcolan018:
@RobDolares 
the new-added metrics is some one renamed?

RobDolares:
> @RobDolares
> the new-added metrics is some one renamed?

@marcolan018 - most of these are not replacing. - These are Prometheus rules in _openshift-monitoring_
So can we still make use of them by adding the record to our allowlist?

Screen Shot 2021-05-04 at 6 52 24 PM

PR #508, opened 2021-05-05
add unit test for certificates scenarios

marcolan018:

### DIFF
--- a/pkg/certificates/cert_agent_test.go
+++ b/pkg/certificates/cert_agent_test.go
@@ -1,0 +1,27 @@
+// Copyright (c) 2021 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package certificates
+
+import (
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	clusterv1 "github.com/open-cluster-management/api/cluster/v1"
+)
+
+func TestCertAgent(t *testing.T) {
+	agent := &ObservabilityAgent{}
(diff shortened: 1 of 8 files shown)


pkg/certificates/cert_agent_test.go
```diff
@@ -0,0 +1,27 @@
...
+	agent.Manifests(nil, nil)
+	options := agent.GetAgentAddonOptions()
+	cluster := &clusterv1.ManagedCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: clusterName,
+		},
+	}
+	configs := options.Registration.CSRConfigurations(cluster)
+	if len(configs) != 2 {
+		t.Fatal("Wrong CSRConfigurations")
```
songleo, line 25: do we need a log to print the `configs` when test failure?

pkg/certificates/cert_controller_test.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func TestOnDelete(t *testing.T) {
+	caSecret := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      serverCACerts,
+			Namespace: namespace,
+		},
+		Data: map[string][]byte{
+			"tls.crt": []byte("new cert-"),
```
songleo, line 84: should be `new cert` ?

pkg/certificates/cert_agent_test.go
marcolan018: here we only need to check the length for the configs

songleo (approved):

PR #509, opened 2021-05-05
add metrics for additional control plane dashboards

RobDolares:
related issue: https://github.com/open-cluster-management/backlog/issues/12023


marcolan018:
@RobDolares 
at this time, we need to be very careful to add new metrics into allowlist due to perf/scalability concern.
Can you provide the count of time series for each metrics in the list? We need to guarantee the new-added one will not have big impact on the performance.

@bjoydeep fyi

RobDolares:
@bjoydeep @marcolan018 
shared time series doc updated with counts of time series for these metrics.
How should we proceed?

PR #511, opened 2021-05-06
filter cache for secrets resource, or store the ones in obs namespace

marcolan018:


marcolan018:
/hold

need to verify this change in 1k sno cluster env. If still cannot work, need to increase the memory limit of mco operator

pkg/util/filtered_cache.go
```diff
@@ -0,0 +1,571 @@
+// Copyright 2020 IBM Corporation
```
clyang82, line 1: Can we directly put the IBM source code into our code? You may bring it to open source meeting to discuss?

main.go
```diff
@@ -92,13 +93,20 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	gvkLabelMap := map[schema.GroupVersionKind]util.Selector{
```
clyang82, line 96: Can we limit to other resources? e.g.: configmap
What resources do we watch in managedcluster namespaces? I am thinking we may limit the deployment/statefulset/configmap/secret etc to the default namespace

marcolan018: add other resources.
also filter manifestwork, only cache those which has our owner label

pkg/util/filtered_cache.go
marcolan018:
```
I think it depends on the license on the IBM repo you are interested in.  ie is it Apache 2.0? We obviously can't use any closed source IBM stuffs and share it openly.
```
Above is the answer from Mike


@clyang82 @morvencao 
do you think we need to use current way, or fork the repo to our org and use it with our fix?

clyang82: I think we may use it in the current way, since the license is Apache 2 and we keep the copyright information.

morvencao: It's apache v2 license.
I think the current header is correct, we have to keep the original copyright at the top of the file and also what we changed.

main.go
clyang82: Can you add `service` as well?

morvencao: As we discussed in slack DM, we need to make sure that the filter we added doesn't impact the watch behavior, it looks like we can only watch resources in the cache.

marcolan018:
already verified the fix can work.
w/o this fix, in the 1k sno cluster,mco. operator will use more than 1G memory. With the filtercache, the memory usage is about 300M

main.go
marcolan018: @morvencao no, it can work other resources which not claimed.  I have verified that.
@clyang82 svc added

marcolan018:
/hold

hold to wait https://github.com/IBM/controller-filtered-cache upgrade controller-runtime, so that we can use it directly

morvencao:
Raised [PR]( https://github.com/IBM/controller-filtered-cache/pull/27) against repo: https://github.com/IBM/controller-filtered-cache, the repo owner agreed to merge that.

marcolan018:
the pr in https://github.com/IBM/controller-filtered-cache already merged, update our code to use it, unhold the change now
/unhold

PR #512, opened 2021-05-06
add recording rule for apiserver_request_duration_seconds_bucket

haoqing0110:
Ref https://github.com/open-cluster-management/backlog/issues/11950


marcolan018 (changes requested):
still missing some changes
1. besides apiserver_request_duration_seconds_bucket, also needs to handle apiserver_request_total as default one
2. needs to update grafana dashboards to use the new metrics

haoqing0110:
@marcolan018  thanks, will create another PR to handle all the record rules.

PR #513, opened 2021-05-06
support metrics removal from default allowlist

songleo:
if a metric name ending in `-`,  we will delete this metric from the default allowlist.

https://github.com/open-cluster-management/backlog/issues/12106


marcolan018:
@songleo 
for "match" section in allowlist, we also need to support removal

PR #516, opened 2021-05-07
fix deleted metrics

songleo:
if a metric name ending in `-`,  we will delete this metric from the default allowlist.

https://github.com/open-cluster-management/backlog/issues/12106


controllers/placementrule/manifestwork.go
```diff
@@ -312,14 +311,26 @@ func getAllowList(client client.Client, name string) (*MetricsAllowlist, error)
 	return allowlist, nil
 }
 
-func handleDeletedMetrics(metricNameList []string) []string {
-	nameList := []string{}
-	for _, name := range metricNameList {
+func handleDeletedMetrics(defaultAllowlist []string, customAllowlist []string) []string {
```
clyang82, line 314: Suggest to use `mergeMetrics` as function name, because it is for both `append` and `delete`.

songleo: Thank you for your advice.

controllers/placementrule/manifestwork.go
```diff
@@ -312,14 +311,26 @@ func getAllowList(client client.Client, name string) (*MetricsAllowlist, error)
...
+
+	allowMetrics := []string{}
+	for _, name := range defaultAllowlist {
+		if !deletedMetrics[name] {
+			allowMetrics = append(allowMetrics, name)
+		}
+	}
+
+	allowMetrics = append(allowMetrics, customMetrics...)
+	return allowMetrics
```
clyang82, line 333: Your current logic cannot handle this case: in custom metrics list, there are both `metrics1` and `metrics1-`. Even if user can manually remove from custom metrics list. but it is better to re-org your logic as follows:
1. form the whole metrics list w/o metrics with `-` as suffix
2. filter out metrics with `-` as suffix

clyang82 (changes requested):

controllers/placementrule/manifestwork.go
songleo: 👍  cool, got it. need to handle repeated metrics.

PR #523, opened 2021-05-11
update Request Rate by  HTTP Return Code panel to show one line per code

haoqing0110:
Add sum() to expr, so that only show on line per http code, same as previous dashboard without recording rule.

without sum()
Screen Shot 2021-05-11 at 9 06 40 AM

with sum()
Screen Shot 2021-05-11 at 9 06 03 AM

compared with previous dashboard when no recording rule
Screen Shot 2021-05-10 at 11 21 44 PM


marcolan018:
@haoqing0110 
thanks for fixing this problem

PR #526, opened 2021-05-11
add mco annotation of cert duration for dev/test

marcolan018:
fix: https://github.com/open-cluster-management/backlog/issues/12368

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"os"
 	"strings"
+	"time"
 
 	ocinfrav1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -444,3 +447,20 @@ func SetCustomRuleConfigMap(hasConfigMap bool) {
...
+	if annotations != nil && annotations[AnnotationCertDuration] != "" {
+		d, err := time.ParseDuration(annotations[AnnotationCertDuration])
+		if err != nil {
+			log.Error(err, "Failed to parse cert duration, use default one", "annotation", annotations[AnnotationCertDuration])
+		} else {
+			certDuration = d
+			return
+		}
+	}
+	certDuration = time.Hour * 24 * 365
```
songleo, line 465: I think we can remove this line. because `certDuration ` default is `time.Hour * 24 * 365`.

songleo: It's my fault. it should override the default. Got it.

songleo (approved):

PR #531, opened 2021-05-13
fixed resource name

songleo:
- we should have a fixed resource name because  the limits on name length
- https://github.com/open-cluster-management/backlog/issues/12476


marcolan018:
seems the changes replace all places where mco name are used
for some places, e.g. app.kubernetes.io/instance, value of this label should be mco name, not a fixed prefix?

songleo:
> seems the changes replace all places where mco name are used
> for some places, e.g. app.kubernetes.io/instance, value of this label should be mco name, not a fixed prefix?

yes, you are right, I am working on it. thanks.

PR #533, opened 2021-05-14
fixed resource name

songleo:
- we should have a fixed resource name because  the limits on name length
- https://github.com/open-cluster-management/backlog/issues/12393


songleo:
depend on https://github.com/open-cluster-management/observability-kind-cluster/pull/5

songleo:
depend on https://github.com/open-cluster-management/observability-e2e-test/pull/182

PR #535, opened 2021-05-14
placement controller reconcile refactoring

marcolan018:
https://github.com/open-cluster-management/backlog/issues/12507

refactoring on current reconcile loop in placement controller. If event is from one cluster namespace, but not from mco cr, or res in observability namespace, it means this event should not affect obsaddon/manifestwork of in other namespaces.


controllers/placementrule/placementrule_controller.go
```diff
@@ -254,13 +263,20 @@ func createAllRelatedRes(
...
 		currentClusters = util.Remove(currentClusters, decision.ClusterNamespace)
-		err = createManagedClusterRes(client, restMapper, mco, imagePullSecret,
-			decision.ClusterName, decision.ClusterNamespace)
-		if err != nil {
-			failedCreateManagedClusterRes = true
-			log.Error(err, "Failed to create managedcluster resources", "namespace", decision.ClusterNamespace)
+		// only handle the request namespace if the request resource is not from observability  namespace
+		if request.Namespace == "" || request.Namespace == config.GetDefaultNamespace() ||
+			request.Namespace == decision.ClusterNamespace {
+			log.Info("Monitoring operator should be installed in cluster", "cluster_name", decision.ClusterName)
```
clyang82, line 270: We may fix the problem that triggered the reconcilier. in other words, we should not watch or just ignore the events do not effect on obsaddon/manifestwork. I think `EnqueueRequestForObject` is for this purpose. I am not why we need to make changes here.

marcolan018: we cannot ignore the event since manifestwork/obsaddon update event is need to be handled also.
Purpose for this change is to go through all manifestworks if global resources updated, e.g. a custom allow list configmap added, but for non-global resource update, e.g. one manifestwork is deleted by user by accident, then reconcile loop should only handle the manifestwork for that namespace but don't need to touch others

PR #537, opened 2021-05-17
update route path for alertmanager.

morvencao:
fix: https://github.com/open-cluster-management/backlog/issues/12475
Update route path for alertmanager, so that only APIs for alertmanager are exposed.


morvencao:
rebased the code for testing, please take another look @clyang82

PR #539, opened 2021-05-18
fix matchlist for deleted metrics

songleo:


clyang82:
@songleo Can it be verified in your e2e test case?

songleo:
> @songleo Can it be verified in your e2e test case?

yes, we can.

PR #540, opened 2021-05-18
use fitlercache for placementrule&serviceaccount

marcolan018:
related with https://github.com/open-cluster-management/backlog/issues/12506
use filteredcache for the watch resources to reduce the memory usage, especially in large scale env


morvencao (commented):
/lgtm

PR #553, opened 2021-05-24
Support advanced config in CR

clyang82:
fixed: https://github.com/open-cluster-management/backlog/issues/11621
fixed: https://github.com/open-cluster-management/backlog/issues/11909

https://github.com/open-cluster-management/backlog/issues/11996


pkg/config/config.go
```diff
@@ -205,34 +230,76 @@ var (
...
+		if advanced.Query != nil {
+			replicas = advanced.Query.Replicas
+		}
+	case ThanosQueryFrontend:
+		if advanced.QueryFrontend != nil {
+			replicas = advanced.QueryFrontend.Replicas
+		}
+	case ThanosQueryFrontendMemcached:
+		if advanced.QueryFrontendMemcached != nil {
+			replicas = advanced.QueryFrontendMemcached.Replicas
```
morvencao, line 276: we may need to figure out how to eliminate the switch sentence, but I can't find a better solution.
Maybe we can change `AdvancedConfig` go struct in API to something like `map[string]ComponentSetting`, but that will losing validation...

api/v1beta2/multiclusterobservability_types.go
```diff
@@ -48,44 +40,171 @@ type MultiClusterObservabilitySpec struct {
...
+
+// Thanos Store Spec
+type Spec struct {
+	// Compute Resources required by the store.
+	// +optional
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+	// Replicas for the store.
+	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
+}
```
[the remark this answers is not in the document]
clyang82: yes. unused code.

controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -458,11 +416,11 @@ func (r *MultiClusterObservabilityReconciler) SetupWithManager(mgr ctrl.Manager)
...
 		For(&mcov1beta2.MultiClusterObservability{}, builder.WithPredicates(mcoPred)).
-		// Watch for changes to secondary resource Deployment and requeue the owner Observatorium
-		Watches(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(deployPred)).
-		// Watch for changes to secondary resource statefulSet and requeue the owner Observatorium
-		Watches(&source.Kind{Type: &appsv1.StatefulSet{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(statefulsetPred)).
-		// Watch for changes to secondary resource ConfigMap and requeue the owner MultiClusterObservability
+		// Watch for changes to secondary resource Deployment and requeue the owner MultiClusterObservability
+		Owns(&appsv1.Deployment{}).
+		// Watch for changes to secondary resource statefulSet and requeue the owner MultiClusterObservability
+		Owns(&appsv1.StatefulSet{}).
```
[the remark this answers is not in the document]
clyang82: Yes. the previous requirement is a little confusion and tricky.

pkg/config/config.go
clyang82: Need more thinking about it. maybe have another PR to enhancement later.

api/v1beta2/multiclusterobservability_types.go
```diff
@@ -48,44 +40,161 @@ type MultiClusterObservabilitySpec struct {
 	ObservabilityAddonSpec *observabilityshared.ObservabilityAddonSpec `json:"observabilityAddonSpec,omitempty"`
 }
 
+type AdvancedConfig struct {
```
marcolan018, line 43: do we only need provide the resource configuration for observabotirum deployment/statefulset? How about others such as grafana

pkg/config/config.go
```diff
@@ -205,34 +230,76 @@ var (
...
-			thanosReplicas[k] = replicas
-			return
+func GetReplicas(component string, advanced *observabilityv1beta2.AdvancedConfig) *int32 {
+	if advanced == nil {
+		return Replicas[component]
+	}
+	var replicas *int32
+	switch component {
+	case ObservatoriumAPI:
+		if advanced.ObservatoriumAPI != nil {
```
marcolan018, line 263: do we need to check whether advanced is nil or not?

api/v1beta2/multiclusterobservability_types.go
clyang82: I think so far we don't have cases need to scale `grafana` and `alertmanager`. Do you have case for them?

pkg/config/config.go
clyang82: we have it from line 257 to line 259

api/v1beta2/multiclusterobservability_types.go
marcolan018: grafana always uses much more memory than most of other containers in our testing. It should have the scale requirement considering more clients.
No case for alertmanager yet since we have almost no workload on it in our test cases. But imo we'd better to have that if all of others expose resources as tweakable params.

clyang82: Agreed. Let me have subsequent PR to support grafana and alertmanager.

marcolan018: ok

PR #554, opened 2021-05-24
Use placements.cluster.open-cluster-management.io

clyang82:
fixed: https://github.com/open-cluster-management/backlog/issues/12747


clyang82:
We cannot switch to `placement` right now. Need think over carefully. right now, the `placement` only apply for `ManagedClusterSet`. it is not applied to `Managedcluster` directly. And there is not default `ManagedClusterSet` created for all managedcluster. it impacts on the current observability behaviour.

clyang82:
close in favor of https://github.com/open-cluster-management/multicluster-observability-operator/pull/644

PR #559, opened 2021-05-26
fix infinite reconcile

songleo:


morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/12472

PR #567, opened 2021-06-02
update cookie secret for alertmanager proxy.

morvencao:
try to resolve the git leak issue found in https://github.com/open-cluster-management/backlog/issues/12516

### DIFF
--- a/manifests/base/alertmanager/alertmanager-proxy.yaml
+++ b/manifests/base/alertmanager/alertmanager-proxy.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+data: {}
+kind: Secret
+metadata:
+  labels:
+    app.kubernetes.io/name: alertmanager-proxy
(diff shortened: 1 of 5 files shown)


morvencao:
Looks like the jobs are not triggerred.

morvencao:
@clyang82 Would you take another look at this?
I found simply deleting the `cookie-secret` will report error, so I have to figure out other solution:
passing the cookie secret file mount from the secret and the secret contains randomly generated seed string.
To avoid the infinite reconcile loop, we need to add `skip-creation-if-exist: "true"` annotation.

clyang82:
@haoqing0110 Do we need have the same fix in management-ingress?

haoqing0110:
@clyang82 yes, similar issue to be fixed in https://github.com/open-cluster-management/management-ingress-chart/pull/110

PR #570, opened 2021-06-03
fix problem in service update

marcolan018:
https://github.com/open-cluster-management/backlog/issues/13041


pkg/deploying/deployer.go
```diff
@@ -144,6 +144,8 @@ func (d *Deployer) updateService(desiredObj, runtimeObj *unstructured.Unstructur
 	}
 
 	if !apiequality.Semantic.DeepDerivative(desiredService.Spec, runtimeService.Spec) {
+		desiredService.ObjectMeta.ResourceVersion = runtimeService.ObjectMeta.ResourceVersion
```
songleo, line 147: I wonder if we need to update these field values. `ResourceVersion` and `ClusterIP ` should update via apiserver?

songleo (approved):

PR #573, opened 2021-06-05
Status goroutings

clyang82:
Have a separate goroutine to update the MCO status. it is heavy to update the status by triggering a reconcile.

### DIFF
--- a/controllers/multiclusterobservability/multiclusterobservability_status.go
+++ b/controllers/multiclusterobservability/multiclusterobservability_status.go
@@ -21,11 +21,11 @@ import (
 )
 
 var (
-	stopStatusUpdate       = make(chan struct{})
-	stopCheckReady         = make(chan struct{})
-	requeueStatusUpdate    = make(chan struct{})
-	mcoCreateTimeout       = 30 * time.Minute
-	updateStatusIsRunnning = false
+	stopStatusUpdate            = make(chan struct{})
+	stopCheckReady              = make(chan struct{})
+	requeueStatusUpdate         = make(chan struct{})
+	updateStatusIsRunnning      = false
+	updateReadyStatusIsRunnning = false
 )
 
 // Start goroutines to update MCO status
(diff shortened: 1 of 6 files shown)


controllers/multiclusterobservability/multiclusterobservability_status.go
```diff
@@ -19,6 +20,87 @@ import (
...
+			updateStatusIsRunnning = true
+			// defer close(stopStatusUpdate)
+			// defer close(requeueStatusUpdate)
+			for {
+				select {
+				case <-stopStatusUpdate:
+					updateStatusIsRunnning = false
+					stopCheckReady <- struct{}{}
+					log.V(1).Info("status update goroutine is stopped.")
+					return
```
morvencao, line 44: nit: `return` is not needed.

marcolan018:
move the status logic out from reconcile, purpose is to reduce the possibility of endless reconcile?

clyang82:
> move the status logic out from reconcile, purpose is to reduce the possibility of endless reconcile?

Yes. 2 purposes here:
1. One is update status is isolated. it is nothing with reconcile. it is just triggered by reconciling to start.
2. Reduce the possibility of endless reconcile

It may apply to certmanagercontroller as well. At least, we should not start certctrl if there is no MCO CR exists in environment.

morvencao (commented):
@clyang82 
It's not very clear.
Can we start the mco update `StartStatusUpdate(r.Client, instance)` with `go run StartStatusUpdate()`, and in that case, we also need to update the  StartStatusUpdate() to move `for select...` out of go routine? WDYT?

controllers/multiclusterobservability/multiclusterobservability_status.go
clyang82: use `return` to break this endless `for`.

clyang82:
> @clyang82
> It's not very clear.
> Can we start the mco update `StartStatusUpdate(r.Client, instance)` with `go run StartStatusUpdate(r.Client, instance)`, and in that case, we also need to update the StartStatusUpdate() to move `for select...` out of go routine? WDYT?
> 
> Also can we add parameter for the end signal channel to `StartStatusUpdate()`.

There are 2 goroutines in `StartStatusUpdate`.

morvencao (commented):
the changes look good to me, leave this to @marcolan018 @songleo for any additional comments.

songleo:
lgtm

songleo (approved):

PR #574, opened 2021-06-06
documenting PV usage

bjoydeep:
This document describes the usage of Persistent Volume by Observability.

### DIFF
--- a/docs/MoreAboutPersistentStorage.md
+++ b/docs/MoreAboutPersistentStorage.md
@@ -1,6 +1,6 @@
-# Persistent Stores used in ACM Observability
+# Persistent Stores used in Open Cluster Management Observability
 
-ACM Observability is a stateful application. It creates following persistent volumes (there are more than 1 copies as it runs as stateful sets).
+Open Cluster Management Observability is a stateful application. It creates following persistent volumes (there are more than 1 copies as it runs as stateful sets).
 
 ### List of Stateful sets
 
--- a/docs/MoreAboutPersistentStorage.md
+++ b/docs/MoreAboutPersistentStorage.md
@@ -1,0 +1,44 @@
+# Persistent Stores used in ACM Observability
+
+ACM Observability is a stateful application. It creates following persistent volumes (there are more than 1 copies as it runs as stateful sets).
+
+### List of Stateful sets
+
+| Name | Purpose |
+| ----------- | ----------- |
+| alertmanager | Alertmanager stores its silencing, nflog etc in its storage. nflog is append-only log of active/resolved notifications along with the notified receiver, and a hash digest of the notification's identifying contents|
+| thanos-compact | The compactor needs local disk space to store intermediate data for its processing as well as bucket state cache. Generally, for medium sized bucket about 100GB should be enough to keep working as the compacted time ranges grow over time. However, this highly depends on size of the blocks. In worst case scenario compactor has to have space adequate to 2 times 2 weeks (if your maximum compaction level is 2 weeks) worth of smaller blocks to perform compaction. First, to download all of those source blocks, second to build on disk output of 2 week block composed of those smaller ones. On-disk data is safe to delete between restarts and should be the first attempt to get crash-looping compactors unstuck. However, it’s recommended to give the Compactor persistent disk in order to effectively use bucket state cache between restarts. |
+| thanos-rule | The thanos ruler evaluates Prometheus recording and alerting rules against chosen query API via repeated query. Rule results are written back to disk in the Prometheus 2.0 storage format. |
+| thanos-receive-default | Thanos receiver accepts incoming data (Prometheus remote-write requests) and writes these into a local instance of the Prometheus TSDB. Periodically (every 2 hours) TSDB blocks are uploaded to the Object storage for long term storage and compaction. The amount of hours/days of data retained in this stateful set (acts a local cache) was fixed in ACM 2.1/2.2. It has been exposed as an API parameter in ACM 2.3: _RetentionInLocal_ |
+| thanos-store-shard| It acts primarily as an API gateway and therefore does not need significant amounts of local disk space. It joins a Thanos cluster on startup and advertises the data it can access. It keeps a small amount of information about all remote blocks on local disk and keeps it in sync with the bucket. This data is generally safe to delete across restarts at the cost of increased startup times. |
+| thanos-store-memcached | Thanos Store Gateway supports an index cache using Memached to speed up postings and series lookups from TSDB blocks indexes. |
+| | |
+
+
+
+
+### Configuring the Stateful sets
+
+In ACM 2.1 and ACM 2.2 only the following was exposed using the Mutlicluster Observability Operator API - _observability.open-cluster-management.io/v1beta1_
+
+```
+    //defaults shown below
+    statefulSetSize: 10Gi
+    statefulSetStorageClass: gp2
+```
+In ACM 2.3, taking into consideration that one size fits all results in wasting space, we have allowed all settings can be individually tweaked in the enhanced API - _observability.open-cluster-management.io/v1beta2_
+
+```
+    //defaults shown below
+    StorageClass: gp2
+	AlertmanagerStorageSize: 1Gi 
+	RuleStorageSize: 1Gi
+	CompactStorageSize: 100 Gi
+	ReceiveStorageSize: 100 Gi
+	StoreStorageSize: 10 Gi
+
+```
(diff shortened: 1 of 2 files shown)


docs/MoreAboutPersistentStorage.md
```diff
@@ -0,0 +1,44 @@
+# Persistent Stores used in ACM Observability
```
clyang82, line 1: I think OCM should be better than ACM here.

docs/MoreAboutPersistentStorage.md
```diff
@@ -0,0 +1,44 @@
...
+    //defaults shown below
+    statefulSetSize: 10Gi
+    statefulSetStorageClass: gp2
+```
+In ACM 2.3, taking into consideration that one size fits all results in wasting space, we have allowed all settings can be individually tweaked in the enhanced API - _observability.open-cluster-management.io/v1beta2_
+
+```
+    //defaults shown below
+    StorageClass: gp2
+	AlertmanagerStorageSize: 1Gi
```
clyang82, line 34: decrease indent.

docs/MoreAboutPersistentStorage.md
```diff
@@ -0,0 +1,44 @@
...
+| thanos-receive-default | Thanos receiver accepts incoming data (Prometheus remote-write requests) and writes these into a local instance of the Prometheus TSDB. Periodically (every 2 hours) TSDB blocks are uploaded to the Object storage for long term storage and compaction. The amount of hours/days of data retained in this stateful set (acts a local cache) was fixed in ACM 2.1/2.2. It has been exposed as an API parameter in ACM 2.3: _RetentionInLocal_ |
+| thanos-store-shard| It acts primarily as an API gateway and therefore does not need significant amounts of local disk space. It joins a Thanos cluster on startup and advertises the data it can access. It keeps a small amount of information about all remote blocks on local disk and keeps it in sync with the bucket. This data is generally safe to delete across restarts at the cost of increased startup times. |
+| thanos-store-memcached | Thanos Store Gateway supports an index cache using Memached to speed up postings and series lookups from TSDB blocks indexes. |
```
clyang82, line 14: `thanos-store-memcached` does not use PV right now. Do we need to list here? if so, we need to list `thanos-query-frontend-memcached` as well.

docs/MoreAboutPersistentStorage.md
```diff
@@ -0,0 +1,44 @@
...
+| thanos-compact | The compactor needs local disk space to store intermediate data for its processing as well as bucket state cache. Generally, for medium sized bucket about 100GB should be enough to keep working as the compacted time ranges grow over time. However, this highly depends on size of the blocks. In worst case scenario compactor has to have space adequate to 2 times 2 weeks (if your maximum compaction level is 2 weeks) worth of smaller blocks to perform compaction. First, to download all of those source blocks, second to build on disk output of 2 week block composed of those smaller ones. On-disk data is safe to delete between restarts and should be the first attempt to get crash-looping compactors unstuck. However, it’s recommended to give the Compactor persistent disk in order to effectively use bucket state cache between restarts. |
+| thanos-rule | The thanos ruler evaluates Prometheus recording and alerting rules against chosen query API via repeated query. Rule results are written back to disk in the Prometheus 2.0 storage format. |
```
marcolan018, line 11: _RetentionInLocal_  configuration also has impact on rule

docs/MoreAboutPersistentStorage.md
dockerymick: The following statement is partially reviewed. Since this is a part of the open source project, should we be concerned about pre-announcing what is exposed in 2.3? Are there different versions of the open-cluster-management project that can be deployed?

Thanos receiver accepts incoming data (Prometheus remote-write requests) and writes these into a local instance of the Prometheus TSDB. Periodically (every 2 hours), TSDB blocks are uploaded to the object storage for long term storage and compaction. The amount of hours or days of data retained in this stateful set, which acts a local cache was fixed in ACM 2.1/2.2. It has been exposed as an API parameter in ACM 2.3: _RetentionInLocal_

docs/MoreAboutPersistentStorage.md
```diff
@@ -0,0 +1,44 @@
+# Persistent Stores used in ACM Observability
+
+ACM Observability is a stateful application. It creates following persistent volumes (there are more than 1 copies as it runs as stateful sets).
+
+### List of Stateful sets
```
dockerymick, line 5: I would size this H2, so use only 2 hashtags (#)

docs/MoreAboutPersistentStorage.md
```diff
@@ -0,0 +1,44 @@
...
+| thanos-receive-default | Thanos receiver accepts incoming data (Prometheus remote-write requests) and writes these into a local instance of the Prometheus TSDB. Periodically (every 2 hours) TSDB blocks are uploaded to the Object storage for long term storage and compaction. The amount of hours/days of data retained in this stateful set (acts a local cache) was fixed in ACM 2.1/2.2. It has been exposed as an API parameter in ACM 2.3: _RetentionInLocal_ |
+| thanos-store-shard| It acts primarily as an API gateway and therefore does not need significant amounts of local disk space. It joins a Thanos cluster on startup and advertises the data it can access. It keeps a small amount of information about all remote blocks on local disk and keeps it in sync with the bucket. This data is generally safe to delete across restarts at the cost of increased startup times. |
+| thanos-store-memcached | Thanos Store Gateway supports an index cache using Memached to speed up postings and series lookups from TSDB blocks indexes. |
+| | |
+
+
+
+
+### Configuring the Stateful sets
```
dockerymick, line 20: Same here, let's use 2 hashtags.  

Also update the section title to read: 

Configuring the stateful sets

docs/MoreAboutPersistentStorage.md
```diff
@@ -0,0 +1,44 @@
...
+
+In ACM 2.1 and ACM 2.2 only the following was exposed using the Mutlicluster Observability Operator API - _observability.open-cluster-management.io/v1beta1_
+
+```
+    //defaults shown below
+    statefulSetSize: 10Gi
+    statefulSetStorageClass: gp2
+```
```
dockerymick, line 28: Based on my previous comment about versioning, I recommend that we update this to read:

In previous versions of the `open-cluster-management` project, only the `statefulSetSize` and `statefulSetStorageClass` could be configured with the Multicluster Observability Operator API - `observability.open-cluster-management.io/v1beta1`

Instead of using italics for the API, use monospace

docs/MoreAboutPersistentStorage.md
```diff
@@ -0,0 +1,44 @@
...
+    //defaults shown below
+    StorageClass: gp2
+	AlertmanagerStorageSize: 1Gi 
+	RuleStorageSize: 1Gi
+	CompactStorageSize: 100 Gi
+	ReceiveStorageSize: 100 Gi
+	StoreStorageSize: 10 Gi
+
+```
```
dockerymick, line 40: Same concern about pre-announcing. I would also make the following suggestion:

Currently, you can define individual settings in the enhanced API - `observability.open-cluster-management.io/v1beta2`. View the following parameters that can be set:

docs/MoreAboutPersistentStorage.md
```diff
@@ -0,0 +1,44 @@
...
+Do note that the default storage class, as configured in the system, is used for configuring the persistent volumes automatically unless a different storage class is specified in the Custom Resource specification. If no storage class exists - for example in default OpenShift bare metal installtions, it will need to be created or the installation of ACM Observability will not succeed.
+
```
dockerymick, line 42: **Note**: The default storage class, as configured in the system, is used for configuring the persistent volumes automatically unless a different storage class is specified in the custom resource specification. If no storage class exists, for example in default OpenShift bare metal installations, one must be created or the installation of observability fails.

docs/MoreAboutPersistentStorage.md
```diff
@@ -0,0 +1,44 @@
...
+### Object Store
```
dockerymick, line 43: Use 2 hashtags, and the section title should read: 

Object store

docs/MoreAboutPersistentStorage.md
```diff
@@ -0,0 +1,44 @@
...
+In addition to the persistent volumes above, the time series historical data is stored in Object stores. Thanos uses object storage as primary storage for metrics and metadata related to them. Details of Object storage and downsampling will be covered in another document.
```
dockerymick, line 44: Skip a line then update the paragraph to read:

In addition to the persistent volumes previously mentioned, the time series historical data is stored in object stores. Thanos uses object storage as the primary storage for metrics and metadata related to them. See more details about the object storage and downsampling. 

^Is there a reference link that can be provided?

dockerymick (changes requested):

PR #575, opened 2021-06-07
Rhowingt/crit alert policy

RobDolares:
Related Issue: https://github.com/open-cluster-management/backlog/issues/8627

### DIFF
--- a/manifests/base/alertmanager/alert_rules.yaml
+++ b/manifests/base/alertmanager/alert_rules.yaml
@@ -34,6 +34,8 @@ data:
         - alert: CriticalPolicyReport
           annotations:
             summary: There is a policy report with severity of critical.
-            description: "The policy: {{ $labels.policy }} has a severity of critical"
-          expr: sum(acm_policyreport_info{severity="critical"}) by (cluster,category, namespace, pod, policy, severity) > 0
+            description: "The policy: {{ $labels.policy }} has a severity of critical on cluster: {{ $labels.cluster }}"
+          expr: sum(acm_policyreport_info{severity="critical"}) by (cluster, category, clusterID, policy, severity) > 0
           for: 1m
+          labels:
+            severity: critical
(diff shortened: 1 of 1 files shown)


manifests/base/alertmanager/alert_rules.yaml
```diff
@@ -31,4 +31,9 @@ data:
             clusterID: "{{ $labels.clusterID }}"
             PersistentVolumeClaim: "{{ $labels.persistentvolumeclaim }}"
             severity: warning            
-
+        - alert: CriticalPolicyReport
+          annotations:
+            summary: There is a policy report with severity of critical.
+            description: "The policy: {{ $labels.policy }} has a severity of critical"
+          expr: sum(acm_policyreport_info{severity="critical"}) by (cluster,category, namespace, pod, policy)
+          for: 1m
```
bjoydeep, line 39: @RobDolares :
We should add the `severity: critical `to this alert rule
Also please check if the `namespace, pod information` really captures the offending pod/namespace or is just prom monitoring details. Cannot tell without looking at the real data.
Is it reasonable to add clusterid to the `by (cluster...) `as well as add `cluster, clusterid `to the description so that customers have an easy way of know which clusters are affected?

manifests/base/alertmanager/alert_rules.yaml
```diff
@@ -31,4 +31,9 @@ data:
             clusterID: "{{ $labels.clusterID }}"
             PersistentVolumeClaim: "{{ $labels.persistentvolumeclaim }}"
             severity: warning            
-
+        - alert: CriticalPolicyReport
+          annotations:
+            summary: There is a policy report with severity of critical.
+            description: "The policy: {{ $labels.policy }} has a severity of critical"
+          expr: sum(acm_policyreport_info{severity="critical"}) by (cluster,category, namespace, pod, policy)
```
marcolan018, line 38: there should be some operator in the expr? such as ">1"

RobDolares: thx - added `> 0`

manifests/base/alertmanager/alert_rules.yaml
RobDolares: Disaiah and I were looking at those options and the redundancy when a visual alert is configured.  That is why the existing filters are applied.  But I agree with clusterID and will update.

As for namespace and pod - you are correct, it is just prom monitoring deets and we can remove those.

PR #577, opened 2021-06-08
remove default limits for endpoint-metrics-operator and metrics collector

songleo:
https://github.com/open-cluster-management/backlog/issues/13102

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -185,8 +185,6 @@ const (
 
 	MetricsCollectorCPURequets    = "100m"
 	MetricsCollectorMemoryRequets = "100Mi"
-	MetricsCollectorCPULimits     = "100m"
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -878,9 +876,7 @@ func GetResources(component string, advanced *observabilityv1beta2.AdvancedConfi
 
 func GetOBAResources(oba *mcoshared.ObservabilityAddonSpec) *corev1.ResourceRequirements {
 	cpuRequests := MetricsCollectorCPURequets
-	cpuLimits := MetricsCollectorCPULimits
```
clyang82, line 881: How about just set cpuLimits and memoryLimits to be 0? it should be working w/o changes. correct?

songleo:
thanks for the @clyang82 suggestion, let me new another PR and close this one.

PR #580, opened 2021-06-09
fix manifest update error

songleo:

### DIFF
--- a/controllers/placementrule/manifestwork.go
+++ b/controllers/placementrule/manifestwork.go
@@ -123,23 +123,38 @@ func createManifestwork(c client.Client, work *workv1.ManifestWork) error {
 	}
 
 	manifests := work.Spec.Workload.Manifests
-	updated := false
+	needUpdate := false
 	if len(found.Spec.Workload.Manifests) == len(manifests) {
 		for i, m := range found.Spec.Workload.Manifests {
 			if !util.CompareObject(m.RawExtension, manifests[i].RawExtension) {
-				updated = true
+				needUpdate = true
 				break
 			}
 		}
 	} else {
-		updated = true
+		needUpdate = true
 	}
 
-	if updated {
+	if needUpdate {
 		log.Info("Updating manifestwork", namespace, namespace, "name", name)
 		work.ObjectMeta.ResourceVersion = found.ObjectMeta.ResourceVersion
 		err = c.Update(context.TODO(), work)
 		if err != nil {
+			if k8serrors.IsConflict(err) {
+				log.Info("Failed to update manifestwork", "Reason", "Object has been modified")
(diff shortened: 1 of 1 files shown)


clyang82:
@songleo  Can you provide some information about this PR? BTW, what case can lead to update failure? The manifestwork is owned by MCO operator.  if it is conflicted, that must be updated already. is my understanding correct?

controllers/placementrule/manifestwork.go
```diff
@@ -123,23 +123,38 @@ func createManifestwork(c client.Client, work *workv1.ManifestWork) error {
...
 		}
 	} else {
-		updated = true
+		needUpdate = true
 	}
 
-	if updated {
+	if needUpdate {
 		log.Info("Updating manifestwork", namespace, namespace, "name", name)
 		work.ObjectMeta.ResourceVersion = found.ObjectMeta.ResourceVersion
```
clyang82, line 140: Can you guide me why we must set the `ResourceVersion`? Isn't it updated automatically?

songleo: @clyang82 I think we do not need to set ResourceVersion before the update. @marcolan018 told me if we do not set this value, it will update failure. right?

songleo:
> @songleo Can you provide some information about this PR? BTW, what case can lead to update failure? The manifestwork is owned by MCO operator. if it is conflicted, that must be updated already. is my understanding correct?

Sometimes when I modify the resource configuration of the Addon, I found that the MCO operator prints a lot of error logs due to update conflicts. So I think it would be nice to add a retry mechanism for updates.

clyang82:
> > @songleo Can you provide some information about this PR? BTW, what case can lead to update failure? The manifestwork is owned by MCO operator. if it is conflicted, that must be updated already. is my understanding correct?
> 
> Sometimes when I modify the resource configuration of the Addon, I found that the MCO operator prints a lot of error logs due to update conflicts. So I think it would be nice to add a retry mechanism for updates.

I am still confusion with that. We watch `manifestwork`. That means we always cache the latest version. Why will have update faillure unless we update the manifestwork synchronously?

songleo:
@clyang82 I got your mean. seems we have another controller to update this obj, so we will meet the conflict error when mco operator updates this obj. need more time to investigate it.

marcolan018:
@clyang82 @songleo 
manifestwork res has status and that part is not controller by our controller.
I guess that's the cause for the possible conflict

clyang82:
> @clyang82 @songleo
> manifestwork res has status and that part is not controller by our controller.
> I guess that's the cause for the possible conflict

the status update should not change the `ResourceVersion`. if so, that maybe a bug of work controller. @songleo You can have a try.

PR #592, opened 2021-06-15
update ingress to use networking.k8s.io/v1

haoqing0110:
Fix https://github.com/open-cluster-management/backlog/issues/13262


clyang82:
Update to use `networking.k8s.io/v1`?

PR #593, opened 2021-06-15
added policy report alert to a separate group

dislbenn:
Currently, the OOTB alert `CriticalPolicyReport` is included in the alerting group `kubernetes-storage`. We need to separate it into its own group (i.e `policy-reports`).


RobDolares (approved):
/lgtm

PR #602, opened 2021-06-16
[WIP]Watch mch-image-manifest-xxx configmap

clyang82:
ref: https://github.com/open-cluster-management/backlog/issues/13426

### DIFF
--- a/main.go
+++ b/main.go
@@ -133,12 +133,24 @@ func main() {
 		os.Exit(1)
 	}
 
+	podNamespace, found := os.LookupEnv("POD_NAMESPACE")
+	if !found {
+		podNamespace = config.GetDefaultMCONamespace()
+	}
+
+	// set image manifests configmap name. e.g.: mch-image-manifest-2.3.0
+	config.SetImageManifestConfigMapName()
+
 	gvkLabelMap := map[schema.GroupVersionKind]filteredcache.Selector{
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -133,13 +133,25 @@ func main() {
...
+	config.SetImageManifestConfigMapName()
+
 	gvkLabelMap := map[schema.GroupVersionKind]filteredcache.Selector{
 		v1.SchemeGroupVersion.WithKind("Secret"): {
 			FieldSelector: fmt.Sprintf("metadata.namespace==%s", config.GetDefaultNamespace()),
 		},
 		v1.SchemeGroupVersion.WithKind("ConfigMap"): {
 			FieldSelector: fmt.Sprintf("metadata.namespace==%s", config.GetDefaultNamespace()),
 		},
+		v1.SchemeGroupVersion.WithKind("ConfigMap"): {
```
marcolan018, line 151: can we have two entities for the same resource kind in the Selector?

clyang82: That works in my testing. @morvencao thoughts?

morvencao: It should not work.
We will lose the first `configmap` setting.
Did you verify the configmap in `open-cluster-management-observability` is still be watching?

clyang82:
close in favor of #611

PR #604, opened 2021-06-17
fix resource watch from filter cache.

morvencao:
This PR tries to fix the resources(secret/configmap) watch issue from outside of the `open-cluster-management-observability` namespace by replace the `controller-filtered-cache` with controller-runtime's `SelectorsByObject` cache function from: https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/use-selectors-at-cache.md 


morvencao:
Looks like the controller-runtime's cache function based on `SelectorsByObject` from: https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/use-selectors-at-cache.md also can't handle the logic `OR` foe resources that meet different `label-selector` and `field-selector`, it only supports logic `AND` for these selectors.

main.go
```diff
@@ -133,30 +135,32 @@ func main() {
...
+			&appsv1.StatefulSet{}: {
+				Field: fields.SelectorFromSet(fields.Set{"metadata.namespace": config.GetDefaultNamespace()}),
+			},
+			&workv1.ManifestWork{}: {
+				Label: labels.SelectorFromSet(labels.Set{"owner": "multicluster-observability-operator"}),
+			},
+			&placementv1.PlacementRule{}: {
+				Field: fields.SelectorFromSet(fields.Set{"metadata.namespace": config.GetDefaultNamespace()}),
+			},
 		},
```
clyang82, line 164:
```suggestion
cacheOpts := cache.Options{
                Namespace: config.GetDefaultNamespace(),
		SelectorsByObject: cache.SelectorsByObject{
			&corev1.ConfigMap{}: {
				Field: fields.SelectorFromSet(fields.Set{"metadata.namespace": "open-cluster-management"}),
			},
			&workv1.ManifestWork{}: {
				Label: labels.SelectorFromSet(labels.Set{"owner": "multicluster-observability-operator"}),
			},
		},
```

morvencao:
close in favor of #611

PR #606, opened 2021-06-17
 updated policy report name

xrajesh:
Signed-off-by: xavier dharmaiyan <xavier@redhat.com>


manifests/base/config/metrics_allowlist.yaml
```diff
@@ -7,7 +7,7 @@ data:
     names:
       - :node_memory_MemAvailable_bytes:sum
       - ALERTS
-      - acm_policyreport_info
+      - policyreport_info
```
bjoydeep, line 10: @xrajesh this will need to change as well: https://github.com/open-cluster-management/multicluster-observability-operator/blob/main/manifests/base/alertmanager/alert_rules.yaml#L40

marcolan018: after update the name, please inject the metrics name as alphabetical order even some of existing ones not follow that

xrajesh: @bjoydeep @marcolan018 addressed both your comments . Please review.

marcolan018 (approved):
/lgtm

manifests/base/alertmanager/alert_rules.yaml
```diff
@@ -37,7 +37,7 @@ data:
           annotations:
             summary: There is a policy report with severity of critical.
             description: "The policy: {{ $labels.policy }} has a severity of critical on cluster: {{ $labels.cluster }}"
-          expr: sum(acm_policyreport_info{severity="critical"}) by (cluster, category, clusterID, policy, severity) > 0
+          expr: sum(policyreport_info{severity="critical"}) by (managed_cluster_id, category, clusterID, policy, severity) > 0
```
bjoydeep, line 40: correct. Thanks @xrajesh . We missed during the original review that the cluster label - added by Observability - will always be local cluster and will be useless. The managed_cluster_id is real identity of the cluster impacted. Then is ClusyerID also redundant? Will it always have the same Hub Server Id since this metric is always being generated at the Hub. @marcolan018  thoughts?

dislbenn: Wouldn't we need the `cluster` label for the managed cluster? Will the managed cluster id contain the managed cluster name?

dislbenn (approved):
/lgtm

xrajesh:
closing pr due to conflicts

PR #605, opened 2021-06-17
Get operand name from environment if any

clyang82:
ref: https://github.com/open-cluster-management/backlog/issues/13247

PR #607, opened 2021-06-17
expose rbac-query-proxy external api from mgmt-ingress to route

marcolan018:
https://github.com/open-cluster-management/backlog/issues/13410

### DIFF
--- a/controllers/multiclusterobservability/grafana.go
+++ b/controllers/multiclusterobservability/grafana.go
@@ -5,10 +5,11 @@ package multiclusterobservability
 
 import (
 	"context"
+	"fmt"
+	"reflect"
 
 	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -23,7 +24,6 @@ import (
 
 const (
 	defaultReplicas int32 = 1
-	caBundle              = "rbac-query-proxy-serving-certs-ca-bundle"
 )
 
 type GrafanaDatasources struct {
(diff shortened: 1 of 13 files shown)


pkg/rendering/renderer_proxy.go
```diff
@@ -0,0 +1,160 @@
...
+	found, image := mcoconfig.ReplaceImage(
+		r.cr.Annotations,
+		spec.Containers[0].Image,
+		mcoconfig.RBACQueryProxyKey)
+	if found {
+		spec.Containers[0].Image = image
+	}
+
+	// temp code to bypass multi-repo commit problem
+	spec.Containers[0].Image = "blue0/rbac-query-proxy:2.3.0"
```
morvencao, line 85: I believe this image is only for testing. right?

marcolan018: @morvencao 
yes, this code is used to hardcode the rbac proxy. image, so that the e2e test case will not fail since the image built by rbac proxy repo still has the old code 
Will remove this temp code once the change in rbac proxy repo merged

controllers/multiclusterobservability/grafana.go
```diff
@@ -138,5 +125,15 @@ func GenerateGrafanaDataSource(
 		return &ctrl.Result{}, err
 	}
 
+	if !reflect.DeepEqual(grafanaDSFound.Data, dsSecret.Data) {
+		log.Info("Updating grafana datasource secret")
+		dsSecret.ObjectMeta.ResourceVersion = grafanaDSFound.ObjectMeta.ResourceVersion
```
clyang82, line 130: why we need to keep `ResourceVersion` without changes?

pkg/rendering/renderer_proxy.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, err
+	}
+
+	crLabelKey := mcoconfig.GetCrLabelKey()
+	dep := obj.(*v1.Deployment)
+	dep.ObjectMeta.Labels[crLabelKey] = r.cr.Name
+	dep.Spec.Selector.MatchLabels[crLabelKey] = r.cr.Name
+	dep.Spec.Template.ObjectMeta.Labels[crLabelKey] = r.cr.Name
+	dep.Spec.Replicas = config.GetReplicas(config.RBACQueryProxy, r.cr.Spec.AdvancedConfig)
+
```
clyang82, line 53: Here we need to set `dep.Spec.Name`. You can get the correct name when this PR https://github.com/open-cluster-management/multicluster-observability-operator/pull/605 is merged.

controllers/multiclusterobservability/grafana.go
morvencao: maybe it will cause another reconcile?

pkg/rendering/renderer_proxy.go
clyang82: it was covered in #605 ignore this comment.

controllers/multiclusterobservability/grafana.go
marcolan018: it's a no-harm change. 
for some resources, the update will fail if ResourceVersion is empty.

PR #610, opened 2021-06-19
[WIP]limit cache resources.

morvencao:
add `cacheResourcesNamespaces` which is the union for all namespaces in which the resources(secret, configmap and serviceaccount...) need to be cached.
this is added to workaround the limitation of controller-filtered-cache, see: https://github.com/IBM/controller-filtered-cache#limitation
to make sure resources outside of the default observability namespace can also be watched
this may add memory footprint of MCO, but shoud be in control.
TODO(morvencao): Remove this once the community support selector logic OR for filter cache: https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/use-selectors-at-cache.md


morvencao:
close in favor of https://github.com/open-cluster-management/multicluster-observability-operator/pull/611

PR #618, opened 2021-06-22
watch mch and read image manifest configmap by label.

morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/13552


morvencao:
@clyang82 PTAL

controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -127,11 +131,38 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
...
+		}
+		mchistOpts := []client.ListOption{
+			client.InNamespace(podNamespace),
+		}
+		err := r.Client.List(context.TODO(), mchList, mchistOpts...)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		// normally there should only one MCH CR in the cluster
```
clyang82, line 153: Yes. we do not support multiple MCH CR case for now.

controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -127,11 +131,38 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
...
+		if len(mchList.Items) == 1 {
+			mch := mchList.Items[0]
+			mchVer := mch.Status.CurrentVersion
```
clyang82, line 156: We need to ensure `CurrentVersion` is equal to `DesiredVersion` then start reconciling. otherwise, just return. like what we support `pause`.

main.go
```diff
@@ -167,6 +164,9 @@ func main() {
 		placementv1.SchemeGroupVersion.WithKind("PlacementRule"): []filteredcache.Selector{
 			{FieldSelector: fmt.Sprintf("metadata.namespace==%s", config.GetDefaultNamespace())},
 		},
+		mchv1.SchemeGroupVersion.WithKind("MultiClusterHub"): []filteredcache.Selector{
```
clyang82, line 167: will it be error-ed if there is no `MultiClusterHub`?

controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -447,12 +456,45 @@ func (r *MultiClusterObservabilityReconciler) SetupWithManager(mgr ctrl.Manager)
 		Owns(&observatoriumv1alpha1.Observatorium{}).
 		// Watch the configmap for thanos-ruler-custom-rules update
 		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(cmPred)).
-		// Watch the configmap for mch-image-manifest-* update
-		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(imageManifestPred)).
 		// Watch the secret for deleting event of alertmanager-config
-		Watches(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(secretPred)).
-		// actually create the controller with the reconciler
-		Complete(r)
+		Watches(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(secretPred))
```
clyang82, line 460: so we handle 2 cases here:
1. if there is no mch exists, then we just ignore the watch. it applies to instance observability on top of server foundation.
2. for ACM product, we assume there is always MCH exists before starting observability service.
Correct?

controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -447,12 +456,45 @@ func (r *MultiClusterObservabilityReconciler) SetupWithManager(mgr ctrl.Manager)
...
+
+	mchGroupKind := schema.GroupKind{Group: mchv1.SchemeGroupVersion.Group, Kind: "MultiClusterHub"}
+	if _, err := r.RESTMapper.RESTMapping(mchGroupKind, mchv1.SchemeGroupVersion.Version); err == nil {
+		imageManifestCMPred := predicate.Funcs{
+			CreateFunc: func(e event.CreateEvent) bool {
+				return false
```
clyang82, line 466: Do we need to handle the `Create` case?

morvencao: I think we don't need to enqueue the event for create option, we can wait for the MCH status is updated and **stable**, then we read the images manifests.

main.go
morvencao: fixed this by adding check MCH CRD exists before adding the filter cache.

controllers/multiclusterobservability/multiclusterobservability_controller.go
morvencao: Correct, we have to check if the MCH CRD existing before adding acheme and filtered cache.

controllers/multiclusterobservability/multiclusterobservability_controller.go
morvencao: done.

controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -127,11 +131,38 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
...
+		// normally there should only one MCH CR in the cluster
+		if len(mchList.Items) == 1 {
+			mch := mchList.Items[0]
+			mchVer := mch.Status.CurrentVersion
+			if mchVer != "" && (mch.Status.CurrentVersion == mch.Status.DesiredVersion) {
+				//read image manifest configmap to be used to replace the image for each component.
+				if _, err = config.ReadImageManifestConfigMap(r.Client, mchVer); err != nil {
+					return ctrl.Result{}, err
+				}
+			}
```
clyang82, line 162: if mch.Status.CurrentVersion != mch.Status.DesiredVersion, just return.

go.mod
```diff
@@ -26,9 +27,22 @@ require (
 )
 
 replace (
+	github.com/metal3-io/baremetal-operator => github.com/openshift/baremetal-operator v0.0.0-20200715132148-0f91f62a41fe
+	github.com/metal3-io/cluster-api-provider-baremetal => github.com/openshift/cluster-api-provider-baremetal v0.0.0-20190821174549-a2a477909c1d
```
clyang82, line 31: Why we need them?

clyang82:
@marcolan018 Please take another look.

controllers/placementrule/placementrule_controller.go
```diff
@@ -122,9 +126,10 @@ func (r *PlacementRuleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	//read image manifest configmap to be used to replace the image for each component.
-	if _, err = config.ReadImageManifestConfigMap(r.Client); err != nil {
-		return ctrl.Result{}, err
-	}
+	// the MultiClusterObservabilityReconciler will read the image manifests, no need to read it again.
```
marcolan018, line 129: to delete readimage logic maybe lead to problem
at starting stage
if placementrule controller reconcile firstly, then there is no image info and that will lead to wrong images used in the manifestwork
The images will not be updated until next reconcile.
in worst cases, the wrong images will be kept for a long time

marcolan018:
after the change, each reconcile needs to read mch crd, cr, then read the manifestimage cm.
if we have some enhancements here to avoid too many k8s calls, is it better or over-engineering?
e.g. 
1. in config, maintain a map[version][images], to avoid read manifestsimage multiple times
2. in mco controller, only enter the mch checking logic only when events on mch detected

go.mod
morvencao: Because the mch operator depends on theses, removing these will report errors.

controllers/multiclusterobservability/multiclusterobservability_controller.go
morvencao: I think we should just skip reading the imahe manifests if mch.Status.CurrentVersion != mch.Status.DesiredVersion. the other steps should still be processed.

controllers/placementrule/placementrule_controller.go
morvencao: Good catch.

morvencao: I added the read image logic back.

morvencao:
> after the change, each reconcile needs to read mch crd, cr, then read the manifestimage cm.
if we have some enhancements here to avoid too many k8s calls, is it better or over-engineering?
e.g.
in config, maintain a map[version][images], to avoid read manifestsimage multiple times
in mco controller, only enter the mch checking logic only when events on mch detected

Updated the watch mch logic a bit to make sure that the controller only read image manifests configmap when the MCH event is detected, also the read options only executed when when the mch is stable(desiredVersion==currentVersion).

morvencao:
/hold
Adding testing for this.

morvencao:
another look? @clyang82 @marcolan018

PR #620, opened 2021-06-24
move proxy from open-cluster-management/rbac-query-proxy for testing

clyang82:
Move proxy from open-cluster-management/rbac-query-proxy for testing


morvencao (commented):
/lgtm

PR #634, opened 2021-06-30
Support configure proxy for thanos and alertmanager

clyang82:
ref: https://github.com/open-cluster-management/backlog/issues/13811

### DIFF
--- a/controllers/multiclusterobservability/observatorium.go
+++ b/controllers/multiclusterobservability/observatorium.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"os"
 	"time"
 
 	obsv1alpha1 "github.com/open-cluster-management/observatorium-operator/api/v1alpha1"
@@ -206,6 +207,9 @@ func newDefaultObservatoriumSpec(mco *mcov1beta2.MultiClusterObservability,
 	obs.Tolerations = mco.Spec.Tolerations
 	obs.API = newAPISpec(mco)
 	obs.Thanos = newThanosSpec(mco, scSelected)
+	if util.ProxyEnvVarsAreSet() {
+		obs.EnvVars = newEnvVars()
+	}
 
 	obs.Hashrings = []*obsv1alpha1.Hashring{
 		{Hashring: "default", Tenants: []string{mcoconfig.GetTenantUID()}},
@@ -218,6 +222,16 @@ func newDefaultObservatoriumSpec(mco *mcov1beta2.MultiClusterObservability,
 		obs.ObjectStorageConfig.Thanos.Key = objStorageConf.Key
 	}
 	return obs
+}
+
+// return proxy variables
+// OLM set these environment variables as a unit
+func newEnvVars() map[string]string {
+	return map[string]string{
+		"HTTP_PROXY":  os.Getenv("HTTP_PROXY"),
+		"HTTPS_PROXY": os.Getenv("HTTPS_PROXY"),
+		"NO_PROXY":    os.Getenv("NO_PROXY"),
+	}
 }
 
 func newAPIRBAC() obsv1alpha1.APIRBAC {
--- a/pkg/rendering/renderer_alertmanager.go
+++ b/pkg/rendering/renderer_alertmanager.go
@@ -4,7 +4,6 @@
 package rendering
 
 import (
-	"os"
 	"strconv"
 
 	v1 "k8s.io/api/apps/v1"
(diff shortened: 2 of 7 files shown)


controllers/multiclusterobservability/observatorium.go
```diff
@@ -220,6 +224,16 @@ func newDefaultObservatoriumSpec(mco *mcov1beta2.MultiClusterObservability,
 	return obs
 }
 
+// return proxy variables
+// OLM set these environment variables as a unit
+func newEnvVars() map[string]string {
+	return map[string]string{
```
marcolan018, line 230: do we need to check one env exists or not before use it?

clyang82: OLM handles these environment variables as a unit. we do not need to check one exists or not.

pkg/rendering/renderer_alertmanager.go
```diff
@@ -83,6 +84,24 @@ func (r *Renderer) renderAlertManagerStatefulSet(res *resource.Resource) (*unstr
...
+			{
+				Name:  "HTTPS_PROXY",
+				Value: os.Getenv("HTTPS_PROXY"),
+			},
+			{
+				Name:  "NO_PROXY",
+				Value: os.Getenv("NO_PROXY"),
+			},
+		}...)
+		log.V(1).Info("Proxy configuration environment variables are set.", "HTTP_PROXY:", os.Getenv("HTTP_PROXY"), "HTTPS_PROXY:", os.Getenv("HTTPS_PROXY"), "NO_PROXY:", os.Getenv("NO_PROXY"))
```
clyang82, line 102: Do not need to set such environments for alertmanager. Alertmanager is using configuration which is defined in `alertmanager.yaml`. the user can update the alertmanager-config secret e.g.:
```
global:
  slack_api_url: xxxx
  http_config:
    proxy_url: http://54.237.28.209:3128
route:
  receiver: default-receiver
  group_by: ['alertname', 'cluster']
  repeat_interval: 30ms
  matchers:
  - severity="critical"
receivers:
```

clyang82:
Here is a document described how I tested - https://github.com/clyang82/blogs/blob/main/Disconnected_environment/Simulate_Disconnected_Environment.md

@bjoydeep @marcolan018 It is ready to go. Thanks.

pkg/rendering/renderer_alertmanager.go
marcolan018: @clyang82 
imo, this is a bad user experience
can mco controller set it automatically based on the env?

bjoydeep: @clyang82 I agree with Marco. Asking the user to ALSO add this is not best. I agree with @marcolan018  - we should add this logic to modify to our MCO controller. BTW can we  check with Simon P on how the platform alertmanager behaves - just for reference.

clyang82: https://github.com/prometheus/alertmanager/issues/1300 the alertmanager stopped working with proxy environment. same as my test results
```
level=error ts=2021-07-02T02:29:46.469Z caller=dispatch.go:309 component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="default-receiver/slack[0]: notify retry canceled after 2 attempts: Post \"<redacted>\": context deadline exceeded"
level=error ts=2021-07-02T02:30:53.882Z caller=dispatch.go:309 component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="default-receiver/slack[0]: notify retry canceled after 2 attempts: Post \"<redacted>\": context deadline exceeded"
```

marcolan018: @clyang82 
I mean mco controller to update alertmanger.yaml content based on the env. If PROXY_URL is available in env and no global.proxy_url set, then inject it.
Either to update the alertmanager-config secret directly, or generate an internal secret which based on it and mco operator env.

clyang82: OpenShift does not mention how to configure alertmanager via proxy - https://docs.openshift.com/container-platform/4.7/monitoring/managing-alerts.html#applying-custom-alertmanager-configuration_managing-alerts. I think we need to update document to tell the user how to configure. the good news is the user who configure the alertmanager should be the same user - IT admin who setup the ACM.

clyang82: @marcolan018 We do not configure any routes by default. if just update with `proxy_url`, it can be easy to overwrite by customer.

bjoydeep: @marcolan018 , @clyang82 and I had a long discussion with the Monitoring team here- https://coreos.slack.com/archives/C0VMT03S5/p1625236195269000?thread_ts=1625195650.257100&cid=C0VMT03S5
it seems that they have decided NOT to automatically use the proxy settings for the monitoring stack because all receivers for example may not need it. In other words it sounds like a design choice OpenShift has made consciously **because it does not work in all cases**. We can of course work to change this behavior - but that as you know will be long drawn process.

marcolan018: @bjoydeep that make sense

PR #636, opened 2021-07-02
add annotation to permit user disable proxy

marcolan018:
in some customer's env, the objectstorage also maybe located in internal network. In that case no proxy is required even if HTTP_PROXY is set by OLM
Add an annotation mco-porxy-disable to let user has the option to disable the PROXY env in mco CR

### DIFF
--- a/controllers/multiclusterobservability/observatorium.go
+++ b/controllers/multiclusterobservability/observatorium.go
@@ -207,7 +207,7 @@ func newDefaultObservatoriumSpec(mco *mcov1beta2.MultiClusterObservability,
 	obs.Tolerations = mco.Spec.Tolerations
 	obs.API = newAPISpec(mco)
 	obs.Thanos = newThanosSpec(mco, scSelected)
-	if util.ProxyEnvVarsAreSet() {
+	if config.IsPorxyRequired(mco.GetAnnotations()) {
 		obs.EnvVars = newEnvVars()
 	}
 
--- a/controllers/multiclusterobservability/observatorium.go
+++ b/controllers/multiclusterobservability/observatorium.go
@@ -207,7 +207,7 @@ func newDefaultObservatoriumSpec(mco *mcov1beta2.MultiClusterObservability,
 	obs.Tolerations = mco.Spec.Tolerations
 	obs.API = newAPISpec(mco)
 	obs.Thanos = newThanosSpec(mco, scSelected)
-	if config.IsPorxyRequired(mco.GetAnnotations()) {
+	if config.IsProxyRequired(mco.GetAnnotations()) {
 		obs.EnvVars = newEnvVars()
 	}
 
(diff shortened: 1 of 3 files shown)


bjoydeep:
Agreed we need this. This is one of the reasons why the Monitoring team did not adopt proxy settings automatically. Annotation provides a more convenient backoff rather than changing the CR.

pkg/config/config.go
```diff
@@ -579,6 +580,27 @@ func IsPaused(annotations map[string]string) bool {
 	return false
 }
 
+// IsPorxyRequired returns true if PROXY env is available and there is no mco-proxy-disable:true annotation in mco instance
+// OLM handles these environment variables as a unit;
+// if at least one of them is set, all three are considered overridden
+// and the cluster-wide defaults are not used for the deployments of the subscribed Operator.
+// https://docs.openshift.com/container-platform/4.6/operators/admin/olm-configuring-proxy-support.html
+func IsPorxyRequired(annotations map[string]string) bool {
```
bjoydeep, line 588: nit/typo- can we name if `IsProxyRequired` please

controllers/multiclusterobservability/observatorium.go
```diff
@@ -207,7 +207,7 @@ func newDefaultObservatoriumSpec(mco *mcov1beta2.MultiClusterObservability,
 	obs.Tolerations = mco.Spec.Tolerations
 	obs.API = newAPISpec(mco)
 	obs.Thanos = newThanosSpec(mco, scSelected)
-	if util.ProxyEnvVarsAreSet() {
+	if config.IsPorxyRequired(mco.GetAnnotations()) {
```
bjoydeep, line 210: nit/typo- can we name if `IsProxyRequired` please

pkg/config/config.go
```diff
@@ -49,6 +49,7 @@ const (
 	AnnotationMCOWithoutResourcesRequests = "mco-thanos-without-resources-requests"
 	AnnotationSkipCreation                = "skip-creation-if-exist"
 	AnnotationCertDuration                = "mco-cert-duration"
+	AnnotationProxyDisable                = "mco-proxy-disable"
```
bjoydeep, line 52: should we call it `mco-proxy-ignore` rather than disable? We are basically telling it to ignore global cluster level setting

bjoydeep:
Thanks @marcolan018 . This looks good to me. However, I will let @clyang82 do another review before we merge it.

clyang82:
@bjoydeep @marcolan018 Does `NO_PROXY` help for this case?

marcolan018:
@clyang82 more details? about how to use NO_PROXY

clyang82:
> @clyang82 more details? about how to use NO_PROXY

Here is an example in my tests
```
- name: HTTP_PROXY
        value: http://54.237.28.209:3128
      - name: HTTPS_PROXY
        value: http://54.237.28.209:3128
      - name: NO_PROXY
        value: .cluster.local,.svc,10.128.0.0/14,127.0.0.1,172.22.0.0/24,172.30.0.0/16,192.168.123.0/24,api-int.obs-china-aws-480-sno-fxw45.dev05.red-chesterfield.com,localhost
```
The customer can add the object storage IP or DNS name into `NO_PROXY` bypass the proxy.

PR #645, opened 2021-07-12
fix overestimation panel on overview dashboard and add required pod phase metric

RobDolares:
Related Issue: https://github.com/open-cluster-management/backlog/issues/9246


marcolan018:
in my 3M3W env, count of time series data of metrics kube_pod_status_phase is 1530. That's a big number and means much impact on the observability perf.
Is this necessary to add, or do we have other options to repalce it?
 @RobDolares @bjoydeep

marcolan018:
don't need to add "kube_pod_status_phase" in allowlist.
other changes lgtm

PR #658, opened 2021-07-20
Adopt ocm-e2e-clusterpool

clyang82:
ref: https://github.com/open-cluster-management/backlog/issues/14323

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,10 +159,6 @@ e2e-tests:
 
 test-e2e-setup:
 	@echo "Seting up E2E Tests environment..."
-ifdef COMPONENT_IMAGE_PIPELINE
-	# override the image for the e2e test
-	@./cicd-scripts/setup-e2e-tests.sh -a install -p $(COMPONENT_IMAGE_PIPELINE)
-endif
 ifdef COMPONENT_IMAGE_NAMES
 	# override the image for the e2e test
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -153,16 +153,12 @@ unit-tests:
...
 
 test-e2e-setup:
 	@echo "Seting up E2E Tests environment..."
-ifdef COMPONENT_IMAGE_PIPELINE
-	# override the image for the e2e test
-	@./cicd-scripts/setup-e2e-tests.sh -a install -p $(COMPONENT_IMAGE_PIPELINE)
-endif
 ifdef COMPONENT_IMAGE_NAMES
 	# override the image for the e2e test
 	@./cicd-scripts/setup-e2e-tests.sh -a install -i $(COMPONENT_IMAGE_NAMES)
```
morvencao, line 164: These changes are trying to make the e2e testing in prow build work, what about the local test?
I think we also need to setup the e2e testing env locally, maybe in future PRs.

clyang82: `COMPONENT_IMAGE_PIPELINE` was introduced to get all images from this pipeline. it is not available in `ocm-e2e-clusterpool`. `COMPONENT_IMAGE_NAMES` is still there for local tests. But you are correct, I need to have a try with local testing.

morvencao (commented):
/lgtm

PR #662, opened 2021-07-21
fix iam name

songleo:
if username contains the number sign '#', we need to replace it with '%23' due to use it in URL parameters
https://github.com/open-cluster-management/backlog/issues/14390


songleo:
use the PR to test in QA env, we can switch the IAM user to admin:

```
ssli@tools(fix-iam-name)$./switch-to-grafana-admin.sh IAM#cqu
User <IAM#cqu> switched to be grafana admin
```

image

PR #663, opened 2021-07-21
Revert "Disable obs addon in openshift 3"

clyang82:
Reverts open-cluster-management/multicluster-observability-operator#649

ref: https://github.com/open-cluster-management/backlog/issues/14284

Couple gaps to support OCP 311:
ocp 311 does not support
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
2. endpoint-metrics-operator cannot get clusterversion in 311.
3. unable to create addon lease
4. unable to update status for observabilityaddon

The proposal solution is to rollback the previous fix to enable obs addon in 311. and add document to describe how to manually create v1beta1 crd in each 311 managed cluster. at least, it can post metrics to hub. the `#2/#3/#4` are still true.


clyang82:
Do not need to revert the previous fix. we have to create another  placementrule to select 311 cluster.

PR #666, opened 2021-07-26
add timeout for alertmgr

songleo:
alertmanager takes 4 minutes to start, so we need to set a timeout for `IntegrityChecking`
https://github.com/open-cluster-management/backlog/issues/14486

### DIFF
--- a/tests/pkg/tests/observability_addon_test.go
+++ b/tests/pkg/tests/observability_addon_test.go
@@ -227,7 +227,10 @@ var _ = Describe("Observability:", func() {
 	})
 
 	JustAfterEach(func() {
-		Expect(utils.IntegrityChecking(testOptions)).NotTo(HaveOccurred())
+		Eventually(func() error {
+			return utils.IntegrityChecking(testOptions)
+			// alertmanager takes 4 minutes to start, so we need to set a timeout for IntegrityChecking
+		}, EventuallyTimeoutMinute*5, EventuallyIntervalSecond*5).Should(Succeed())
 	})
 
 	AfterEach(func() {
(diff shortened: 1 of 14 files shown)


tests/pkg/tests/observability_addon_test.go
```diff
@@ -227,7 +227,10 @@ var _ = Describe("Observability:", func() {
 	})
 
 	JustAfterEach(func() {
-		Expect(utils.IntegrityChecking(testOptions)).NotTo(HaveOccurred())
+		Eventually(func() error {
+			return utils.IntegrityChecking(testOptions)
+			// alertmanager takes 4 minutes to start, so we need to set a timeout for IntegrityChecking
+		}, EventuallyTimeoutMinute*5, EventuallyIntervalSecond*5).Should(Succeed())
```
clyang82, line 233: I do not understand why we need to add `Eventually` in `JustAfterEach`. If alertmanager takes 4 minutes to start, we need to wait for enough time in where triggers.

songleo: Because I found that even though the MCO status is ready, I found this error in many case files. for example:

- observability_retention_test.go
- observability_addon_test.go
- observability_alert_test.go
- observability_grafana_test.go 

 and we run the tests randomly, so I add this timeout to each file.

clyang82: We need to do more investigation to understand why it happened. As you know that, we have such code to ensure all components are ready https://github.com/open-cluster-management/multicluster-observability-operator/blob/2882f636584360e4644de290acdc4b91742cc433/tests/pkg/tests/observability_install_test.go#L148-L157

PR #669, opened 2021-07-27
clean up the operand names map for cr deletion.

morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/14560

This PR is to fix the inconsistent operand names after MCO is upgraded from release-2.2 and then create new MCO CR.

Need to be cherry-picked to release-2.3 Z stream.


clyang82:
/lgtm

@morvencao please create a PR in release-2.3 and hold it until release-2.3 is open for business.

PR #675, opened 2021-07-29
Add debug information for troubleshooting

clyang82:
I found that there are 2 replicasets for `grafana` `rbac-query-proxy` component. One rs is using the default image, the other is using the desired image. alertmanager should have the same behaviour. It can be a root cause of https://github.com/open-cluster-management/backlog/issues/14486
so add more debug info to see what happened


morvencao (commented):
/lgtm

morvencao (commented):
/lgtm

PR #676, opened 2021-07-29
handle observatorium cr for pvc resize forbidden.

morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/14103


t
```diff
@@ -0,0 +1,81 @@
+[1mdiff --git a/operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go b/operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go[m
```
clyang82, line 1: test code?

operators/multiclusterobservability/controllers/multiclusterobservability/observatorium.go
```diff
@@ -96,6 +101,14 @@ func GenerateObservatoriumCR(
 	newSpec := observatoriumCR.Spec
 	oldSpecBytes, _ := yaml.Marshal(oldSpec)
 	newSpecBytes, _ := yaml.Marshal(newSpec)
+
+	log.Info("handling storage resize for observatorium", "isResizeForbidden", isResizeForbidden)
+	// isResizeForbidden will reset after operator restart, double check with scAllowVolumeExpansion
+	if isResizeForbidden || !scAllowVolumeExpansion {
+		rollbackStorageResizeForObservatorium(&oldSpec, &newSpec)
```
clyang82, line 108: Need to apply the same logic for alertmanager.

Another point: How about just rollback the changes for mco cr directly if the storageclass cannot support resize feature? thoughts?

operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -579,11 +597,50 @@ func updateStorageSizeChange(c client.Client, matchLabels map[string]string, sto
...
+					}
+				}
+				return !isThereNotReady, nil
+			})
+			if err != nil {
+				return err
+			}
+		}
+	*/
+
```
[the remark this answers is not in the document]
morvencao: Yes, we can remove this.
But I have to mentioned that adding wait logic for PVC resize will result in operatorium-operator report error for update statefulset, that's why I commented this.

operators/multiclusterobservability/controllers/multiclusterobservability/observatorium.go
morvencao: In my opinion, we should not revert the MCO CR without notifying end-user.
But the currently solution is not the perfect, the user can only check the MCO CR logs for the reason why the PVC is not resized.

The best solution should be validating resize request with validation webhook, so that it can return the resize error and reason directly to the end user when they apply the MCO CR.

morvencao: Besides, it will be complicated if we use the revision to rollback the MCO CR, we also need to consider the other changes, for example, retention change, together with the storage size change.

t
morvencao: removed this file.

operators/multiclusterobservability/controllers/multiclusterobservability/observatorium.go
clyang82: Yes. I agree that we should not revert any user's changes without notifying. What is effort to use validation webhook? that is right way I think. I remembered we already exposed webhook api in 2.3.

clyang82: > Besides, it will be complicated if we use the revision to rollback the MCO CR, we also need to consider the other changes, for example, retention change, together with the storage size change.
> 
> For alertmanager, the statefulset rollback logic will be much more complicated than the observatorium CR, I remove that part because it will cause other issues(infinite reconcile loop, storage size mess up)

Yes. I have the same concern about infinite reconcile loop.

morvencao: We can add validation webhook for 2.4, what about 2.3?

clyang82: How about keep it as a known issue in 2.3?

morvencao:
close this in favor of https://github.com/open-cluster-management/multicluster-observability-operator/pull/681

PR #677, opened 2021-07-30
enable some e2e test case

songleo:
https://github.com/open-cluster-management/backlog/issues/14628

### DIFF
--- a/tests/pkg/tests/observability_grafana_dev_test.go
+++ b/tests/pkg/tests/observability_grafana_dev_test.go
@@ -19,7 +19,6 @@ var _ = Describe("Observability:", func() {
 	// Do not need to run this case in canary environment
 	// If we really need it in canary, ensure the grafana-dev-test.sh is available in observability-e2e-test image and all required commands exist
 	It("[P1][Sev1][Observability][Integration] Should run grafana-dev test successfully (grafana-dev/g0)", func() {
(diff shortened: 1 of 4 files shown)


tests/pkg/tests/observability_grafana_dev_test.go
```diff
@@ -19,8 +19,7 @@ var _ = Describe("Observability:", func() {
 	// Do not need to run this case in canary environment
 	// If we really need it in canary, ensure the grafana-dev-test.sh is available in observability-e2e-test image and all required commands exist
 	It("[P1][Sev1][Observability][Integration] Should run grafana-dev test successfully (grafana-dev/g0)", func() {
-		Skip("Skip grafana-dev temporarily. need more tests")
-		cmd := exec.Command("../../cicd-scripts/grafana-dev-test.sh")
+		cmd := exec.Command("../../grafana-dev-test.sh")
```
clyang82, line 22: We do not have `grafana-dev-test.sh` in canary environment. in other words, if you want to have it test in canary env, you need to add it into `Dockerfile`. otherwise, skip for canary test.

songleo: we have the keyword `Integration ` in this case. so we can skip this test in the canary test. right?

clyang82: Yes.

PR #685, opened 2021-08-05
update components resource request

songleo:
https://github.com/open-cluster-management/backlog/issues/14901

### DIFF
--- a/examples/mco/e2e/v1beta2/observability.yaml
+++ b/examples/mco/e2e/v1beta2/observability.yaml
@@ -100,8 +100,8 @@ spec:
         cpu: 200m
         memory: 700Mi
       requests:
-        cpu: 200m
-        memory: 200Mi
+        cpu: 10m
+        memory: 100Mi
   storageConfig:
     alertmanagerStorageSize: 1Gi
     compactStorageSize: 1Gi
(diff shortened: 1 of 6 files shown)


clyang82:
We observed from canary environment and found that the receiver cannot be started due to `3 Insufficient cpu`. when log into the canary environment and we can see the cpu requests from three worker nodes as follows:
```
cpu  3379m (96%) 
cpu  3225m (92%) 
cpu  3477m (99%)
```
ideally, the receiver should be dispatched to the worker node which cpu requests is 96%. but it does not meet the requests. other 2 worker nodes also cannot meet the requests. so the receive-3 is pending. 

Do more check, we found that `open-cluster-management-addon-observability  metrics-collector-deployment-6d8d649845-z9n2z                 200m (5%)     200m (5%)      200Mi (1%)       700Mi (4%)      16m` is allocated into worker node (96%). if metrics-collector-deployment is dispatched to other worker node, the receive can be started successfully. That is why we meet this problem from time to time.

Do some analysis, I think the observatorium-operator and endpoint-observability-operator requests more CPU. for the observatorium-operator, it used to install thanos components and only watch observatorim CR. so no need to request 100m CPU. endpoint-observability-operator is a quit simple one. do not need to request 100m as well.

PR #688, opened 2021-08-06
make mco wait for mch ready before reconciling and reading image manifests

morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/14817

### DIFF
--- a/operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
+++ b/operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"time"
 
 	"github.com/go-logr/logr"
 	routev1 "github.com/openshift/api/route/v1"
@@ -130,8 +131,9 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, err
 	}
 
+	// check if the MCH CRD exists
 	mchCrdExists, _ := r.CRDMap[config.MCHCrdName]
-	if req.Name == config.MCHUpdatedRequestName && mchCrdExists {
+	if mchCrdExists {
 		mchList := &mchv1.MultiClusterHubList{}
 		mchistOpts := []client.ListOption{
 			client.InNamespace(config.GetMCONamespace()),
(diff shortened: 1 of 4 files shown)


operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -130,8 +131,9 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, err
 	}
 
+	// check if the MCH CRD exists
 	mchCrdExists, _ := r.CRDMap[config.MCHCrdName]
-	if req.Name == config.MCHUpdatedRequestName && mchCrdExists {
+	if mchCrdExists {
```
marcolan018, line 136: how about flow like below?

if mchCrdExists && imageList is empty {
    if req.Name == config.MCHUpdatedRequestName {
        // read mch CR and then read image list
    } else {
        return and reconcile later
    }
}

this flow can avoid duplicate checking on mch, also fix the problem we met before

morvencao: So we will only reconcile the `MCHUpdated` request? 
For the first reconcile, it's OK. But what about the changes of the watching objects(configmap, secret....)?

marcolan018: we just need that reconcile to read the image info and store in memory. 
in the following reconciles, they can use the value directly, correct?

morvencao: the only concern is that the image manifest change after that may not be caught.

morvencao: I added ` imageManifestsMap is empty` check to avoid read configmap for each reconcile.
The `if req.Name == config.MCHUpdatedRequestName` is not added because if the MCO is recreated, the operator will not catch the `MCHUpdatedRequest` event.

morvencao:
Basically, we need to handle these cases:

1. Fresh install without MCH CRD - never read the image manifests configmap
2. Fresh install with MCH CRD - read the image manifests configmap
3. Upgrade with MCH CRD - read the image manifests configmap
4. MCO operator is restarted after the MCH is stable - I did some experiments in my env, the mch update event will not be caught any more, the only event caught is mch create event by the placementrule controller, the reason is that placementrule controller is started in the multiclusterobservability_controller reconcile loop:
```
root@mcdev:~# oc -n open-cluster-management logs multicluster-observability-operator-675474dbc5-45rgv | grep "====="
2021-08-09T04:32:13.024Z	INFO	controller_multiclustermonitoring	=============== catch the mch CREATE event 	{"mch name": "multiclusterhub"}
root@mcdev:~# oc -n open-cluster-management logs multicluster-observability-operator-675474dbc5-45rgv | grep "====="
2021-08-09T04:32:13.024Z	INFO	controller_multiclustermonitoring	=============== catch the mch CREATE event 	{"mch name": "multiclusterhub"}
root@mcdev:~# oc -n open-cluster-management logs multicluster-observability-operator-675474dbc5-45rgv | grep "mch-update"
2021-08-09T04:32:15.915Z	INFO	controller_placementrule	Reconciling PlacementRule	{"Request.Namespace": "open-cluster-management", "Request.Name": "mch-updated-request"}
```
5. MCO CR is deleted and then recreated - read the image manifests configmap again and reuse the last read imageMainifestsMap(we don't clean the imageMainifestsMap when the MCO CR is deleted, any issue?)

morvencao:
Another look? @marcolan018 @clyang82

clyang82:
2 points I have:

1. How about move the logic of setting imagemanifests map into `CreateFunc` and `UpdateFunc`? And we can extract this logic into a new function so that it can be reused by 2 controllers.
2. in 2 controllers reconcilier, just wait for the length of imagemanifests map is equal to 1. 

Do they make sense?

PR #691, opened 2021-08-10
Register debug to http.DefaultServeMux

clyang82:
If you're not using the default mux you just have to register any/all of those you want with whatever mux you're using. for our case, we can simply hand off any request prefixed with /debug/ to the http.DefaultServeMux


morvencao (commented):
/lgtm

PR #695, opened 2021-08-12
Adding SLI metrics to default allowlist

dislbenn:


marcolan018:
@dislbenn 
The two new-added metrics have too many time series data and will lead to much increase usage for memory in hub side. We should avoid to add such metrics in allowlist.
If it's required by thanos recording rule, you can consider to add the recoding rule in allowlist, but not to add the original metrics

marcolan018 (changes requested):
don't add metrics which has large number of time series data

PR #715, opened 2021-08-23
replace route in kind testing env.

morvencao:
This PR tries to replace the openshift Route with service's DNS for testing in KinD cluster.

### DIFF
--- a/examples/minio/minio-pvc.yaml
+++ b/examples/minio/minio-pvc.yaml
@@ -6,7 +6,7 @@ metadata:
   name: minio
   namespace: open-cluster-management-observability
 spec:
-  storageClassName: gp2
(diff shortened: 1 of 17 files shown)


examples/minio/minio-pvc.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   name: minio
   namespace: open-cluster-management-observability
 spec:
-  storageClassName: gp2
+  storageClassName: standard
```
morvencao, line 9: @songleo the default storageclass in openshift is `gp2`, so this change will result PV can't be created in e2e env.
I know you need this in KinD env, can we change in script for e2e in KinD?

PR #740, opened 2021-09-03
fix required fields for mco.

morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/15736

There is an issued mco CR that causes mco operator crashing, after investigation, the `storageConfig` was mistaken for `storageConfigObject`.

By design, in v1beta2 mco API, the `storageConfig` and `metricObjectStorage` are required fields, so in that case, the CR should be refused to applied in to the cluster.

However, unfortunately, the required fields are not in the generated MCO CRD, the reason is:

> The omitempty tag in Foo string `json:"foo,omitempty"` had more precedence than the kubebuilder marker +kubebuilder:validation:Required therefore making the resultant generated crd field to be optional.
The +kubebuilder:validation:Required marker does works as desired when omitempty is not used.

This PR remove the `omitempty` tag from go type definition so that the `// + required` can actually take effect.


morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/15736

PR #745, opened 2021-09-07
Update to use ResourceVersion instead of Generation for MCO CR Update

clyang82:
If the user adds new annotation (e.g.: mco-grafana-image), it will not trigger reconciler. That is because the generation is not changed.


morvencao (commented):
/lgtm

PR #752, opened 2021-09-13
update label_values() in dashboards to avoid heavy query

marcolan018:
https://github.com/open-cluster-management/backlog/issues/15916


morvencao (commented):
thanks
/lgtm

PR #754, opened 2021-09-14
replace image for managedcluster resource for mch update event.

morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/15996

- add OR logic for MCH resource update check: the original checks may cause upgrade issue about image replacement.


clyang82:
As we discussed offline, please add more UT for this case. Thanks.

morvencao:
@clyang82 PTAL

PR #755, opened 2021-09-15
[wip]fix metrics check issue

songleo:
In canary test env, we just check the obs addon status in the first managedcluster, for the other managedclusters, we can not to check the addon status. if the addon status is not ready, we can not check the metrics from these managedclusters. 

 fix https://github.com/open-cluster-management/backlog/issues/16007

### DIFF
--- a/tests/pkg/tests/observability_grafana_test.go
+++ b/tests/pkg/tests/observability_grafana_test.go
@@ -4,8 +4,6 @@
 package tests
 
 import (
-	"fmt"
-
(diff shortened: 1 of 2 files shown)


clyang82:
As we discussed, we should check the oba status in each managed cluster namespace to ensure they are ready. that means we wonder to continue do testing in all managed cluster.

clyang82:
close it in favor of https://github.com/open-cluster-management/multicluster-observability-operator/pull/771

PR #760, opened 2021-09-15
in non-ocp, disable useless metrics scrape in prometheus

marcolan018:
https://github.com/open-cluster-management/backlog/issues/16110


clyang82:
@marcolan018 Do we need to update `scrape_interval` to keep consistent with `interval` of `observabilityAddonSpec`?

marcolan018:
@clyang82 not included in this pr
in fact, we may not update scrape interval for prometheus. Some testing still in progress

PR #769, opened 2021-09-17
use topk for cpu/memory optimization/utilization panels.

morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/15916

### DIFF
--- a/operators/multiclusterobservability/manifests/base/alertmanager/alert_rules.yaml
+++ b/operators/multiclusterobservability/manifests/base/alertmanager/alert_rules.yaml
@@ -48,3 +48,33 @@ data:
           labels:
             slo: "API Server Request Duration"
             target: 0.9900
+      - name: grafana-dashboard
+        rules:
+        - expr: sum(cluster:kube_pod_container_resource_requests:cpu:sum) by (cluster) / sum(kube_node_status_allocatable{resource="cpu"}) by (cluster)
+          record: cluster:cpu_requested:ratio
+          labels:
+            usage: grafana-dashboard
+        - expr: 1 - avg(rate(node_cpu_seconds_total{mode="idle"}[5m])) by (cluster)
+          record: cluster:cpu_utilized:ratio:5m
+          labels:
+            usage: grafana-dashboard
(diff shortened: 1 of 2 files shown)


morvencao:
@marcolan018 @clyang82 Another look? as suggested, I used the RecordingRules in Thanos Ruler to accelerate the queries for grafana dashboards.

operators/multiclusterobservability/manifests/base/alertmanager/alert_rules.yaml
```diff
@@ -48,3 +48,33 @@ data:
           labels:
             slo: "API Server Request Duration"
             target: 0.9900
+      - name: grafana-dashboard
```
marcolan018, line 51: is it possible to use recording rule in managed cluster side? do some aggregation in managed cluster to create new cluster-level metrics, then in hub side, when we try to query among all clusters, the query will be lightweight.
I have this concern because if those queries are heavy, to add them as recording rule in thanos rule means those queries will be performed every 30 seconds, that will lead to much resource consumption.
@morvencao @clyang82 comments?

morvencao: the reason why I added these to the thanos side insteafd of managed cluster side is that the queries need do aggregation/filter with the cluster label. I didn't test the overhead to the hub side for these added recordingrules.

operators/multiclusterobservability/manifests/base/grafana/dash-acm-clusters-overview.yaml
```diff
@@ -1327,7 +1228,7 @@ data:
           "steppedLine": false,
           "targets": [
             {
-              "expr": "topk(5, (1 - avg(rate(node_cpu_seconds_total{mode=\"idle\"}[$__rate_interval])) by (cluster)))",
+              "expr": "topk(5, (1 - avg(rate(node_cpu_seconds_total{mode=\"idle\"}[$__interval])) by (cluster)))",
```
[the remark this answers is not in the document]
morvencao: fixed this.
I missed this when I rebased my code changes.

morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/15160

PR #770, opened 2021-09-18
add custom image registry feature for mangedcluster

songleo:
https://github.com/open-cluster-management/backlog/issues/15917


operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -276,10 +280,41 @@ func createManifestWorks(c client.Client, restMapper meta.RESTMapper,
...
+		images := imageListConfigMap.Data
+		for key, oldImage := range images {
+			newImage, err := imageRegistryClient.Cluster(clusterName).ImageOverride(oldImage)
+			if err == nil {
+				imageListConfigMap.Data[key] = newImage
+			}
+		}
+	}
+
+	if pullSecret != nil {
```
marcolan018, line 312: is it possible the custom pull sectet has conflict with the default  pull secret?

operators/multiclusterobservability/controllers/placementrule/customize_img.go
```diff
@@ -0,0 +1,25 @@
...
+	imageregistryv1alpha1 "github.com/open-cluster-management/multicloud-operators-foundation/pkg/apis/imageregistry/v1alpha1"
+	"github.com/open-cluster-management/multicloud-operators-foundation/pkg/helpers/imageregistry"
+)
+
+var (
+	managedClusterImageRegistry = map[string]string{}
+)
+
+func updateManagedClusterImageRegistry(obj client.Object) {
+	if imageReg, ok := obj.GetLabels()[imageregistryv1alpha1.ClusterImageRegistryLabel]; ok {
```
marcolan018, line 18: do we need to add "delete" logic here?
Assume one managed cluster has custom registry firstly, after some time, delete it.

operators/multiclusterobservability/main.go
```diff
@@ -127,6 +128,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := imageregistryv1alpha1.AddToScheme(scheme); err != nil {
```
marcolan018, line 131: do we need to use filteredcache to handle imageregistry resource?

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
marcolan018: a little confused by the logic for the custom pull secret.
the customPullSecret will use the same name as the original pullSecret, and replace the original one? Or co-exist with different name.
Also, in current logic, pullSecret is a global var, if it's updated in one manifestwork, seems it will affect the following ones?

marcolan018 (changes requested):

operators/multiclusterobservability/controllers/placementrule/customize_img.go
songleo: @zhiweiyin318 Can we update it automatically when disabling custom registry from managedcluster? I remember that seems we can not update it automatically.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
songleo: @marcolan018 thanks for your suggestions, let me check it.

operators/multiclusterobservability/main.go
songleo: @marcolan018 discuss with @morvencao, we do not need filteredcache for imageregistry resource.

morvencao: Looks like the `imageregistry` resource don't have fix label and field for us to filter.
Also given that the `imageregistry` resource should be at most one in one cluster, so we don't need to filter it for cache, IMO.

PR #774, opened 2021-09-23
Combine to use one query per panel

clyang82:
ref: https://github.com/open-cluster-management/backlog/issues/15916


morvencao (commented):
/lgtm

PR #775, opened 2021-09-23
Use default snapshot for all our images

clyang82:
Update to use default snapshot for all our owned images. We pass the `mco-imageTagSuffix` in e2e test. but the grafana is still using the `quay.io/openshift/origin-grafana:4.8.0`. This PR is to define the default `DefaultImgTagSuffix` only in config.go. so we can use this default value in case cannot replace the correct images.


morvencao (commented):
/lgtm

clyang82:
Remove support for loading and deleting dashboard by slug in grafana 8.x https://github.com/grafana/grafana/pull/35104
that is why the `grafana-dev` test can be passed. cannot find the grafana dashboard information by using `127.0.0.1:3001/api/dashboards/db/$dashboard_name`

morvencao:
@clyang82 Is this ready to merge?

PR #779, opened 2021-09-27
Updated SLO dashboard label and remove SLO name from recording rule

dislbenn:
This PR will remove the `slo` label from the recording rule. The SLO name is already displayed within the dashboard; therefore, in this contexts, it's not needed.

Related issue: https://github.com/open-cluster-management/backlog/issues/16542


dislbenn:
The name is also in the recording rule name, so no need to re-add it again as a label. We can always re-add the label if we add more OOTB SLO recording rules; however, for now, it's not needed.

@marcolan018 can I have a quick review on this PR? 🙂

PR #789, opened 2021-10-05
[WIP] Add rules for policy_governance_info metric

JustinKuli:
This metric would have too many time series for us to want to keep all
of them: the total time series is `p * (c + 1)` where `p` is the number
of policies, and `c` is the number of clusters.

These rules aggregate the total and noncompliant policies by cluster and
by policy, so the number of time series from this will only be
`2*p + 2*c`.

Refs:
 - https://github.com/open-cluster-management/backlog/issues/16198

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>


JustinKuli:
/cc @marcolan018

Can I get your opinion on adding these rules?

operators/multiclusterobservability/manifests/base/config/metrics_allowlist.yaml
```diff
@@ -162,4 +162,12 @@ data:
       - record: cluster:kube_pod_container_resource_requests:memory:sum
         expr: sum(sum(sum(kube_pod_container_resource_requests{resource=\"memory\"}) by (pod,namespace,container) * on(pod,namespace) group_left(phase) max(kube_pod_status_phase{phase=~\"Running|Pending|Unknown\"} >0) by (pod,namespace,phase)) by (pod,namespace,phase))
       - record: sli:apiserver_request_duration_seconds:trend:1m
-        expr: sum(increase(apiserver_request_duration_seconds_bucket{apiserver=\"kube-apiserver\",le=\"1\",verb=~\"POST|PUT|DELETE|PATCH\"}[1m])) / sum(increase(apiserver_request_duration_seconds_count{apiserver=\"kube-apiserver\",verb=~\"POST|PUT|DELETE|PATCH\"}[1m]))
\ No newline at end of file
+        expr: sum(increase(apiserver_request_duration_seconds_bucket{apiserver=\"kube-apiserver\",le=\"1\",verb=~\"POST|PUT|DELETE|PATCH\"}[1m])) / sum(increase(apiserver_request_duration_seconds_count{apiserver=\"kube-apiserver\",verb=~\"POST|PUT|DELETE|PATCH\"}[1m]))
+      - record: cluster:policy_governance_info:propagated_count
```
marcolan018, line 166: is metrics policy_governance_info available in all clusters or only in hub?
how many time series data for metrics policy_governance_info{type="propagated"} for one cluster?

if this metrics does not lead to much time series data, it's not recommended to use recording rule becuase each recording rule will add one addtional query call in managed cluster side for each scrape loop

marcolan018: @JustinKuli 
if the metrics only in hub side, it's not necessary to use recording rule to handle it. You can just simply include policy_governance_info in allowlist directly.

PR #811, opened 2021-11-03
Update apiserver_request_duration_seconds recording rules

clyang82:
ref: https://github.com/open-cluster-management/backlog/issues/17656


marcolan018 (commented):
I think we should also update the "expr" in dashboard which use the recording rule metrics

PR #819, opened 2021-11-09
add e2e case for alert forward from managed clusters.

morvencao:
Add e2e case for alert forward from openshift clusters, by checking the `Watchdog` alert are forwarded to the hub, because `Watchdog` alert is always triggered, it is used to make sure the alert trigger is working. 

For *KS, add `Watchdog` alertrule to make sure the entire alerting pipeline is functional.

### DIFF
--- a/tests/pkg/tests/observability_alert_test.go
+++ b/tests/pkg/tests/observability_alert_test.go
@@ -6,7 +6,6 @@ package tests
 import (
 	"context"
 	"crypto/tls"
-	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -247,7 +246,7 @@ var _ = Describe("Observability:", func() {
 		klog.V(3).Infof("Successfully deleted CM: thanos-ruler-custom-rules")
 	})
 
-	It("[P2][Sev2][Observability][Integration] Should have alert named Watchdog forwarded to alertmanager generated (alert/g0)", func() {
+	It("[P2][Sev2][Observability][Integration] Should have alert named Watchdog forwarded to alertmanager (alert/g0)", func() {
 		amURL := url.URL{
 			Scheme: "https",
 			Host:   "alertmanager-open-cluster-management-observability.apps." + testOptions.HubCluster.BaseDomain,
@@ -257,14 +256,9 @@ var _ = Describe("Observability:", func() {
 		q.Set("filter", "alertname=Watchdog")
 		amURL.RawQuery = q.Encode()
 
-		caCrt, err := utils.GetRouterCA(hubClient)
-		Expect(err).NotTo(HaveOccurred())
-		pool := x509.NewCertPool()
-		pool.AppendCertsFromPEM(caCrt)
-
 		client := &http.Client{
 			Transport: &http.Transport{
-				TLSClientConfig: &tls.Config{RootCAs: pool},
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // skip CA verify because CA secret not found in KinD env.
 			},
 		}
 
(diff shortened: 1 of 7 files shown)


tests/pkg/tests/observability_alert_test.go
```diff
@@ -237,6 +247,83 @@ var _ = Describe("Observability:", func() {
 		klog.V(3).Infof("Successfully deleted CM: thanos-ruler-custom-rules")
 	})
 
+	It("[P2][Sev2][Observability][Stable] Should have alert named Watchdog forwarded to alertmanager generated (alert/g0)", func() {
```
clyang82, line 250: Suggest to use [Integration] firstly. we can move it to stable once it is stable.

morvencao: done.

morvencao:
@clyang82 another look? New commits cover the *KS.

tests/pkg/utils/mco_managedcluster.go
```diff
@@ -71,3 +72,85 @@ func ListManagedClusters(opt TestOptions) ([]string, error) {
...
+		if labels != nil {
+			vendorStr := ""
+			if vendor, ok := labels["vendor"]; ok {
+				vendorStr = vendor.(string)
+			}
+			obsControllerStr := ""
+			if obsController, ok := labels["feature.open-cluster-management.io/addon-observability-controller"]; ok {
+				obsControllerStr = obsController.(string)
+			}
+			if (vendorStr == "GKE" || vendorStr == "EKS" || vendorStr == "AKS") && obsControllerStr == "available" {
```
clyang82, line 143: we support all *KS. so here suggest to use `vendorStr != `OpenShift`

morvencao: done

morvencao:
I found that in e2e-kind testing env, the managedclusters only have `vendor` label, no other labels, so `ListManagedClusters` function with `obs-addon-controller == avavilable` will always return empty list, so I reverted the `obs-addon-controller` label check.

clyang82:
> I found that in e2e-kind testing env, the managedclusters only have `vendor` label, no other labels, so `ListManagedClusters` function with `obs-addon-controller == avavilable` will always return empty list, so I reverted the `obs-addon-controller` label check.

@elgnay Could you share why there is no `feature.open-cluster-management.io/addon-observability-controller` label for KinD cluster? is it by design? Thanks.

morvencao:
@clyang82 Maybe we need to use https://github.com/open-cluster-management/registration-operator release-2.4 instead of release 2.3, not sure if that's is a new feature in 2.4.

morvencao:
After upgrading https://github.com/open-cluster-management/registration-operator to `release-2.4`, the managedcluster label still only has `vendor`:

```
$ kubectl get managedcluster --show-labels 
NAME            HUB ACCEPTED   MANAGED CLUSTER URLS   JOINED   AVAILABLE   AGE     LABELS
local-cluster   true           https://localhost      True     True        7m59s   vendor=GKE
```

elgnay:
> After upgrading https://github.com/open-cluster-management/registration-operator to `release-2.4`, the managedcluster label still only has `vendor`:
> 
> ```
> $ kubectl get managedcluster --show-labels 
> NAME            HUB ACCEPTED   MANAGED CLUSTER URLS   JOINED   AVAILABLE   AGE     LABELS
> local-cluster   true           https://localhost      True     True        7m59s   vendor=GKE
> ```

Yes, it's a feature of 2.4 release. The addon labels are generated in my Kind env. Please make sure
1. the registration-hub-controller is running;
2. the addon CR is created in the cluster namespace;

PR #820, opened 2021-11-09
fix codesmell & vulnerabilities

songleo:


morvencao (commented):
overall look good to me, but I suggest to add linter(rg, [golangci_lint](https://github.com/golangci/golangci-lint)) to find the issue before the PR merged.
We can do it in followup PR.

/lgtm

PR #825, opened 2021-11-15
update imagePullPolicy to IfNotPresent

songleo:
fix https://github.com/open-cluster-management/backlog/issues/17910


songleo:
maybe we should have a doc issue to update related doc. let me check if we need it.

PR #826, opened 2021-11-15
Updated slo recording rule metric label filter

dislbenn:
Related Issue: https://github.com/open-cluster-management/backlog/issues/17736


marcolan018:
@dislbenn 
I checked one GKE cluster, in the metrics it collected, there is also no label named as apiserver

Screen Shot 2021-11-16 at 10 40 25 AM

dislbenn:
> @dislbenn I checked one GKE cluster, in the metrics is collected, there is also no label named as apiserver
> 
> Screen Shot 2021-11-16 at 10 40 25 AM

@marcolan018 Correct. For the changes made to this PR, the metrics are filtering the results based upon the `job` label and the `service` label. The job is using the value `apiserver` and the service is using the value `Kubernetes`. The thought behind the changes are that the GKE and the other *ks clusters should have those labels available to them. The `local-cluster` should also have those labels, so the results should return each of the clusters. 

Should we prioritize changing the labels that are being filtered for? The only other option I can think of is to add the `apiserver` value to each *ks cluster, so it can be included by default.

marcolan018:
@dislbenn 
oh, that make sense.

/approve
/lgtm

PR #842, opened 2021-12-01
fix missing data with recording rule from thanos ruler.

morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/18149

We have tried plenty of cases for different `lookback-delta` value, smaller value means less tolerance for latency.
After many experiments, we found the default value `5m` can deal with most of cases, so we set it to at least `5m`.
If `5m` lookback-delta still can't make that, then it means that the network should be broken for a long time, then we can't blame the ruler.


clyang82:
@marcolan018 PTAL

marcolan018:
But in the issue #18149, I think the default scrape interval is used. And in such case, the lookback-delta is 10 minutes. User still run into this problem.

clyang82:
@marcolan018 I remembered this environment is a upgrade environment. in other words, the interval is 60s instead of 300s. Let us have a talk in our scrum meeting soon.

operators/multiclusterobservability/controllers/multiclusterobservability/observatorium.go
```diff
@@ -377,6 +377,7 @@ func newRuleSpec(mco *mcov1beta2.MultiClusterObservability, scSelected string) o
 	} else {
 		ruleSpec.Retention = mcoconfig.RetentionInLocal
 	}
+	ruleSpec.EvalInterval = fmt.Sprintf("%ds", mco.Spec.ObservabilityAddonSpec.Interval)
```
clyang82, line 380: I think we need to expose it in MCO CRD. so here if there is no defined in MCO CR, then set mco.Spec.ObservabilityAddonSpec.Interval as default value. if so, use the customized value.

morvencao: done.

morvencao:
Another look? @clyang82

PR #844, opened 2021-12-03
add default metrics test

songleo:
fix https://github.com/open-cluster-management/backlog/issues/17909


tests/ignored-metric-list
```diff
@@ -0,0 +1,79 @@
+authenticated_user_requests
```
clyang82, line 1: You may create a sub-folder `tests/pkg/testdata` to store this ignore-metrics-list.

tests/pkg/utils/client.go
```diff
@@ -31,7 +31,7 @@ func GetKubeClientDynamic(opt TestOptions, isHub bool) dynamic.Interface {
 	if !isHub && len(opt.ManagedClusters) > 0 {
 		url = opt.ManagedClusters[0].ClusterServerURL
 		kubeConfig = opt.ManagedClusters[0].KubeConfig
-		kubeContext = ""
+		kubeContext = opt.ManagedClusters[0].KubeContext
```
clyang82, line 34: It should be addressed in your previous PR. correct?

songleo: yes, let me rebase it, thanks.

PR #866, opened 2021-12-28
bump prom dependency.

morvencao:
ref: https://github.com/open-cluster-management/backlog/issues/18558


morvencao:
Looks like we need to upgrade kustomize lib to v4.

PR #879, opened 2022-01-21
Support custom certs for the object store and test with tls-enabled m…

clyang82:
Support custom certs for the object store and test with tls-enabled minio

https://github.com/stolostron/backlog/issues/19100


clyang82:
blocked by https://github.com/thanos-io/thanos/pull/5096

clyang82:
ACM cannot be deployed successfully. Have asked in devops channel for help...

examples/minio-tls/thanos-object-storage.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+data:
```
marcolan018, line 2: for example secret, can we use "stringData", but not to use "data" to hide all details?

operators/multiclusterobservability/api/shared/multiclusterobservability_shared.go
```diff
@@ -39,6 +39,12 @@ type PreConfiguredStorage struct {
 	// Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
 	// +required
 	Name string `json:"name"`
+	// TLS secret contains the custom certificate for the object store
+	// +optional
+	TLSSecretName string `json:"tlsSecretName,omitempty"`
+	// TLS secret mount path in thanos store/ruler/compact/receiver
+	// +optional
+	TLSSecretMountPath string `json:"tlsSecretMountPath,omitempty"`
```
marcolan018, line 47: is it necessary to expose "mountpath" to end user?

examples/minio-tls/thanos-object-storage.yaml
clyang82: Yes. that makes more sense to show the content of secret directly.

operators/multiclusterobservability/api/shared/multiclusterobservability_shared.go
clyang82: Yes. I think it is necessary to expose "mountpath" to end user, because you have to configure the full path of certs in objectstore. it is directly using by thanos components. for example: `ca_file: /etc/minio/certs/ca.crt`. the store component find this file and use this ca.crt. if user does not support `mountpath`, we cannot mount the cert secret into the correct path. Does it make sense?

clyang82: Parse the `ca_file` to get the mountpath from object storage secret. Thanks @marcolan018 for your comments.

marcolan018 (approved):
/lgtm

PR #886, opened 2022-02-04
2.4 - delete obs addon when enableMetrics is set to false

subbarao-meduri:
related: [#19015](https://github.com/stolostron/backlog/issues/19015)

### DIFF
--- a/tests/pkg/utils/mco_oba.go
+++ b/tests/pkg/utils/mco_oba.go
@@ -31,6 +31,19 @@ func CheckOBAStatus(opt TestOptions, namespace, status string) error {
 	}
 }
 
+func CheckOBADeleted(opt TestOptions, namespace string) error {
+	dynClient := NewKubeClientDynamic(
+		opt.HubCluster.ClusterServerURL,
+		opt.KubeConfig,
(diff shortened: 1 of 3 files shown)


tests/pkg/utils/mco_oba.go
```diff
@@ -31,6 +31,19 @@ func CheckOBAStatus(opt TestOptions, namespace, status string) error {
...
 }
 
+func CheckOBADeleted(opt TestOptions, namespace string) error {
+	dynClient := NewKubeClientDynamic(
+		opt.HubCluster.ClusterServerURL,
+		opt.KubeConfig,
+		opt.HubCluster.KubeContext)
+
+	_, err := dynClient.Resource(NewMCOAddonGVR()).Namespace(namespace).Get(context.TODO(), "observability-addon", metav1.GetOptions{})
+	if err == nil {
```
marcolan018, line 41: better to check error status/message, only return nil when error is notfound one

subbarao-meduri: The code is changed to specifically check for `NotFound` error.

PR #931, opened 2022-03-03
support WriteStorage

marcolan018:
https://github.com/stolostron/backlog/issues/19999


operators/multiclusterobservability/bundle/manifests/core.observatorium.io_observatoria.yaml
```diff
@@ -1,3 +1,5 @@
+
+---
```
clyang82, line 2: Need to remove the redundant `---`. it will lead to the failure installation.

clyang82 (changes requested):
Did you commit all your changes? I cannot find the definition of `PreConfiguredStorage`

operators/multiclusterobservability/bundle/manifests/core.observatorium.io_observatoria.yaml
marcolan018: done

marcolan018:
@clyang82 
PreConfiguredStorage is the one used by MetricsObjectSorage before, not introduced by this PR.

clyang82 (commented):
Do you plan to add some UT for it?

marcolan018:
@clyang82 
test cases will be covered in seperate pr

marcolan018:
@clyang82 UT added
will check the possibility to add the functional test later

clyang82:
I think we still need to watch the write storage secret for reconciling. there is one case in my mind that if I make the change in storage secret, there is nothing changes against observatorium CR. so there won't trigger reconciling. but we need to update the secret for observatorium api. right? if so, you can have a separated PR for this case.

/lgtm

PR #934, opened 2022-03-07
add observabilityaddon spec nil check.

morvencao:
ref: https://github.com/stolostron/backlog/issues/20463

We didn't add default value for observabilityaddon spec in release-2.2, so it will be nil if user doesn't specific the value.
For release-2.3 and beyond, the observabilityaddon spec and default value for `EnableMetrics` is true, so it will not be a  problem.


morvencao:
closing this in favor of https://github.com/stolostron/multicluster-observability-operator/pull/935

PR #940, opened 2022-03-10
Rename rules to recording_rules, update collect_rules schema

subbarao-meduri:


marcolan018:
@subbarao-meduri 
we'd better to add the unit test for collect rule in this pr to increase the coverage value in sonar

PR #947, opened 2022-03-15
Added observatorium-dynamic datasource to dashboards

dislbenn:
Related issue: https://github.com/stolostron/backlog/issues/19142

### DIFF
--- a/operators/multiclusterobservability/manifests/base/grafana/dash-k8s-compute-resources-node-pods.yaml
+++ b/operators/multiclusterobservability/manifests/base/grafana/dash-k8s-compute-resources-node-pods.yaml
@@ -53,7 +53,7 @@ data:
           "bars": false,
           "dashLength": 10,
           "dashes": false,
-          "datasource": "Observatorium-Dynamic",
+          "datasource": "$datasource",
           "fieldConfig": {
             "defaults": {
               "links": []
@@ -168,7 +168,7 @@ data:
           "type": "row"
         },
         {
-          "datasource": "Observatorium-Dynamic",
+          "datasource": "$datasource",
           "fieldConfig": {
             "defaults": {
               "color": {
@@ -463,7 +463,7 @@ data:
           "bars": false,
           "dashLength": 10,
           "dashes": false,
-          "datasource": "Observatorium-Dynamic",
+          "datasource": "$datasource",
           "fieldConfig": {
             "defaults": {
               "links": []
@@ -577,7 +577,7 @@ data:
           "type": "row"
         },
         {
-          "datasource": "Observatorium-Dynamic",
+          "datasource": "$datasource",
           "fieldConfig": {
             "defaults": {
               "color": {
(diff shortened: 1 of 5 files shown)


marcolan018 (changes requested):
we should not hardcode the datasource for each panels.
The better way is to update datasource var defined in dashboard json.
```
      "templating": {
        "list": [
          {
            "current": {
              "selected": false,
              "text": "",
              "value": ""
            },
            "description": null,
            "error": null,
            "hide": 2,
            "includeAll": false,
            "label": null,
            "multi": false,
            "name": "datasource",
            "options": [],
            "query": "prometheus",
            "refresh": 2,
            "regex": "Observatorium-Dynamic",
            "skipUrlSync": false,
            "type": "datasource"
          }
        ]
      },
```
In above datasource definition. I think add a new line " "regex": "Observatorium-Dynamic"," or similiar field can specify the datasource

dislbenn:
> we should not hardcode the datasource for each panels. The better way is to update datasource var defined in dashboard json.
> 
> ```
>       "templating": {
>         "list": [
>           {
>             "current": {
>               "selected": false,
>               "text": "",
>               "value": ""
>             },
>             "description": null,
>             "error": null,
>             "hide": 2,
>             "includeAll": false,
>             "label": null,
>             "multi": false,
>             "name": "datasource",
>             "options": [],
>             "query": "prometheus",
>             "refresh": 2,
>             "regex": "Observatorium-Dynamic",
>             "skipUrlSync": false,
>             "type": "datasource"
>           }
>         ]
>       },
> ```
> 
> In above datasource definition. I think add a new line " "regex": "Observatorium-Dynamic"," or similiar field can specify the datasource

@marcolan018 I added the regex that contains `Observatorium-Dynamic`. Can you review if things look good?

operators/multiclusterobservability/manifests/base/grafana/dash-k8s-compute-resources-node-pods.yaml
```diff
@@ -490,7 +507,7 @@ data:
           "steppedLine": false,
           "targets": [
             {
-              "expr": "sum(container_memory_working_set_bytes{cluster=\"$cluster\", node=\"$node\", container!=\"\",job=\"kubelet\", metrics_path=\"\/metrics\/cadvisor\", image!=\"\"}) by (pod)",
+              "expr": "sum(container_memory_working_set_bytes{cluster=\"$cluster\", node=\"$node\", container!=\"\",job=\"kubelet\", metrics_path=\"/metrics/cadvisor\", image!=\"\"}) by (pod)",
```
marcolan018, line 510: why modify expr here?

marcolan018:
the changes for datasource looks good for me, but better to verify that in your env with 2nd datasource available.

My concern is why the pr includes some changes on the expr in the dashboard json?

@dislbenn

dislbenn:
> the changes for datasource looks good for me, but better to verify that in your env with 2nd datasource available.
> 
> My concern is why the pr includes some changes on the expr in the dashboard json?
> 
> @dislbenn

It seems that those changes were made automatically when the dashboard was updated. Let me revert those expressions.

operators/multiclusterobservability/manifests/base/grafana/dash-k8s-compute-resources-node-pods.yaml
dislbenn: This seems to be made when I updated the dashboard. Let me take a look at it.

dislbenn:
@marcolan018 expression changes have been reverted.

PR #950, opened 2022-03-16
Set dynamic datasource interval to default when collect rules are empty

subbarao-meduri:

### DIFF
--- a/operators/multiclusterobservability/controllers/multiclusterobservability/grafana.go
+++ b/operators/multiclusterobservability/controllers/multiclusterobservability/grafana.go
@@ -74,7 +74,7 @@ func GenerateGrafanaDataSource(
 
 	DynamicTimeInterval := mco.Spec.ObservabilityAddonSpec.Interval
 
-	if DynamicTimeInterval > 30  && !config.CollectRulesEnabled {
+	if DynamicTimeInterval > 30 && config.CollectRulesEnabled {
 		DynamicTimeInterval = 30
 	}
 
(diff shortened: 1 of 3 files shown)


operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -598,6 +598,8 @@ func mergeCollectorRuleGroupList(defaultCollectRuleGroupList []operatorconfig.Co
 		}
 	}
 
+	config.CollectRulesEnabled = len(mergedCollectRuleGroups) > 0
```
marcolan018, line 601: should be this? Any ootb group disabled, don't use the dynamic datasource?
```
config.CollectRulesEnabled = len(mergedCollectRuleGroups) == 2
```

subbarao-meduri: Good point. Fixed in the latest commit.

operators/multiclusterobservability/controllers/multiclusterobservability/grafana.go
```diff
@@ -74,7 +74,7 @@ func GenerateGrafanaDataSource(
 
 	DynamicTimeInterval := mco.Spec.ObservabilityAddonSpec.Interval
 
-	if DynamicTimeInterval > 30 {
+	if DynamicTimeInterval > 30  && !config.CollectRulesEnabled {
```
marcolan018, line 77: should be
```
if DynamicTimeInterval > 30 && config.CollectRulesEnabled {
```

subbarao-meduri: resolved.

PR #951, opened 2022-03-16
add the rule name as a label for hash

marcolan018:
add the rule name as a label for hash, to handle the case that different rule has same label set

Signed-off-by: Marco llan@redhat.com


subbarao-meduri (approved):
LGTM

PR #958, opened 2022-03-18
disable metrics collection which defined in collect rule

marcolan018:
Signed-off-by: Marco llan@redhat.com


subbarao-meduri (approved):
LGTM

PR #959, opened 2022-03-18
Support SNO collect rules only

subbarao-meduri:

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
+++ b/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
@@ -602,8 +602,6 @@ func mergeCollectorRuleGroupList(defaultCollectRuleGroupList []operatorconfig.Co
 		}
 	}
 
-	config.CollectRulesEnabled = len(mergedCollectRuleGroups) > 0
-
 	return mergedCollectRuleGroups
 }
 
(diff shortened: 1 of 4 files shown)


operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -585,20 +585,24 @@ func mergeMetrics(defaultAllowlist []string, customAllowlist []string) []string
...
 
-	mergedCollectRuleGroups := []operatorconfig.CollectRuleGroup{}
 	for _, collectRuleGroup := range defaultCollectRuleGroupList {
 		if !deletedCollectRuleGroups[collectRuleGroup.Name] {
 			mergedCollectRuleGroups = append(mergedCollectRuleGroups, collectRuleGroup)
 		}
 	}
 
-	config.CollectRulesEnabled = len(mergedCollectRuleGroups) == 2
+	config.CollectRulesEnabled = len(mergedCollectRuleGroups) > 0
```
marcolan018, line 605: seems we need to remove this logic now. 
for the dynamic datasource, it always keep its' original value regardless of collect rule disable or not

subbarao-meduri: Removed logic in latest commit.

PR #985, opened 2022-04-05
fix error in test case to verify metrics in allowlist

marcolan018:
also add more detailed info in the logs of this test case


subbarao-meduri (approved):
/lgtm 
/approve

PR #1028, opened 2022-05-24
update the peer value in alertmanager sts to handle upgrade case

marcolan018:
fix: https://github.com/stolostron/backlog/issues/22605


subbarao-meduri (approved):
/lgtm

PR #1063, opened 2022-07-13
Collect Configuration Policy controller saturation metrics

mprahl:
This will allow users to set alerts when the Configuration Policy
controller is saturated. This will also allow them to determine if they
need to increase concurrency on the Configuration Policy controller.

Relates:
https://github.com/stolostron/backlog/issues/23547

### DIFF
--- a/operators/multiclusterobservability/manifests/base/config/metrics_allowlist.yaml
+++ b/operators/multiclusterobservability/manifests/base/config/metrics_allowlist.yaml
@@ -130,6 +130,9 @@ data:
       - cluster:policy_governance_info:propagated_noncompliant_count
       - policy:policy_governance_info:propagated_count
       - policy:policy_governance_info:propagated_noncompliant_count
+      - config_policies_evaluation_duration_seconds_bucket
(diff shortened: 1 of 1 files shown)


mprahl:
@subbarao-meduri could you please review?

subbarao-meduri:
@mprahl The changes look good. 

How many `labels` do each of these metrics create? It gives us an estimate of the memory overhead these metrics could create in a large scale (1k/2k managed clusters) environments. 

Given that you are not currently a ALERT that goes with these metrics out-of-box, another approach could be to document how a specific environment could add these as custom metrics along with configuring an ALERT that uses these metrics. Thoughts?

mprahl:
@subbarao-meduri, I or @JustinKuli will add an alert hopefully by the end of the sprint.

/hold for now

JustinKuli:
@subbarao-meduri , there won't be any additional labels on this histogram, so currently there would be 16 total data points: 1 `sum`, 1 `count`, and 14 `buckets`.

PR #1102, opened 2022-10-05
adding policy for centralized alert management

bjoydeep:
Signed-off-by: Joydeep Banerjee <jbanerje@redhat.com>

Adding sample policies to configure centralized alert management. Attached are 3 policies with a README. Actually, as I have seen, this does `not` interfere even if ACM Observability is enabled. Please verify.

### DIFF
--- a/examples/policies/README.md
+++ b/examples/policies/README.md
@@ -1,0 +1,16 @@
+### Goal of the 3 policies
+If apply these policies to a managed cluster, it will configure the Prometheus of the managed cluster to forward all it alerts to the alertmanager on the ACM hub. It does so by configuring the cluster-monitoring-config configmap in openshift-monitoring namespace of the managed cluster. This cannot be applied to ACM hub unless [this feature](https://issues.redhat.com/browse/ACM-1756) is delivered.
+
+### Notes
+- When the policies are deleted, we do not delete the resources create by the policies. This is just to play it safe because we are dealing with cluster-monitoring-config configmap - a critical item in configuring OCP monitoring.
+- we cannot create policy set out of these because these policies need to be created in different namespaces as it accesses resources in those namespaces in the ACM Hub for hub side templating.
+- These polices will be supported on ACM 2.6; Not sure if ACM 2.5 will be able to support these levels of hub side and managed cluster side templates. However, it is not neccessary that these resources be delivered through policy either. As long as the right resources - configmap, secrets are delivered with correct content, the function will be served.
+
+### Description
+
+Policy | Description | Additional Info
+--- | --- | ---
+policy-hub-alert-routing.yaml| This needs to be created in openshift-ingress namespace. This takes the CA certs needed  to complete the router handshake and injects them into a secret in the openshift-monitoring namespace. | Needless to say, if custom certificates are used, appropriate changes need to be made to this policy.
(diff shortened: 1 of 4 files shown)


examples/policies/policy-cluster-monitoring-config.yaml
```diff
@@ -0,0 +1,81 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
```
marcolan018, line 2: what will this policy do if the cluster-monitoring-config configmap already existed

bjoydeep: @marcolan018 seems it does nothing - works well. I just tested it. Check the jb-4 Hub server that I slacked you this am.

examples/policies/policy-cluster-monitoring-config.yaml
```diff
@@ -0,0 +1,81 @@
...
+                          staticConfigs:
+                          - {{hub (lookup "route.openshift.io/v1" "Route" "open-cluster-management-observability" "alertmanager").spec.host hub}}
+                          tlsConfig:
+                            ServerName: ""
+                            ca:
+                              key: service-ca.crt
+                              name: hub-alertmanager-router-ca
+                            insecureSkipVerify: false
+                        externalLabels:
+                          cluster: '{{ fromClusterClaim "id.openshift.io" }}'
```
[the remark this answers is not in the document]
bjoydeep: Yes, it does. Reference doc: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.6/html/governance/governance#fromclusterclaim-func

examples/policies/README.md
```diff
@@ -0,0 +1,16 @@
+### Goal of the 3 policies
+If apply these policies to a managed cluster, it will configure the Prometheus of the managed cluster to forward all it alerts to the alertmanager on the ACM hub. It does so by configuring the cluster-monitoring-config configmap in openshift-monitoring namespace of the managed cluster. This cannot be applied to ACM hub unless [this feature](https://issues.redhat.com/browse/ACM-1756) is delivered.
+
+### Notes
+- When the policies are deleted, we do not delete the resources create by the policies. This is just to play it safe because we are dealing with cluster-monitoring-config configmap - a critical item in configuring OCP monitoring.
```
marcolan018, line 5: in current logic in endpoint operator, if observability is disabled, we will revert the changes we made before, which is more graceful than the policy

bjoydeep: We can make that as well by adding one line to all the 3 policies. But if a customer is also adding changes to that cluster-monitoring-config outside our changes (since this is a shared object/API), then we will have problems. So I am having a pause. I am not sure about the best approach here.

bjoydeep: You can rightly point out that if this configmap is being controlled by policy, as long as policy is enabled, no other changes can be made to the configmap.

bjoydeep:
@ch-stark This was list of 3 policies which we were thinking of adding as examples..

examples/policies/policy-cluster-monitoring-config.yaml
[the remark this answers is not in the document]
imiller0: Is this clusterClaim always defined on the spoke clusters or does it need to be explicitly created?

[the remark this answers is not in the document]
bjoydeep: clusterclaim is always created @imiller0 . One per label you add for that cluster.

subbarao-meduri (approved):
/lgtm

PR #1115, opened 2022-10-24
Pickup rotated hub kubeconfig

subbarao-meduri:


operators/endpointmetrics/controllers/lease/lease_controller.go
```diff
@@ -0,0 +1,167 @@
+package lease
```
marcolan018, line 1: why do we need this?
The lease is handled by pkg/util/lease.go now. If want to fix the hub kubeconfig roated problem, should we fix it in the lease.go?

operators/endpointmetrics/controllers/status/status_controller_test.go
```diff
@@ -69,7 +69,7 @@ func TestStatusController(t *testing.T) {
 		t.Fatalf("reconcile: miss the error for missing obervabilityaddon")
 	}
 
-	// test status in local pushed to hub
+	// tes:99in local pushed to hub
```
marcolan018, line 72: is this a typo?

operators/endpointmetrics/controllers/status/status_controller.go
```diff
@@ -7,15 +7,16 @@ import (
 	"os"
 	"reflect"
 
+	"github.com/stolostron/multicluster-observability-operator/operators/endpointmetrics/pkg/util"
```
marcolan018, line 10: don't make this change.
to organize package from stolostron org in a separate section is the code style all obs codes are using.

operators/endpointmetrics/controllers/lease/lease_controller.go
subbarao-meduri: `lease.go` was using a utility function `open-cluster-management.io/addon-framework/pkg/lease.NewLeaseUpdater`  earlier, and this did not allow for refreshing hub kubeconfig. In the newly refactored code, lease.go is copied into the project, modified to take a path to hub kubeconfig so it can be refreshed when hub kubeconfig becomes invalid. Do you see a simpler way to address this?

operators/endpointmetrics/controllers/status/status_controller_test.go
subbarao-meduri: Yes, will fix it in the next update

operators/endpointmetrics/controllers/lease/lease_controller.go
marcolan018: no, just want to confirm this is the best option recommended by foundation squad

PR #1125, opened 2022-11-08
[release-2.5] Ability to disable alert forwarding

subbarao-meduri:

### DIFF
--- a/operators/endpointmetrics/pkg/rendering/renderer.go
+++ b/operators/endpointmetrics/pkg/rendering/renderer.go
@@ -126,6 +126,7 @@ func Render(
 			}
 			spec.ExternalLabels[operatorconfig.ClusterLabelKeyForAlerts] = hubInfo.ClusterName
 			if hubInfo.AlertmanagerEndpoint == "" {
+				log.Info("AlertManagerEndpoint is nil")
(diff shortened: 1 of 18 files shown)


operators/endpointmetrics/pkg/rendering/renderer.go
```diff
@@ -198,7 +198,7 @@ func Render(
 					s.GetName(),
 				)
 			}
-			// replace the hub alertmanager address
+			// replace the hub alertmanager address. Address will be set to null when alerts are disabled
```
marcolan018, line 201: is this pr handle the *KS scenario?

subbarao-meduri: Yes @marcolan018

subbarao-meduri: When alerts are disabled, they are disabled for *all* clusters, not just for OCP. Are there specific concerns @marcolan018 ?

marcolan018: I cannot find any related change related with *KS.
if alert forwarding disabled in *KS, just send an empty url to the prometheus?

PR #1131, opened 2022-11-18
Blank labels removed and appended ones done Lexicographically

ngraham20:
Signed-off-by: Nathaniel Graham <ngraham@redhat.com>

### DIFF
--- a/collectors/metrics/pkg/metricfamily/label.go
+++ b/collectors/metrics/pkg/metricfamily/label.go
@@ -1,7 +1,6 @@
 package metricfamily
 
 import (
-	"sort"
 	"sync"
 
 	clientmodel "github.com/prometheus/client_model/go"
@@ -64,7 +63,7 @@ func appendLabels(
 		value := pair.GetValue()
 		// remove any name == "" or value == nil
 		if name != "" && value != "" {
-			withoutEmpties = append(withoutEmpties, existing[i])
+			withoutEmpties = insertLexicographicallyByName(withoutEmpties, existing[i])
 		}
 	}
 	existing = withoutEmpties
@@ -86,9 +85,7 @@ func appendLabels(
 			existing = insertLexicographicallyByName(existing, v)
 		}
 	}
-	sort.SliceStable(existing, func(i, j int) bool {
-		return existing[i].GetName() < existing[j].GetName()
-	})
+
 	return existing
 }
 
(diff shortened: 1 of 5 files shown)


collectors/metrics/pkg/metricfamily/label.go
```diff
@@ -53,21 +53,44 @@ func appendLabels(
 	existing []*clientmodel.LabelPair,
 	overrides map[string]*clientmodel.LabelPair) []*clientmodel.LabelPair {
 	var found []string
+
+	// override matching existing labels
 	for i, pair := range existing {
 		name := pair.GetName()
+		// remove any name == "" or value == nil
+		if name == "" || pair != nil {
```
marcolan018, line 61: should we use pair.getValue()?

collectors/metrics/pkg/metricfamily/label.go
```diff
@@ -53,21 +53,44 @@ func appendLabels(
...
+		// only append real names and values
+		// don't append anything overwritten above
+		if k != "" && v != nil && !contains(found, k) {
+			existing = insertLexicographicallyByName(existing, v)
 		}
 	}
 	return existing
 }
 
+func insertLexicographicallyByName(
```
marcolan018, line 81: we also need this function in https://github.com/stolostron/multicluster-observability-operator/blob/072204b63ed0fab76611e21a8d5b63a274afbbc3/collectors/metrics/pkg/metricfamily/hypershift_transformer.go#L81-L121

collectors/metrics/pkg/metricfamily/label.go
ngraham20: Oh, I totally misunderstood what we were checking for. Updating to your suggestion

collectors/metrics/pkg/metricfamily/label.go
```diff
@@ -53,18 +55,53 @@ func appendLabels(
...
 	for k, v := range overrides {
-		if !contains(found, k) {
-			existing = append(existing, v)
+		// only append real names and values
+		// don't append anything overwritten above
+		if k != "" && v.GetValue() != "" && !contains(found, k) {
+			existing = insertLexicographicallyByName(existing, v)
 		}
 	}
+	sort.SliceStable(existing, func(i, j int) bool {
```
marcolan018, line 89: do we need to sort all labels here?

ngraham20: ah, good point. Probably not, I will test and update here

ngraham20: Turns out removing that breaks it. I've gone ahead and just lexicographically inserted instead of appending further up in the code to resolve this

PR #1146, opened 2023-01-09
Do not reconcile if annotation present

ngraham20:
Signed-off-by: Nathaniel Graham <ngraham@redhat.com>

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
+++ b/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
@@ -118,15 +118,6 @@ func (r *PlacementRuleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, nil
 	}
 
-	// Do not reconcile objects if this instance of mch has the
-	// `disableAddonAutomaticInstallationAnnotationKey` annotation
-	if value, ok := mco.GetAnnotations()[disableAddonAutomaticInstallationAnnotationKey]; ok &&
-		strings.EqualFold(value, "true") {
-
-		reqLogger.Info("Cluster has disable addon automatic installation annotation. Skip addon deploy")
-		return reconcile.Result{}, nil
-	}
-
 	if !deleteAll && !mco.Spec.ObservabilityAddonSpec.EnableMetrics {
 		reqLogger.Info("EnableMetrics is set to false. Delete Observability addons")
 		deleteAll = true
@@ -525,6 +516,17 @@ func updateManagedClusterList(obj client.Object) {
 	}
 }
 
+// Do not reconcile objects if this instance of mch has the
+// `disableAddonAutomaticInstallationAnnotationKey` annotation
+func isAutomaticAddonInstallationDisabled(obj client.Object) bool {
+	annotations := obj.GetAnnotations()
+	if val, ok := annotations[disableAddonAutomaticInstallationAnnotationKey]; ok && strings.EqualFold(val, "true") {
+		log.Info("Cluster has disable addon automatic installation annotation. Skip addon deploy")
+		return true
+	}
+	return false
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *PlacementRuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	c := mgr.GetClient()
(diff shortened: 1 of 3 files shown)


operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -116,6 +118,15 @@ func (r *PlacementRuleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, nil
 	}
 
+	// Do not reconcile objects if this instance of mch has the
+	// `disableAddonAutomaticInstallationAnnotationKey` annotation
+	if value, ok := mco.GetAnnotations()[disableAddonAutomaticInstallationAnnotationKey]; ok &&
```
subbarao-meduri, line 123: The annotation should be checked on each of the `ManagedCluster` objects, not on MCO custom resource.

ngraham20: Makes sense. I'll update that and move it into the predicate per slack discussion

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -514,44 +516,74 @@ func updateManagedClusterList(obj client.Object) {
...
 	c := mgr.GetClient()
 	ingressCtlCrdExists := r.CRDMap[config.IngressControllerCRD]
 	clusterPred := predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
 			log.Info("CreateFunc", "managedCluster", e.Object.GetName())
+
 			updateManagedClusterList(e.Object)
 			updateManagedClusterImageRegistry(e.Object)
+
+			if isAutomaticAddonInstallationDisabled(e.Object) {
```
ngraham20, line 541: Could this be moved to the top of Create() as an early return?

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -514,44 +516,74 @@ func updateManagedClusterList(obj client.Object) {
...
-					managedClusterImageRegistryMutex.Lock()
-					delete(managedClusterImageRegistry, e.ObjectNew.GetName())
-					managedClusterImageRegistryMutex.Unlock()
-				} else {
-					updateManagedClusterList(e.ObjectNew)
-					updateManagedClusterImageRegistry(e.ObjectNew)
-				}
-				return true
+
+			if e.ObjectNew.GetResourceVersion() == e.ObjectOld.GetResourceVersion() {
```
ngraham20, line 550: I reorganized here in favor of early returns. Could the disable addon function also be an early return, or is it preferred to be at the bottom to allow other information to propegate?

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -514,44 +516,74 @@ func updateManagedClusterList(obj client.Object) {
...
 			log.Info("DeleteFunc", "managedCluster", e.Object.GetName())
+
 			managedClusterListMutex.Lock()
 			delete(managedClusterList, e.Object.GetName())
 			managedClusterListMutex.Unlock()
 			managedClusterImageRegistryMutex.Lock()
 			delete(managedClusterImageRegistry, e.Object.GetName())
 			managedClusterImageRegistryMutex.Unlock()
+
+			if isAutomaticAddonInstallationDisabled(e.Object) {
```
ngraham20, line 583: could this be moved to the top as an early return?

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -514,44 +516,74 @@ func updateManagedClusterList(obj client.Object) {
...
 		CreateFunc: func(e event.CreateEvent) bool {
 			log.Info("CreateFunc", "managedCluster", e.Object.GetName())
+
 			updateManagedClusterList(e.Object)
 			updateManagedClusterImageRegistry(e.Object)
+
+			if isAutomaticAddonInstallationDisabled(e.Object) {
+				return false
+			}
+
```
ngraham20, line 544: Could this be moved to the top of Create() as an early return?

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -514,44 +516,74 @@ func updateManagedClusterList(obj client.Object) {
 	}
 }
 
+// Do not reconcile objects if this instance of mch has the
+// `disableAddonAutomaticInstallationAnnotationKey` annotation
+func isAutomaticAddonInstallationDisabled(obj client.Object) bool {
+	annotations := obj.GetAnnotations()
+	if val, ok := annotations[disableAddonAutomaticInstallationAnnotationKey]; ok && strings.EqualFold(val, "true") {
```
marcolan018, line 523: need to check whether annotations is nil

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
marcolan018: yes, you need to do that to avoid updateManagedClusterList() method invoked when disable annotations is added

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -514,44 +516,74 @@ func updateManagedClusterList(obj client.Object) {
...
-					managedClusterImageRegistryMutex.Unlock()
-				} else {
-					updateManagedClusterList(e.ObjectNew)
-					updateManagedClusterImageRegistry(e.ObjectNew)
-				}
-				return true
+
+			if e.ObjectNew.GetResourceVersion() == e.ObjectOld.GetResourceVersion() {
+				return false
 			}
```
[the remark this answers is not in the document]
marcolan018: the check should at the top, not at bottom. At least should be placed before the invocaion of method updateManagedClusterList()

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
ngraham20: it looks like golang will not panic when indexing a nil map. I can still add the check for code clarity, but it seems Go only panics if you try to insert into a nil map

subbarao-meduri (approved):
/lgtm

PR #1148, opened 2023-01-11
Generate unique uid for dashboards

dislbenn:
Related issue: https://issues.redhat.com/browse/ACM-1732
Signed-off-by: dislbenn <dbennett@redhat.com>


operators/multiclusterobservability/manifests/base/grafana/cluster-role.yaml
```diff
@@ -8,9 +8,7 @@ rules:
   resources:
   - configmaps
   verbs:
-  - get
-  - list
-  - watch
+  - '*'
```
marcolan018, line 11: better to not to use "*" in cluster-role definition

loaders/dashboards/pkg/controller/dashboard_controller.go
```diff
@@ -259,17 +270,51 @@ func updateDashboard(old, new interface{}, overwrite bool) {
...
+				klog.Warningf("failed to update configmap: %v", err)
+			}
+		}
+
+		uid := dashboard["uid"].(string)
+
+		if _, ok := uidentifierList[uid]; !ok {
+			uidentifierList[uid] = name
+		} else if uidentifierList[uid] != name {
+			klog.Warningf("Dashboard named '%v' has the same uid. generating new uid for dashboard '%v'",
```
marcolan018, line 304: what's the best option if duplicate uid detected?
to generate new uid, or reject the import for the dashboard?

marcolan018: https://github.com/stolostron/multicluster-observability-operator/blob/bea1ed3ef1e448e1478af0ee1a10175a089aa595/operators/multiclusterobservability/pkg/util/clustermanagementaddon.go#L24

cluster overview dashboard's id is used here. We need additional check to guarantee the id not changed for this dashboard.

PR #1152, opened 2023-01-16
use oc from builder image

subbarao-meduri:

Per CICD team, `oc` exists in the builder image. To be compatible across linux and arm builds, copy `oc` from the builder image instead of downloading. This also solves the issue of not having to worry about the version of `oc` to use


ngraham20 (commented):
this is the actual review submit. The ltgm is above

PR #1172, opened 2023-03-23
don't fail forward process if any error in recording rule format(2.7)

marcolan018:


subbarao-meduri (approved):
/lgtm

PR #1177, opened 2023-04-14
[ACM-4956] use cluster_version metric for cluster labels

subbarao-meduri:
Switch to using `cluster_version` metric instead of the expensive `kube_pod_info` metric to obtain cluster labels


marcolan018 (approved):
lgtm

PR #1178, opened 2023-04-21
[ACM-5057] Use cluster_version metric for cluster label_values (#1177)

dislbenn:
Switch to using cluster_version metric instead of the expensive kube_pod_info metric to obtain cluster labels


subbarao-meduri (approved):
/lgtm

PR #1182, opened 2023-04-25
[ACM-5061] Updated "USE Method / Node" dashboard to allow multiple node selection

dislbenn:
Updating the dashboard to provide users with the ability to select multiple nodes when viewing the node dashboard.

JIRA Issue: https://issues.redhat.com/browse/ACM-5061


subbarao-meduri (approved):
/lgtm

PR #1184, opened 2023-04-26
[ACM-5059] - Update dashboards to use relative time range when drilling down

dislbenn:
JIRA Issue: https://issues.redhat.com/browse/ACM-5059


subbarao-meduri (approved):
/lgtm

PR #1186, opened 2023-05-01
[ACM-5059]Updated links to pass relative time in alert dashboard

dislbenn:
Updating the alert analysis dashboard to pass the relative time range when drilling down.
Jira Issue: https://issues.redhat.com/browse/ACM-5059


subbarao-meduri (approved):
/lgtm

PR #1187, opened 2023-05-05
[ACM-5263] Updated ViolatedPolicyReport alert expression

dislbenn:
Jira issue: https://issues.redhat.com/browse/ACM-5263

Updating the ViolatedPolicyReport alert expression to include the `cluster` label.


operators/multiclusterobservability/manifests/base/alertmanager/alert_rules.yaml
```diff
@@ -37,7 +37,7 @@ data:
           annotations:
             summary: "There is a policy report violation with a {{ $labels.severity }} severity level detected."
             description: "The policy: {{ $labels.policy }} has a severity of {{ $labels.severity }} on cluster: {{ $labels.cluster }}"
-          expr: sum(policyreport_info) by (managed_cluster_id, category, clusterID, policy, severity) > 0
+          expr: sum(policyreport_info * on (managed_cluster_id) group_left (cluster) label_replace(acm_managed_cluster_labels, "cluster", "$1", "name", "(.*)")) by (cluster, category, clusterID, managed_cluster_id, policy, severity) > 0
```
bjoydeep, line 40: How does backward compatibility work? I do not mean back porting. Back porting will not work if we do not have if we do not have the acm_managed_cluster_labels metrics. But if a customer was already seeing this alert and acting on top of it -- what is their response if we change this alert? What does OpenShift do?
Also we need to think through if the `cluster` label is ok. It probably is not. Lets review this with https://docs.google.com/document/d/1niTURWaStVXWSzJQ28en3KB6L6mJj2jyaanoelJ5vpM/edit

dislbenn: @bjoydeep in the doc it states that the `cluster` and `clusterID` are not being added by OCP; however, it is added by the metrics-collector. These labels are already being overwritten by the metrics-collector if they existed within the original metrics, but for the `policyreport_info` metric, this label is not added by default; therefore, the label is never added.

These changes will just create a new label called `cluster` for the alert expression and provide only the alert with the added cluster label. I would assume if the `cluster` label is being used by a customer, that the label would overwritten with the `cluster` name.

bjoydeep: @dislbenn in fact adding a label for identifying the real cluster can only be good. Today it is missing - and that causes all sorts of issues. Question is what should the name of the label be.
For all other ALERTS the name of the labels that identify the cluster that has a problem is - `cluster` and `clusterID`.
So using `cluster` would make sense. 
Overwriting is a function of the Alert forwarding from managed clusters and we cannot stop the customer from adding things - and that is not applicable for alerts evaluated by the Thanos ruler in ACM hub.

However, what do we do with `ClusterID`. Traditionally this means the offending cluster. In the case of this alert,  its `managed_cluster_id` .

bjoydeep: Here is a proposal. Check it carefully. The expression looks complex:
`sum without (managed_cluster_id) (label_replace((sum without (clusterID) (label_replace((sum(policyreport_info * on (managed_cluster_id) group_left (cluster) label_replace(acm_managed_cluster_labels, "cluster", "$1", "name", "(.*)")) by (cluster, category, clusterID, managed_cluster_id, policy, severity)),"hub_clusterID", "$1", "clusterID", "(.*)"))),"clusterID", "$1", "managed_cluster_id", "(.*)"))`
There are 2 label_replace statements in there. 

1. replaces clusterID with hub_clusterID and drops the clusterID
2. replaces managed_cluster_id with clusterID and drops the managed_cluster_id

operators/multiclusterobservability/manifests/base/grafana/dash-acm-alert-analysis.yaml
```diff
@@ -633,7 +633,7 @@ data:
           "targets": [
             {
               "exemplar": true,
-              "expr": "sum(ALERTS{alertstate=\"firing\"}) by (cluster,alertname,severity)",
+              "expr": "sum(ALERTS{alertstate=\"firing\", severity=~\"$severity\"}) by (cluster, alertname, severity)",
```
bjoydeep, line 636: Creating a `cluster` label as you have proposed above will help in making this display better

operators/multiclusterobservability/manifests/base/alertmanager/alert_rules.yaml
dislbenn: Adding the following expression:
```yaml
expr: sum(policyreport_info * on (managed_cluster_id) group_left (cluster, hub_cluster_id) label_replace(acm_managed_cluster_labels, "cluster", "$1", "name", "(.*)")) by (cluster, hub_cluster_id, managed_cluster_id, category, clusterID, policy, severity) > 0
```

This will allow us to do the following:

1. Join on the `cluster` and `hub_cluster_id`
2. Allow us to overwrite the `clusterID` label with the `managed_cluster_id` within the alerting rule.

By doing the above, this will allow us to get the necessary labels for the alert.

operators/multiclusterobservability/manifests/base/alertmanager/alert_rules.yaml
```diff
@@ -37,9 +37,10 @@ data:
           annotations:
             summary: "There is a policy report violation with a {{ $labels.severity }} severity level detected."
             description: "The policy: {{ $labels.policy }} has a severity of {{ $labels.severity }} on cluster: {{ $labels.cluster }}"
-          expr: sum(policyreport_info) by (managed_cluster_id, category, clusterID, policy, severity) > 0
+          expr: sum(policyreport_info * on (managed_cluster_id) group_left (cluster, hub_cluster_id) label_replace(acm_managed_cluster_labels, "cluster", "$1", "name", "(.*)")) by (cluster, hub_cluster_id, managed_cluster_id, category, clusterID, policy, severity) > 0
           for: 1m
           labels:
+            clusterID: "{{ $labels.managed_cluster_id }}"
```
bjoydeep, line 43: I am assuming you have checked that real alerts add the other labels:
- cluster
- hub_cluster_id.

This is neat and slick.

Where I hesitate a little is if we just run this query in line 40, the result is confusing on the columns:
- clusterID, hub_cluster_id and managed_cluster_id.
That confusion is sorted out only when we look at the label swapping.
That gives me a little pause.

PR #1201, opened 2023-05-30
[ACM-5628][ACM-5700] Resolved ACM alert dashboards errors

dislbenn:
- Jira issue: https://issues.redhat.com/browse/ACM-5628
  - Fix: Added an override field for the display name in the Grafana panel. This will allow the `local-cluster` name to show when it's the only cluster available.
- https://issues.redhat.com/browse/ACM-5700
  - Fix: Added an `important` alert stat panel at the top and the link to the lower level dashboard.


subbarao-meduri (approved):
/lgtm

PR #1202, opened 2023-05-30
upgrade add-on framework package to get past 50K ManifestWork limit

subbarao-meduri:
Related to: https://issues.redhat.com/browse/ACM-5663

New add-on framework version allows for passing up to 500K ManifestWork objects


subbarao-meduri:
Redoing PR.

PR #1204, opened 2023-06-01
[ACM-5752] Update default allowlist

subbarao-meduri:
https://issues.redhat.com/browse/ACM-5752


bjoydeep:
@subbarao-meduri do you know how many versions should we go back in ACM to back port this?

PR #1207, opened 2023-06-06
[ACM-5782][release-2.8] Added resync for required managedcluster labels

dislbenn:
Updated `rbac-query-proxy` resync to prevent adding the labels `name` and `cluster.open-cluster-management.io/clusterset` to `managedcluster` label ignore list.

PR #1209, opened 2023-06-14
[release-2.9] Added resync for required managedcluster labels (#1207)

dislbenn:
This feature was implemented only in ACM 2.8; therefore, we need to bring it into this current release.


subbarao-meduri (approved):
/lgtm

PR #1215, opened 2023-06-30
insert labels lexicographically when hypershift

ngraham20:


ngraham20:
Manual testing shows that "Out of order label set" errors are no longer present with this change.

ngraham20:
There remains an "Out of order Sample" that this fix does not address, however.

PR #1237, opened 2023-08-31
Rename OpenShift configmap reloader key to configmap_reloader

subbarao-meduri:
PRs https://github.com/stolostron/release/pull/508, https://github.com/stolostron/release/pull/509 and https://github.com/stolostron/release/pull/510 which were done to eliminate seeming dulicatoin/overlap were in error, and removed needed keys. This probably came about because of the close naming of the prometheus_config_reloader and prometheus-config-reloader keys, which suggested they were two entries for the same image, but they were not.

1. `prometheus_config_reloader` is a key that points to an ACM-built image for a Prometneus-specific image that triggers reloading on changes to Prometheus config info.
2. `configmap_reloader` is a key that points to an external (from Openshift) image that is not specific to Prometheus and causes containers to be restarted on changes to ConfigMaps or Secrets. This image also happens to be used by Observability pods (Stateful sets), but since its not Prometheus specific, we're dripping that prefix from its name.)

Observability operator code will be updated to use the new key `configmap_reloader` instead of the old `promethues-configmap-reload` across all releases.


marcolan018 (approved):
/lgtm

PR #1238, opened 2023-09-05
[ACM-7321] Configure alerts to detect acm hub compactor problems

subbarao-meduri:
These alerts are modeled after [RHOBS compactor alerts ](https://github.com/rhobs/configuration/blob/2e740eca3d9321dc916c9d54a17526829d8dffee/resources/observability/prometheusrules/observatorium-thanos-production.prometheusrules.yaml#L79-L151)

### DIFF
--- a/examples/export/v1beta2/custom-certs/observability.yaml
+++ b/examples/export/v1beta2/custom-certs/observability.yaml
@@ -45,7 +45,7 @@ spec:
         limits:
           cpu: 1
           memory: 4Gi
-      replicas: 2
+      replicas: 3
       serviceAccountAnnotations:
         test.com/role-arn: 's3_role'
     rule:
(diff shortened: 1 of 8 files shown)


bjoydeep:
@subbarao-meduri we also have seen Thanos fail when the Compactor PV run out of space. It is interesting that you chose not to monitor the space. Probably because it would get reflected in the `halted` metric?

subbarao-meduri:
> @subbarao-meduri we also have seen Compactor fail when the Compactor PV run out of space. It is interesting that you chose not to monitor the space. Probably because it would get reflected in the `halted` metric?

@bjoydeep yes, any underlying infrastructure problems will surface through failing metrics, compactor halting, compactions failing, bucket operations failing etc. We could add another alert for PV size, but I think that would be somewhat redundant, imo.

PR #1252, opened 2023-10-11
[ACM-7644] support addonDeploymentConfig.ProxyConfig for managed cluster components

subbarao-meduri:
This PR includes changes needed by story [ACM-7023](https://issues.redhat.com/browse/)

- detects and propagates proxy settings from `addonDeploymentConfig` to endpoint operator. This is necessary so endpoint operator could use proxy settings while writing addonStatus
- propagates settings from endpoint operator to metrics collector. This is necessary for metrics ingestion path via mTLS to transparently tunnel metrics to hub Observatorium gateway. The proxy must be support tunnel mode and must not terminate  mTLS connection for this to work.


operators/endpointmetrics/controllers/observabilityendpoint/metrics_collector.go
```diff
@@ -427,7 +463,7 @@ func getMetricsAllowlist(ctx context.Context, c client.Client,
 	}
 
 	cmList := &corev1.ConfigMapList{}
-	err = c.List(ctx, cmList, &client.ListOptions{})
+	_ = c.List(ctx, cmList, &client.ListOptions{})
```
saswatamcode, line 466: Shouldn't we check for errors here?

saswatamcode (approved):
Seems good to me, as far as I can tell. Looks like tests are passing as well.

Just small comment.

operators/endpointmetrics/controllers/observabilityendpoint/metrics_collector.go
subbarao-meduri: Agreed. We were ignoring the error before too. There are other places in this method where error is only logged, but not returned. I will create a separate issue to address this.

PR #1253, opened 2023-10-11
*: Refactor codebase with linters/formatters and upstream patterns

saswatamcode:
Lots of TODOs still pending (that will be done in separate PRs), but this does the following.

- Bump prom client, and add rules to use similar upstream patterns for instrumentation
- Deprecate ioutil
- Make all errors consistent and move completely to `github.com/efficientgo/core/errors`
- Add bingo and several linters/formatters
- Deprecate fake.NewClient
- Fix a bunch of lint problems (although many still exist within tests)


collectors/metrics/cmd/metrics-collector/main.go
```diff
@@ -282,17 +287,30 @@ type Options struct {
...
 	if err != nil {
 		return err
 	}
 
+	cfg.Registry = metricsReg
+
 	worker, err := forwarder.New(*cfg)
 	if err != nil {
-		return fmt.Errorf("failed to configure metrics collector: %v", err)
+		return errors.Wrap(err, "failed to configure metrics collector")
```
philipgough, line 313: any particular reason we would choose to import this package to handle errors in favour of stdlib?

cicd-scripts/install-dependencies.sh
```diff
@@ -7,11 +7,11 @@ echo "install dependencies"
...
-    elif [[ "$OSTYPE" == "darwin"* ]]; then
-            curl -L https://github.com/operator-framework/operator-sdk/releases/download/${_OPERATOR_SDK_VERSION}/operator-sdk_darwin_amd64 -o operator-sdk
-    fi
-    chmod +x operator-sdk
-    sudo mv operator-sdk /usr/local/bin/operator-sdk
+  if [[ $OSTYPE == "linux-gnu" ]]; then
+    curl -L https://github.com/operator-framework/operator-sdk/releases/download/${_OPERATOR_SDK_VERSION}/operator-sdk_linux_amd64 -o operator-sdk
+  elif [[ $OSTYPE == "darwin"* ]]; then
+    curl -L https://github.com/operator-framework/operator-sdk/releases/download/${_OPERATOR_SDK_VERSION}/operator-sdk_darwin_amd64 -o operator-sdk
+  fi
```
philipgough, line 14: could probably make this a one liner with some uname command as a subshell?

collectors/metrics/pkg/forwarder/forwarder.go
```diff
@@ -34,31 +33,10 @@ const (
...
-		Name: "federate_errors",
-		Help: "The number of times forwarding federated metrics has failed",
-	})
-)
-
 type RuleMatcher interface {
 	MatchRules() []string
 }
 
-func init() {
```
philipgough, line 56: 👍

collectors/metrics/pkg/forwarder/forwarder.go
```diff
@@ -233,6 +217,18 @@ func New(cfg Config) (*Worker, error) {
...
 		simulatedTimeseriesFile: cfg.SimulatedTimeseriesFile,
+		gaugeFederateSamples: promauto.With(cfg.Registry).NewGauge(prometheus.GaugeOpts{
+			Name: "federate_samples",
+			Help: "Tracks the number of samples per federation",
+		}),
+		gaugeFederateFilteredSamples: promauto.With(cfg.Registry).NewGauge(prometheus.GaugeOpts{
+			Name: "federate_filtered_samples",
+			Help: "Tracks the number of samples filtered per federation",
+		}),
+		gaugeFederateErrors: promauto.With(cfg.Registry).NewGauge(prometheus.GaugeOpts{
```
philipgough, line 228: should we add a metric to calc total also?

collectors/metrics/test/integration/prereq.sh
```diff
@@ -2,50 +2,49 @@
...
-			rm openshift-client-mac-4.4.14.tar.gz
-		elif [[ "$(uname)" == "Linux" ]]; then # we are in travis, building in rhel 
-			curl -LO https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.4.14/openshift-client-linux-4.4.14.tar.gz
-			tar xzvf openshift-client-linux-4.4.14.tar.gz  # xzf to quiet logs
-			rm openshift-client-linux-4.4.14.tar.gz
-		fi
-		# this package has a binary, so:
+  command -v kubectl
+  if [ $? -ne 0 ]; then
+    echo "=====Setup kubectl====="
```
philipgough, line 10: maybe we could take some inspiration from https://github.com/rhobs/configuration/blob/main/Makefile#L216-L218 around handling this? seems cleaner?

collectors/metrics/test/integration/setup.sh
```diff
@@ -2,181 +2,182 @@
...
-	echo $(pwd)
-	mv README.md.tmp README.md 
-	chmod +x ./kubectl
-	if [[ ! -f /usr/local/bin/kubectl ]]; then
-		sudo cp ./kubectl /usr/local/bin/kubectl
-	fi
-	# kubectl are now installed in current dir 
-	echo -n "kubectl version" && kubectl version
+  echo "=====Setup kubectl====="
+  # kubectl required for kind
```
philipgough, line 35: this all looks repeated? maybe we can simplify

collectors/metrics/test/integration/setup.sh
```diff
@@ -2,181 +2,182 @@
...
-    sleep 60
-    echo -n "available services: " && kubectl get svc --all-namespaces
+  echo "=====Setting up prometheus in kind cluster====="
+
+  WORKDIR=$(pwd)
+  echo "Install prometheus operator."
+  echo "Current directory"
+  echo $(pwd)
+  cd ${WORKDIR}/..
+  git clone https://github.com/coreos/kube-prometheus.git
```
philipgough, line 101: url is outdated and will cause a redirect

collectors/metrics/cmd/metrics-collector/main.go
saswatamcode: It's what we use for thanos upstream (or aiming to, promql-engine already uses it). This returns the stacktrace which is super useful at times. github.com/pkg/errors did similar I think, but is no longer maintained.

collectors/metrics/test/integration/setup.sh
```diff
@@ -2,181 +2,182 @@
...
+  echo "Replace namespace with openshift-monitoring"
```
philipgough, line 102: im trying to understnad whats happening here? on KinD cluster, why do we need to use openshift-monitoring namespace with prom operator? 🤔

cicd-scripts/install-dependencies.sh
saswatamcode: Yup! Just fyi, these shellscript changes are from formatting, not explicit changes

collectors/metrics/pkg/forwarder/forwarder.go
saswatamcode: Yup, will add. I think there is also something else here that causes tests to fail, will take a look.

collectors/metrics/pkg/http/client.go
```diff
@@ -1,101 +0,0 @@
-package http
```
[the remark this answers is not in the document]
saswatamcode: Yeah, builds without it!

collectors/metrics/test/integration/prereq.sh
saswatamcode: Yeah, haven't gone through all the shellscripts yet. I might clean them up explicitly in another follow-up!

cicd-scripts/install-dependencies.sh
philipgough: yep, i get that, just flagging as I go but will pause on that and we can tackle bash maybe in isolation in a follow up

collectors/metrics/test/integration/setup.sh
saswatamcode: Yup!

collectors/metrics/cmd/metrics-collector/main.go
thibaultmg: I also prefer using the stdlib with `fmt.Errorf("%w")`. Regarding the stack trace, I don't see why we would need it. I mean, if so, it is a symptom of bad error handling that is not descriptive enough in the stack.
I don't like depending on those small libs while the stdlib does the job. But it's me 😉

saswatamcode:
### Errors

> I would avoid adding the new error package

I understand the reason for the new error package and not wanting to depend on small package, but there are a few of reasons, why I think it is generally better than combination of stdlib `fmt.Errorf` and `errors.New`,

- Wrapping errors with `fmt.Errorf` needs a %w which is super easy to forget to do, and easy to miss in review. I think we can construct some linter rule for it using some custom lint tools (faillint, golangci-lint don't do this I think), but that's extra check to maintain. Due to the nature of such a single arg function, it is prone to inconsistent wrapping.
- Relying on `errors.New` for new errors and then `fmt.Errorf` for wrapping doesn't look clean, and increases cognitive load.  I'd rather prefer using single package consistently for everything.

> Regarding the stack trace, I don't see why we would need it. I mean, if so, it is a symptom of bad error handling that is not descriptive enough in the stack.
I don't like depending on those small libs while the stdlib does the job.

- I get what you're saying about non-descriptive errors @thibaultmg, but I'd argue stack traces aren't exactly to describe the error or "cause" of the error, but rather to highlight the specific code path that the error resulted from, which is extremely useful when debugging issues with complex codebases like this one. And stacktraces are useful enough that people even created o11y signal from it i.e profiling. I'd rather have stacktraces and not use it, rather than not have stacktraces but need it. 😅 Wdyt?

And if anything, we maintain `efficientgo/core`, so there isn't much worry about governance.

### Bingo

> and bingo

Curious about this, why do we want to avoid bingo? Is it the extra files or something else you noticed @thibaultmg? 

### Logging

Also, I have an outstanding question apart from errors, which is logging. Seems like this repo uses `go-kit/log` (which we use upstream quite a lot in Prometheus and Thanos), and `klog` which is the standard k8s logger, quite a lot. Also, Kubebuilder-based operators usually use `zapr` which is what obo uses. I'd like to move to one consistent logger. What do you think we should use here then? My personal pick would be `go-kit/log` as we can do multiple formats and filter via log level, and it's the one current team is most familiar with. But what do you think @thibaultmg @philipgough? 🙂

douglascamata:
On error handling, I heavily agree with @thibaultmg's point. 

* I do not think adding a `%w` is easy to forget and miss in review. In fact, it becomes instinct after a few uses. Some of us in the team might see it as easy to forget only because we lack experience with the standard library's error wrapping. It's a matter of getting acquainted with the habit (less things to keep inside our heads = `cognitive_load--`).
* I don't think `fmt.Errorf` implies in extra cognitive load -- where would such load come from? It might not look clean for some, which is fair, but not a good reason to rely on a 3rd party library. In fact, we are better off using the Golang's standard library way, which can be improved over time. The Go team will take care of the backwards compatibility and migration for us.
* Extra: by using the official error wrapping of Go, in the future we might get stacktraces for free and won't need a 3rd party package for this. To me it's fine to use a 3rd party package for stacktraces **if** they prove helpful. More than once I already caught myself with a stacktrace that didn't help me pinpoint the issue.

On Bingo, I also agree with @thibaultmg's point. 

* It's not about "why avoid Bingo?" it's about "why use Bingo if `go install` is enough"? Bingo comes from a time where `go install` didn't exist. Nowadays we have a proper solution from the Go team which they will evolve over time together with all the rest of the toolchain. It's much more likely that a Go developer already knows `go install` (less things to keep inside our heads = `cognitive_load--`) than Bingo. 
* It's not only about how many files Bingo adds to the project, but also how much maintenance is required to upgrade tooling (more things to keep inside our heads = `cognitive_load++`). Meanwhile, `go install` is just a few lines in a README, Makefile or Shell script -- to upgrade something I change the version number and that's all. Also note that the commands in Bingo and `go install` have slightly different syntax sometimes (more things to keep inside our heads = `cognitive_load++`).

Overall, my opinion is: I like to stick with the standard library whenever possible. We should all be very familiar with it. When not possible, let's look for tools that are "de-facto" standards in the community. There are many good reasons for this, including but not limited to:

* By using the standard library and "de-facto" standard packages, we reduce onboarding time and cost of codebase because we will be using constructs that are familiar to every Go developer (less things to keep inside our heads = `cognitive_load--`). On the maintainership side, we don't have to create and maintain a package for this. The less code we have to maintain (less things to keep inside our heads = `cognitive_load--`), the more we can focus on what matters most.
* We see that the standard library is constantly evolving, getting new features and incorporating feedback from people using it. The Go team has been doing a great job at this and keeping backwards compatibility. It's a bad idea to lock ourselves out of it or push the responsibility of adopting updates to the standard library to someone else.

thibaultmg:
Hey, 

> Relying on errors.New for new errors and then fmt.Errorf for wrapping doesn't look clean, and increases cognitive load. I'd rather prefer using single package consistently for everything.

I think it is more of a matter of what tools you are used to. Because for me, it is this `errors.Wrap` that increases the cognitive load. I have to check the package to understand why it's there.
Regarding the stack trace, this can be added when debugging with 3 lines of code. And it happens rarely. So adding a specific dependency for that doesn't seem worth it to me.

> Curious about this, why do we want to avoid bingo? Is it the extra files or something else you noticed

I admit it brings useful things to the table. But I don't think we need it enough to not use `go install`. We could isolate these dev bins in a `.bin` dir in the package and have all tools consume these deps from here. The other useful things is the ability to selectively add deps. We could have one make target by dep. To me there are few things that bingo brings that cannot be done with standard tooling plus some small conventions.

> Also, I have an outstanding question apart from errors, which is logging.

I don't have strong opinions on this. In my personal projects I use zerolog which I like. With go 1.21 I would give a shot to `slog` from the standard lib. 

Overall I totally agree with @douglascamata remarks about the benefits of using the standard lib as much as possible. This is what I strive for in my projects. But again, It is not a blocker if the majority of the team enjoys these tools 😉

thibaultmg:
Bingo is not a blocker. You have already integrated it, let's go with that. After, I don't think anybody will spend time changing it just for the sake of removing it. 
Still good to remove the errors package 😄 .

douglascamata:
> That being said, I also do not feel strongly about blocking any alternatives to bingo in the context of this repo. So @thibaultmg / @douglascamata if you feel like bingo is causing you more difficulties or is hindering you in any way, please feel free to raise a follow-up to this PR with an alternative. 🙂

It's not about difficulties of Bingo. As I said before, it's more about using a non-standard tool when a standard one that can achieve the same things already exists. Bingo is far from a standard tool -- besides Thanos and grpc-middleware, only some other Red Hat projects are using it. Everyone else out there has figured how to live without it and it's not difficult. We can achieve with `go install` everything that Bingo does and the only "manual work" will be to add a variable in a Makefile (done once per lifetime of the project, not a burden at all). 

Not a blocker for me too.

collectors/metrics/pkg/metricfamily/transform_test.go
```diff
@@ -15,6 +15,7 @@ func family(name string, timestamps ...int64) *clientmodel.MetricFamily {
 	return families
 }
 
+//nolint:unparam
```
douglascamata, line 18: Why the `nolint` clause? The timestamp param is used.

collectors/metrics/pkg/metricsclient/metricsclient.go
```diff
@@ -41,42 +41,33 @@ const (
...
+	GaugeRequestRetrieve *prometheus.GaugeVec
+	GaugeRequestSend     *prometheus.GaugeVec
 }
 
 type PartitionedMetrics struct {
 	Families []*clientmodel.MetricFamily
 }
 
-func New(logger log.Logger, client *http.Client, maxBytes int64, timeout time.Duration, metricsName string) *Client {
+func New(logger log.Logger, m *ClientMetrics, client *http.Client, maxBytes int64, timeout time.Duration, metricsName string) *Client {
```
douglascamata, line 63: Why not rename this `m` param to something descriptive, like `metrics` or `clientMetrics`?

loaders/dashboards/pkg/util/grafana_util_test.go
```diff
@@ -32,7 +33,7 @@ func createFakeServer(t *testing.T) {
 	)
 	err := http.ListenAndServe(":3002", server3002)
 	if err != nil {
-		t.Fatal("fail to create internal server at 3002")
```
douglascamata, line 35: Same as my other comment: `*testing.T.Fatal` is very different from `log.Fatal`. I don't think this should be changed.

operators/endpointmetrics/controllers/observabilityendpoint/match_evaluator.go
```diff
@@ -3,9 +3,8 @@
 package observabilityendpoint
 
 import (
+	"golang.org/x/exp/slices"
```
douglascamata, line 6: Why not use the `slices` package from the standard library?

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
```diff
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strconv"
 
+	"golang.org/x/exp/slices"
```
douglascamata, line 11: This can be imported form the stdlib now.

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller_test.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	ocinfrav1 "github.com/openshift/api/config/v1"
 	hyperv1 "github.com/openshift/hypershift/api/v1alpha1"
+	"golang.org/x/exp/slices"
```
douglascamata, line 13: Can be imported from stdlib.

operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -10,13 +10,15 @@ import (
 	"reflect"
 	"time"
 
+	cerr "errors"
+
 	"github.com/go-logr/logr"
 	routev1 "github.com/openshift/api/route/v1"
+	"golang.org/x/exp/slices"
```
douglascamata, line 17: Can import this from stdlib.

operators/multiclusterobservability/controllers/multiclusterobservability/observatorium.go
```diff
@@ -6,15 +6,17 @@ package multiclusterobservability
...
 	"os"
 	"path"
 	"reflect"
 	"time"
 
+	"errors"
+
 	routev1 "github.com/openshift/api/route/v1"
 	obsv1alpha1 "github.com/stolostron/observatorium-operator/api/v1alpha1"
+	"golang.org/x/exp/slices"
```
douglascamata, line 19: Can import from stdlib.

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -5,14 +5,16 @@ package placementrule
...
 	"reflect"
 	"strings"
 	"sync"
 	"time"
 
+	"errors"
+
 	"github.com/go-logr/logr"
 	operatorv1 "github.com/openshift/api/operator/v1"
+	"golang.org/x/exp/slices"
```
douglascamata, line 17: Can import from stdlib.

operators/multiclusterobservability/pkg/certificates/cert_controller.go
```diff
@@ -12,6 +12,7 @@ import (
 	"reflect"
 	"time"
 
+	"golang.org/x/exp/slices"
```
douglascamata, line 15: Can import from stdlib.

operators/multiclusterobservability/pkg/certificates/certificates.go
```diff
@@ -194,6 +194,7 @@ func createCACertificate(cn string, caKey *rsa.PrivateKey) ([]byte, []byte, erro
 	return caKeyBytes, caBytes, nil
 }
 
+//nolint:unparam
```
douglascamata, line 197: Are there really unused params here?

operators/multiclusterobservability/pkg/config/azure_conf.go
```diff
@@ -4,9 +4,10 @@
 package config
 
 import (
-	"errors"
 	"strings"
```
douglascamata, line 8: Why is there always an empty line between the other stdlib imports and the import of `"errors"`? This doesn't seem formatted with `goimports`.

operators/multiclusterobservability/pkg/util/managedclusteraddon.go
```diff
@@ -93,6 +93,8 @@ func CreateManagedClusterAddonCR(c client.Client, namespace, labelKey, labelValu
 		}
 
 		// got the created managedclusteraddon just now, uopdating its status
+		// TODO(saswatamcode): Remove deprecated field
+		//nolint:staticcheck
```
douglascamata, line 97: Why the `nolint` clause?

tests/pkg/tests/observability_alert_test.go
```diff
@@ -28,6 +28,10 @@ import (
 	"github.com/stolostron/multicluster-observability-operator/tests/pkg/utils"
 )
 
+const (
+	TrueStr = "true"
+)
```
douglascamata, line 33: Do we really need an exported const here for this? 😅 

I see it being used in many places, but I've never seen this pattern before.

douglascamata (commented):
In general, I noticed a few things repeated in many files:

* Imports are not properly sorted. There seems to be an empty line before the imports of the `errors` package. Are you sure they were `goimports` sorted?
* Some usages of `testing.T.Fatal` were replaced with `log.Fatal` and these two have very different behaviors: the 1st will stop running that test and run the next one, while the 2nd will exit the whole Go process.
* The `slices` package is being imported from the 3rd party package `x/slices`. We can import it from the stdlib.

I might have forgotten to mention something due to the overall size of the PR and time taken to review it. Some files also have their own very specific comments.

I would love if this PR had been split into many smaller ones. Reviewing this was very difficult due to all the changes (adding periods to **all** comments, changing the indentation size in **all** shell scripts, fixing spacing in *all* comments, etc ,etc). Github didn't render many of the files and I had to open them up one by one... it took a very long time.

loaders/dashboards/pkg/controller/dashboard_controller_test.go
```diff
@@ -86,7 +86,7 @@ func createFakeServer(t *testing.T) {
 
 	err := http.ListenAndServe(":3001", server3001)
 	if err != nil {
-		t.Fatal("fail to create internal server at 3001")
+		stdlog.Fatal("fail to create internal server at 3001")
```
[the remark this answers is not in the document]
saswatamcode: `go vet` complains about the use of t.Fatalf in a non-test goroutine. log.Fatalf calls os.exit, while t.Fatal fails the test.
But you're right I think there is a better way here, will address in a follow-up.

operators/endpointmetrics/controllers/observabilityendpoint/match_evaluator.go
saswatamcode: So, I'm not entirely sure how this repo is built, and the Dockerfiles seem to use 1.20. `slices` was added to stdlib in 1.21 I think

operators/multiclusterobservability/pkg/certificates/certificates.go
saswatamcode: Yes, this one is valid `ou` is unused, but didn't want to refactor in this PR, as at first glance would need some logical changes. Added TODO.

operators/multiclusterobservability/pkg/config/azure_conf.go
saswatamcode: You're right, botched replace. Fixed it.

operators/multiclusterobservability/pkg/rendering/renderer.go
```diff
@@ -151,21 +148,3 @@ func (r *MCORenderer) Render() ([]*unstructured.Unstructured, error) {
...
-	if !ok {
-		return nil, fmt.Errorf("failed to find webhooks spec field")
-	}
-	webhook := webooks[0].(map[string]interface{})
-	clientConfig := webhook["clientConfig"].(map[string]interface{})
-	service := clientConfig["service"].(map[string]interface{})
-
-	service["namespace"] = mcoconfig.GetDefaultNamespace()
-	return u, nil
-}
```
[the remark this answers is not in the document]
saswatamcode: It was unused.

operators/multiclusterobservability/pkg/util/managedclusteraddon.go
saswatamcode: The reason is mentioned in the TODO. That field is deprecated, so probably needs some logical changes. For now it would fail staticcheck, due to it being marked deprecated.

tests/pkg/tests/observability_alert_test.go
saswatamcode: Ah, I remember something screamed at me in my IDE to do this. 😆 
But probably doesn't need to be exported.

thibaultmg (approved):
Looks Good! 🚀

thibaultmg (approved):
Looks Good! 🚀

douglascamata (approved):
LGTM 🚀

PR #1255, opened 2023-10-12
*: Bump prometheus client libs

saswatamcode:
Bumps client_golang, client_model and common to latest

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,9 +27,9 @@ require (
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.57.0
 	github.com/prometheus-operator/prometheus-operator/pkg/client v0.53.1
 	github.com/prometheus/alertmanager v0.25.0
-	github.com/prometheus/client_golang v1.16.0
-	github.com/prometheus/client_model v0.4.0
-	github.com/prometheus/common v0.42.0
+	github.com/prometheus/client_golang v1.17.0
+	github.com/prometheus/client_model v0.5.0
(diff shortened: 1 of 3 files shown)


subbarao-meduri (approved):
/lgtm

saswatamcode:
I think I keep hitting https://issues.redhat.com/browse/ACM-7864 with the CI i.e, `csv_abnormal` isn't present as a metric. I think this is a race condition in the test, where if cluster service version quickly succeeds, then we don't get the metric and a ginkgo test fails. I think we can ignore this abnormal state metric @subbarao-meduri?
This is a simple version client_golang bump PR, so it doesn't really change anything functionally, aside from how the MCO itself exposes metrics.

subbarao-meduri (approved):
/lgtm

subbarao-meduri:
> I think I keep hitting https://issues.redhat.com/browse/ACM-7864 with the CI i.e, `csv_abnormal` isn't present as a metric. I think this is a race condition in the test, where if cluster service version quickly succeeds, then we don't get the metric and a ginkgo test fails. I think we can ignore this abnormal state metric @subbarao-meduri? This is a simple version client_golang bump PR, so it doesn't really change anything functionally, aside from how the MCO itself exposes metrics.
> 
> I've removed it in [0fca5fb](https://github.com/stolostron/multicluster-observability-operator/commit/0fca5fbac486ab9b215438829b33326c24ad0a37) to test

This makes sense, @saswatamcode - There have been intermittent timing related e2e failures in tests, and this is likely one of them.

PR #1263, opened 2023-10-19
[ACM-8232] call  addonMgr.start asynchronously

subbarao-meduri:
A workaround suggested by @elgnay  is to call `addonMgr.start()` asynchronously.
Together with  fix https://github.com/stolostron/multiclusterhub-operator/pull/1169/ should resolve the issue.

I have verified that this works in my local env.

### DIFF
--- a/operators/multiclusterobservability/pkg/certificates/cert_controller.go
+++ b/operators/multiclusterobservability/pkg/certificates/cert_controller.go
@@ -58,11 +58,15 @@ func Start(c client.Client, ingressCtlCrdExists bool) {
 		os.Exit(1)
 	}
 
-	err = addonMgr.Start(context.TODO())
-	if err != nil {
-		log.Error(err, "Failed to start addon manager")
-		os.Exit(1)
-	}
+	go func() {
+		ctx := context.TODO()
+		err = addonMgr.Start(ctx)
+		if err != nil {
+			log.Error(err, "Failed to start addon manager")
+			os.Exit(1)
+		}
+		<-ctx.Done()
+	}()
 
 	kubeClient, err := kubernetes.NewForConfig(ctrl.GetConfigOrDie())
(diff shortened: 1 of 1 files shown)


operators/multiclusterobservability/pkg/certificates/cert_controller.go
```diff
@@ -58,11 +58,15 @@ func Start(c client.Client, ingressCtlCrdExists bool) {
 		os.Exit(1)
 	}
 
-	err = addonMgr.Start(context.TODO())
-	if err != nil {
-		log.Error(err, "Failed to start addon manager")
-		os.Exit(1)
-	}
+	go func() {
+		ctx := context.TODO()
```
saswatamcode, line 62: Probably needs some centralized context, in the future

saswatamcode (commented):
/lgtm as far as I understand this.

saswatamcode:
The lint check is fixed by #1262

douglascamata:
The `goconst` linter's failing again with the `"true"` string. Maybe we should disable this linter? (cc @saswatamcode)

saswatamcode:
>The goconst linter's failing again with the "true" string. Maybe we should disable this linter? (cc @saswatamcode)

Fixed in #1262! Just needs to be rebased for check to pass.
Fwiw, all of these "true" string checks, actually just check if a env var is set or not. We can refactor them out and remove that string

elgnay:
@subbarao-meduri I think we no longer need this change. https://github.com/stolostron/multiclusterhub-operator/pull/1169/ is enough to fix the issue.

douglascamata:
If we indeed don't need this, as @elgnay says, I prefer to avoid the change. 😬

subbarao-meduri:
Closing this PR, as confirmed not needed.

PR #1264, opened 2023-10-19
[ACM-3748] Set max source resolution to auto in Grafana datasource

douglascamata:
Sets `max_source_resolution=auto` in the Grafana datasource's `customQueryParameters`. This should force the underlying Thanos Query Frontend to perform additional queries for downsampled data in case of empty responses.


douglascamata:
Tested this with a local cluster and it worked! 🎉 

image

subbarao-meduri (approved):
/lgtm

PR #1265, opened 2023-10-19
[ACM-8259] fix mco status update race condition

thibaultmg:
Fix race condition in mco controller StartStatusUpdate function.


saswatamcode (approved):
Thanks! I think this makes sense!

thibaultmg:
Is this a regression or flaky test?
If this is a regression, let's discard this PR and we'll fix as we work more on the code base. This is just a quick fix as I was reading the code.

PR #1270, opened 2023-10-27
[release-2.8] [ACM-3748] set max source resolution to auto

subbarao-meduri:
Backport fix to service streams


subbarao-meduri:
New commits got picked up in cherry-pick? will redo.

PR #1274, opened 2023-10-31
Non-root containers 

douglascamata:
The objective is to clear out all [the security hotspots from project as reported by SonarCloud](https://sonarcloud.io/project/security_hotspots?id=open-cluster-management_multicluster-observability-operator.).


subbarao-meduri (approved):
/lgtm

PR #1276, opened 2023-11-02
Rebuild image to fix `gosec` scan results in SonarCloud

douglascamata:
This should pick up a newer version of the image-builder, with an updated `gosec`. With this new version it should properly skip a false-positive that previous versions don't ignore.


saswatamcode (approved):

saswatamcode:
@douglascamata we can merge directly if you're blocked here.

douglascamata:
Let's do this then, @saswatamcode 😄

PR #1289, opened 2023-12-06
add watch logic for addondeployment

coleenquadros:
https://issues.redhat.com/browse/RHOBS-949
https://issues.redhat.com/browse/ACM-8312

From what I understand is that the placement controller has no logic to watch over the changes to addondeployment config. So any update to it will not be picked up by managerclusteraddon and clustermanagementaddon which are watched already but they only hold a ref to addondeploymentconfig and not the actual config.

The PR adds the watch logic in the placement controller to monitor changes in addondeployment config and sets up the request to reconcile and also apply change to all managed clusters

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
+++ b/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
@@ -324,6 +324,30 @@ func createManifestWorks(
 					container.Env[j].Value = strconv.FormatBool(installProm)
 				}
 			}
+			// If ProxyConfig is specified as part of addonConfig, set the proxy envs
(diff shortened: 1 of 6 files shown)


douglascamata:
@coleenquadros there's an e2e test failing. Can you have a look, please?

douglascamata (approved):
Looks good to me. Would love to achieve that 70% test coverage that SonarCloud is asking for though. It's fine if it's not feasible because we can't add 100% coverage of the proxy scenarios.

coleenquadros:
Testing working as expected - changed the proxy setting values in addondeploymentconfig  and the placement controller reconciled to pick up the new proxy settings

PR #1297, opened 2023-12-14
Ensure conflict status errors are surfaced in metrics-collector

saswatamcode:
While this is a 4xx statuscode, it is important to log errors, and not log "metrics pushed successfully", as the request did fail.


philipgough (approved):
lgtm

PR #1300, opened 2024-01-02
Set GOFLAGS to empty string for RHTAP builds

subbarao-meduri:
Required by [ACM-7181](https://issues.redhat.com/browse/ACM-7181)

The golang builder defaults to vendor. Need to set GOFLAGS to empty string for RHTAP builds.


Kyl-Bempah (approved):
LGTM

PR #1304, opened 2024-01-29
Change datasource timeout config for Grafana and add matching timeout…

philipgough:
… annotation for Route

The previous configuration `queryTimeout` is not valid as per https://grafana.com/docs/grafana/v8.5/administration/provisioning/#json-data (it should be timeout defined as seconds).

The Route needs to have a corresponding annotation to avoid queries timing out after 30s, which is the default in OpenShifts HAProxy config

### DIFF
--- a/operators/multiclusterobservability/controllers/multiclusterobservability/grafana.go
+++ b/operators/multiclusterobservability/controllers/multiclusterobservability/grafana.go
@@ -55,9 +55,10 @@ type GrafanaDatasource struct {
 }
 
 type JsonData struct {
-	TLSAuth               bool   `yaml:"tlsAuth,omitempty"`
-	TLSAuthCA             bool   `yaml:"tlsAuthWithCACert,omitempty"`
-	QueryTimeout          string `yaml:"queryTimeout,omitempty"`
+	TLSAuth   bool `yaml:"tlsAuth,omitempty"`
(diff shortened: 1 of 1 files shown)


operators/multiclusterobservability/controllers/multiclusterobservability/grafana.go
```diff
@@ -55,9 +55,10 @@ type GrafanaDatasource struct {
 }
 
 type JsonData struct {
-	TLSAuth               bool   `yaml:"tlsAuth,omitempty"`
-	TLSAuthCA             bool   `yaml:"tlsAuthWithCACert,omitempty"`
-	QueryTimeout          string `yaml:"queryTimeout,omitempty"`
+	TLSAuth   bool `yaml:"tlsAuth,omitempty"`
```
jacobbaungard, line 58: Nit, but would be nice to keep the formatting here is consistent.

philipgough: Ack, it is formatted correctly via gofmt however the comment on this particular field throws off the indentation. Since this is a bug fix I'd just keep it minimal here as is, but I intend on doing a follow up which reviews and explains all the fields in this file which will address this!

douglascamata (approved):
/lgtm

PR #1307, opened 2024-01-31
Change datasource timeout config for Grafana and add matching timeout…

philipgough:
… annotation for Route (#1304)


jacobbaungard (approved):
/lgtm

PR #1312, opened 2024-02-06
[ACM-9321] Stop firing IngressWithoutClassName alert

subbarao-meduri:
Per https://issues.redhat.com/browse/OCPBUGS-18537, define .spec.ingressClassName so IngressWithoutClassName alert does not fire.


bjoydeep (approved):

douglascamata (approved):
/lgtm

PR #1316, opened 2024-02-08
Hub metrics collector feature

coleenquadros:


operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -136,14 +136,21 @@ func (r *PlacementRuleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, nil
 	}
 
+	//Check for MulticlusterGlobalHub CRD
+	mcghCrdExists := r.CRDMap[config.MCGHCrdName]
+	//if Multicluster Global hub exists we block metrics-collector creation in spokes
+	if mcghCrdExists {
+		mco.Spec.ObservabilityAddonSpec.EnableMetrics = false
+	}
+
```
subbarao-meduri, line 145: @coleenquadros when MCGH CRD is detected, we don't want to standup hub Observability stack either (including Thanos). Not just spokes. We should do this check in https://github.com/stolostron/multicluster-observability-operator/blob/main/operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go and bail out if detected.

PR #1323, opened 2024-02-12
alert: make ACMRemoteWriteError less prone to flakes/flapping

philipgough:
This is a first pass at making this alert less noisy without introducing too much change on it initially.  
An alert should be actionable but in many cases here, the response codes can be in the 4xx unrecoverable range which means the alert can fire when essentially there is no data loss or the customer cannot intervene.

https://issues.redhat.com/browse/ACM-4820


douglascamata (approved):
/lgtm

PR #1324, opened 2024-02-12
Fix HCP dashboards

rokej:
Adding the new HCP dashboards YAMLs to kustomization.yaml and adding a value filter to the HCP list table.


subbarao-meduri (approved):
/lgtm

PR #1325, opened 2024-02-13
Set security context of workloads

jacobbaungard:
This commit sets the following security context of all workloads
directly handled by the operator:

```
securityContext:
    privileged: false
    readOnlyRootFilesystem: true
```

this as required by: RHOBS-1001


operators/multiclusterobservability/manifests/base/observatorium/operator.yaml
```diff
@@ -23,37 +23,40 @@ spec:
...
-          - podAffinityTerm:
-              labelSelector:
-                matchExpressions:
-                - key: app.kubernetes.io/name
-                  operator: In
-                  values:
-                  - observatorium-operator
-              topologyKey: topology.kubernetes.io/zone
-            weight: 70
+            - podAffinityTerm:
```
moadz, line 26: Any reason why the yaml is being formatted differently?

moadz (approved):

operators/multiclusterobservability/manifests/base/observatorium/operator.yaml
jacobbaungard: There were some formatting inconsistencies before (or at least my eyes struggled with the formatting), so I ran the files I needed to change through `yamllint` as I found it hard to make changes correctly otherwise.

PR #1328, opened 2024-02-14
fix a metric query to use instant value

rokej:
One of the queries needs to be fixed to use an instant time series value.


subbarao-meduri (approved):
/lgtm

PR #1329, opened 2024-02-15
[ACM-8543] Tag Observatorium CR with a hash of all the config it uses

douglascamata:
This should make the CR's hash change. As the CR's hash is used to annotate the Observatorium API pods, these will be rolled out and pick up any new value from the configmaps or secrets.


thibaultmg (approved):
/lgtm

PR #1330, opened 2024-02-15
[ACM-9653] Updating prometheus-k8s CRD's to 0.68

moadz:
Currently `global` variables defined for `scrapeInterval` and `evaluationInterval` are blank strings, and causes prometheus pod to panic on init as the duration cannot be blank. This was a regression due to an upstream CRD change that was not applied in endpoint operator.

This was validated by editing the `prometheus/k8s` CR to supply a value for the flags, pod starts as usual.

**Cause**

There was a change in the CRD [default types in upstream prometheus operator](https://github.com/prometheus-operator/prometheus-operator/issues/5197), when we [upgraded to v0.68 in our stolostron fork](https://github.com/stolostron/prometheus-operator/pull/32), we did not update the MCO CRD's to account for the null strings. 

**Fix**
This PR updates the CRD to include the `scrapeInterval` kube-builder default from upstream prometheus-operator. Prior to this change, it did not exist in the CR and thus the default was `""`, causing the prom panic.
```
                scrapeInterval:
                  default: 30s
```

**Impact**

All *KS clusters that rely on prometheus-k8s operator for their MCO deployed monitoring stack.


subbarao-meduri (approved):
/lgtm

PR #1334, opened 2024-02-16
[ACM-8543] Fix reconciliation on Observatorium CR for update scenario

douglascamata:
Changes to the config hash approach:

* Gave up on hashing the secrets (mostly certs): they are already used by the embed `certcontroller` to add a label to the `observatorium-api` and its `kube-rbac-proxy` sidecar. I prefer to avoid the concurrency issue that comes with updating the Observatorium CR, which will also trigger the Observatorium pods to change. 
  Ideally this mechanism should live outside of the `certcontroller`. It's already watching for **all** secrets in the operator's namespace -- not only certificates. It could also be made to watch configmaps.
* I changed such mechanism above to use an `strategicMerge` **patch** instead of full update, as it only adds a label to a list.

### DIFF
--- a/operators/multiclusterobservability/controllers/multiclusterobservability/observatorium.go
+++ b/operators/multiclusterobservability/controllers/multiclusterobservability/observatorium.go
@@ -7,6 +7,7 @@ package multiclusterobservability
 import (
 	"bytes"
 	"context"
+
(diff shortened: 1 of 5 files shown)


philipgough (approved):
/lgtm

philipgough (approved):
/lgtm

PR #1340, opened 2024-02-20
Hub metrics collector feature

coleenquadros:
Ticket - https://issues.redhat.com/browse/ACM-8509
This PR refactors the deployment of the metrics-collector in the hub/local-cluster. We want to ensure that the metrics-collector on the hub runs independent of hubselfmanagement toggle. 

- The placement controller of MCO operator treats the hub cluster as a special case, and directly creates the endpoint operator and the related resources (mTLS certificates, metrics lists, hub-info-secret that drives CMO configuration changes), in the open-cluster-management-observability namespace, instead of creating them as ManifestWork objects. It further ensures that it has no dependencies on the add-on framework. There will no longer be an Observability add-on on the hub cluster. In this respect, the hub cluster (local-cluster managed cluster) will be treated as a special case. In turn, the endpoint operator:
- Creates a metrics collector deployment always in open-cluster-management-observability  namespace on the hub. This namespace will be different from open-cluster-management-addon-observability used in all other spokes
- Makes modifications to CMO operator configuration to ensure that hub cluster Prometheus continues to forward alerts to ACM Alert Manager
- The MCO operator placement controller ignores managed cluster create/update/delete events for local-cluster, so observability add-on is not creating managed cluster components in the old way. There will be only one way the endpoint operator is created on the hub.
- The MCO operator takes ownership of refreshing mTLS certificates for the collector on the hub, and restarts metrics collector deployment when the certificates expire/change/refreshed.
- The MCO operator will be responsible for installing these hub components when ACM is installed and Observability is enabled.
- The MCO operator will be responsible for upgrading these components when ACM is upgraded.
- The MCO operator will be responsible for uninstalling these components when ACM is uninstalled or Observability is disabled.

Since we no longer use observability add-on on the hub, we lose the ability to report failure to send metrics as via observability addon status. To compensate for this:

- Metrics collector should publish metrics related to failed metrics upload count. 
- A local alert is raised when the upload metrics count exceeds a defined threshold
- The count should be reset after the error is cleared and metrics are propagating again. 

Known issues (to be tracked as defects)
1. Certificate rotation for hub metrics collector after it expires in a year 
2. Image replacement annotations don't work for hub endpoint operator and metrics collector (dev feature)
3. Some scenarios where "local-cluster" is still not available in grafana
4. Retrofit e2e tests (dev feature)


operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
```diff
@@ -169,7 +179,11 @@ func (r *ObservabilityAddonReconciler) Reconcile(ctx context.Context, req ctrl.R
 		if err != nil {
 			if errors.IsNotFound(err) {
 				log.Error(err, "OCP prometheus service does not exist")
-				util.ReportStatus(ctx, r.Client, obsAddon, "NotSupported")
+				// ACM 8509: Special case for hub/local cluster metrics collection
+				// We do not report status for hub endpoint operator
+				if !hubMetricsCollector {
+					util.ReportStatus(ctx, r.Client, obsAddon, "NotSupported")
```
marcolan018, line 185: is it better to check whether current deployment is hub metrics collector inside the func ReportStatus()?

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
```diff
@@ -243,6 +257,20 @@ func (r *ObservabilityAddonReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		_, err := updateMetricsCollectors(
+			ctx,
+			r.Client,
+			obsAddon.Spec,
+			*hubInfo, clusterID,
+			clusterType,
+			1,
+			forceRestart)
+		if err != nil {
+			log.Info("Error while creating hub metrics collector")
```
marcolan018, line 271: should we use log.Error here?

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -423,10 +457,135 @@ func createManifestWorks(
...
+		log.Info("Creating resource for hub metrics collection", "cluster", clusterName)
+		err = createUpdateResourcesForHubMetricsCollection(c, manifests)
+	} else {
+		err = createManifestwork(c, work)
+	}
+
 	return err
 }
 
+func createCSR() ([]byte, []byte) {
```
marcolan018, line 472: seems it's not a correct choice to use CSR flow here. Have you check the possibility to use the functions in https://github.com/stolostron/multicluster-observability-operator/blob/main/operators/multiclusterobservability/pkg/certificates/certificates.go?

operators/endpointmetrics/controllers/observabilityendpoint/metrics_collector.go
```diff
@@ -314,6 +322,22 @@ func createDeployment(params CollectorParams) *appsv1.Deployment {
...
+		metricsCollectorDep.Spec.Template.Spec.Containers[0].Env = append(metricsCollectorDep.Spec.Template.Spec.Containers[0].Env,
+			corev1.EnvVar{
+				Name:  "STANDALONE",
+				Value: "true",
+			})
+		//Since there is no obsAddOn for hub-metrics-collector, we need to set the resources here
+		metricsCollectorDep.Spec.Template.Spec.Containers[0].Resources = corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("100Mi"),
+				corev1.ResourceCPU:    resource.MustParse("10m"),
```
clyang82, line 336: I think the resource should be from mco spec instead of hardcode here. Does it make sense?

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
clyang82: maybe just generate a private key and then sign it to have the client cert.

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -93,6 +96,25 @@ func (r *PlacementRuleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	reqLogger := log.WithValues("Request.Namespace", req.Namespace, "Request.Name", req.Name)
 	reqLogger.Info("Reconciling PlacementRule")
 
+	// ACM 8509: Special case for hub/local cluster metrics collection
+	// We want to ensure that the local-cluster is always in the managedClusterList
+	// In the case when hubSelfManagement is enabled, we will delete it from the list and modify the object
+	// to cater to the use case of deploying in open-cluster-management-observability namespace
+	delete(managedClusterList, "local-cluster")
```
clyang82, line 103: in future, we should totally remove the dependencies of `local-cluster` and `obsaddon` for this case.

clyang82 (commented):
Did you check if we create the obsaddon in open-cluster-management-observability namespace or not? I think no. but just double check. thanks.

operators/endpointmetrics/controllers/observabilityendpoint/metrics_collector.go
coleenquadros: Yes but if there is no obsaddon created for local-cluster, how would it pick the default mco resources?

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -423,10 +457,135 @@ func createManifestWorks(
...
+	})
+
+	privateKey := pem.EncodeToMemory(&pem.Block{
+		Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(keys),
+	})
+
+	return csr, privateKey
+}
+
+func createMtlsCertSecretForHubCollector(c client.Client) error {
```
elgnay, line 503: The overall procedure to create client certificate looks good. The client cert may expire, cert rotation is required. You can find some example code from registration agent: https://github.com/open-cluster-management-io/ocm/blob/1c3cb033b096c0e95575dfbf6d5d12126f5d7e57/pkg/registration/clientcert/cert_controller.go#L290

operators/endpointmetrics/pkg/util/status.go
```diff
@@ -34,7 +34,10 @@ var (
 	}
 )
 
-func ReportStatus(ctx context.Context, client client.Client, i *oav1beta1.ObservabilityAddon, t string) {
+func ReportStatus(ctx context.Context, client client.Client, i *oav1beta1.ObservabilityAddon, t string, reportStatus bool) {
+	if !reportStatus {
+		return
+	}
```
douglascamata, line 40: It seems very weird to have a function that takes a boolean that makes such function do nothing. I think this check should be done by the callers of `ReportStatus` before they decide wether to call it.

operators/multiclusterobservability/controllers/placementrule/hub_ocp_monitoring_util.go
```diff
@@ -0,0 +1,266 @@
...
+	clusterRoleBindingName         = "metrics-collector-view"
+	caConfigmapName                = "metrics-collector-serving-certs-ca-bundle"
+	hubMetricsCollectionNamespace  = "open-cluster-management-observability"
+	etcdServiceMonitor             = "acm-etcd"
+	kubeApiServiceMonitor          = "acm-kube-apiserver"
+)
+
+// revertClusterMonitoringConfig reverts the configmap cluster-monitoring-config and relevant resources
+// (observability-alertmanager-accessor and hub-alertmanager-router-ca) for the openshift cluster monitoring stack.
+func RevertHubClusterMonitoringConfig(ctx context.Context, client client.Client) error {
```
douglascamata, line 40: Do we need all these logs at `info` level? Won't them spam too much?

operators/multiclusterobservability/controllers/placementrule/hub_ocp_monitoring_util.go
```diff
@@ -0,0 +1,266 @@
...
+	if foundClusterMonitoringConfiguration.PrometheusK8sConfig == nil {
+		log.Info(
+			"configmap data doesn't key: prometheusK8s, no need action",
+			"name",
+			clusterMonitoringConfigName,
+			"key",
+			clusterMonitoringConfigDataKey,
+		)
+		return nil
+	} else {
```
douglascamata, line 118: You don't need this `else` here.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -423,10 +447,150 @@ func createManifestWorks(
...
+	//		return err
+	//	}
+	//}
+	//if isHypershift {
+	//	err = DeleteServiceMonitors(context.TODO(), c)
+	//	if err != nil {
+	//		log.Error(err, "Failed to delete service monitors for hub metrics collection")
+	//		return err
+	//	}
+	//}
```
douglascamata, line 590: Delete this?

operators/multiclusterobservability/controllers/placementrule/manifestwork_test.go
```diff
@@ -34,6 +35,10 @@ const (
 	workSize       = 13
 )
 
+func init() {
+	os.Setenv("UNIT_TEST", "true")
+}
```
douglascamata, line 40: Are you sure this is a good idea? Once this env var is set, I see nothing unsetting it. 🤔

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
```diff
@@ -75,57 +81,66 @@ func (r *ObservabilityAddonReconciler) Reconcile(ctx context.Context, req ctrl.R
 	log := log.WithValues("Request.Namespace", req.Namespace, "Request.Name", req.Name)
 	log.Info("Reconciling")
 
-	// Fetch the ObservabilityAddon instance in hub cluster
-	hubObsAddon := &oav1beta1.ObservabilityAddon{}
-	err := r.HubClient.Get(ctx, types.NamespacedName{Name: obAddonName, Namespace: hubNamespace}, hubObsAddon)
-	if err != nil {
-		hubClient, obsAddon, err := util.RenewAndRetry(ctx)
+	isHypershift := true
```
douglascamata, line 84: Shouldn't this be false by default for safety?

operators/endpointmetrics/controllers/observabilityendpoint/metrics_collector.go
```diff
@@ -112,7 +118,12 @@ func getCommands(params CollectorParams) []string {
 	for _, group := range params.allowlist.CollectRuleGroupList {
 		if group.Selector.MatchExpression != nil {
 			for _, expr := range group.Selector.MatchExpression {
-				if !evluateMatchExpression(expr, clusterID, params.clusterType, params.obsAddonSpec, params.hubInfo,
+				if hubMetricsCollector {
+					if !evluateMatchExpression(expr, clusterID, params.clusterType, params.hubInfo,
+						params.allowlist, params.nodeSelector, params.tolerations, params.replicaCount) {
+						continue
+					}
+				} else if !evluateMatchExpression(expr, clusterID, params.clusterType, params.obsAddonSpec, params.hubInfo,
```
douglascamata, line 126: This `else` is not needed because of the `continue` above.

operators/endpointmetrics/controllers/observabilityendpoint/metrics_collector.go
```diff
@@ -171,6 +182,11 @@ func getCommands(params CollectorParams) []string {
 }
 
 func createDeployment(params CollectorParams) *appsv1.Deployment {
+	// falsePtr := false
```
douglascamata, line 185: Delete this?

operators/endpointmetrics/controllers/observabilityendpoint/metrics_collector.go
```diff
@@ -328,6 +440,186 @@ func createDeployment(params CollectorParams) *appsv1.Deployment {
...
+// static:
+//   - path: /metrics
+//     resourceRequest: false
+//     user:
+//       name: system:serviceaccount:openshift-monitoring:prometheus-k8s
+//     verb: get`,
+// 		},
+// 	}
+// 	return secret
+// }
```
douglascamata, line 465: Delete this?

douglascamata (commented):
Lots of code comments that, IMHO, should be uncommented or removed.

I'm also concerned about lots of logging at info level that maybe should be removed or have the level raised a bit.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
saswatamcode: Sorry this was me, trying to get it to work

operators/endpointmetrics/pkg/util/status.go
coleenquadros: it was previously implemented that way where we would check if we needed to call  but report status was called in multiple places and in a previous review comment was asked to check for obsaddon and report status accordingly

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
```diff
@@ -10,12 +10,15 @@ import (
 	"os"
 	"strconv"
 
+	operatorutil "github.com/stolostron/multicluster-observability-operator/operators/pkg/util"
```
philipgough, line 13: this can be moved back down to avoid unneeded diff

operators/multiclusterobservability/controllers/placementrule/manifestwork_test.go
coleenquadros: I think this is only run in test runs and wont be set in the actual deployment

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -423,10 +447,234 @@ func createManifestWorks(
...
+	//	gvk := obj.GetObjectKind().GroupVersionKind()
+	//	switch gvk.Kind {
+	//	case "Namespace", "ObservabilityAddon":
+	//		// We do not want to create ObservabilityAddon and namespace open-cluster-management-add-on observability for hub cluster
+	//		continue
+	//	case "ClusterRole", "ClusterRoleBinding", "CustomResourceDefinition":
+	//	default:
+	//		obj.SetNamespace(config.GetDefaultNamespace())
+	//	}
+	//
```
thibaultmg, line 511: Why are we ignoring other types of errors? Isn't risky?

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -423,10 +447,234 @@ func createManifestWorks(
...
+	//		role := obj.(*rbacv1.ClusterRoleBinding)
+	//		if len(role.Subjects) > 0 {
+	//			role.Subjects[0].Namespace = config.GetDefaultNamespace()
+	//		}
+	//	}
+	//
+	//	switch obj.GetObjectKind().GroupVersionKind().Kind {
+	//	case "Deployment":
+	//		currentObj = &appsv1.Deployment{}
+	//	case "Secret":
```
thibaultmg, line 522: Same as above

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -423,10 +447,234 @@ func createManifestWorks(
...
+	//			currentConfigMap := currentObj.(*corev1.ConfigMap)
+	//			if !reflect.DeepEqual(obj.Data, currentConfigMap.Data) {
+	//				needsUpdate = true
+	//			}
+	//		}
+	//
+	//		if needsUpdate {
+	//			err = c.Update(context.TODO(), obj)
+	//			if err != nil {
+	//				log.Error(err, "Failed to update resource", "kind", obj.GetObjectKind().GroupVersionKind().Kind)
```
thibaultmg, line 571: Can we delete this?

operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -360,11 +362,15 @@ func (r *MultiClusterObservabilityReconciler) initFinalization(
 	if mco.GetDeletionTimestamp() != nil && slices.Contains(mco.GetFinalizers(), resFinalizer) {
 		log.Info("To delete resources across namespaces")
 		// clean up the cluster resources, eg. clusterrole, clusterrolebinding, etc
+		operatorconfig.IsMCOTerminating = true
```
thibaultmg, line 365: We don't need that aren't we? The caller sets it using returned boolean.

operators/multiclusterobservability/controllers/placementrule/hub_ocp_monitoring_util.go
```diff
@@ -0,0 +1,266 @@
...
+			return nil
+		} else {
+			log.Error(err, "failed to check configmap", "name", clusterMonitoringConfigName)
+			return err
+		}
+	}
+
+	// do not touch the configmap if are not already a manager
+	touched := false
+	log.Info("checking field.Manager")
```
thibaultmg, line 59: A lot of these logs are debug level to me. I think we are generating too much noise at the info level.

operators/multiclusterobservability/controllers/placementrule/hub_ocp_monitoring_util.go
```diff
@@ -0,0 +1,266 @@
...
+			"name",
+			clusterMonitoringConfigName,
+			"key",
+			clusterMonitoringConfigDataKey,
+		)
+		return nil
+	}
+	foundClusterMonitoringConfigurationJSON, err := yaml.YAMLToJSON([]byte(foundClusterMonitoringConfigurationYAML))
+	if err != nil {
+		log.Error(err, "failed to transform YAML to JSON", "YAML", foundClusterMonitoringConfigurationYAML)
```
thibaultmg, line 92: Just a global remark regarding the logs again that doesn't need to be changed here.
We are logging a lot errors at nested functions level. This function is called 3 levels down the Reconcile function and we are logging returned errors at every one of them (except in the Reconcile function...). So we are generating 3 logs for the same error. We could instead log only in the `Reconcile` function, while wrapping errors with the additional info in sub functions (except in some rare cases were details cannot be wrapped in the error).

operators/multiclusterobservability/controllers/placementrule/hub_ocp_monitoring_util.go
thibaultmg, line 118: This nested else could be avoided for readability

thibaultmg: +1

operators/multiclusterobservability/controllers/placementrule/manifestwork_test.go
douglascamata: I'm not sure wether this will run only on test because it's not in a separate `placementrule_test` package, but it shares the same `placementrule` package as the implementation.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -615,6 +863,9 @@ func generateMetricsListCM(client client.Client) (*corev1.ConfigMap, *corev1.Con
 
 func getObservabilityAddon(c client.Client, namespace string,
 	mco *mcov1beta2.MultiClusterObservability) (*mcov1beta1.ObservabilityAddon, error) {
+	if namespace == config.GetDefaultNamespace() {
```
thibaultmg, line 866: Could we add a comment explaining this check?

operators/multiclusterobservability/controllers/placementrule/obsaddon.go
```diff
@@ -57,6 +59,9 @@ func deleteObsAddon(c client.Client, namespace string) error {
 }
 
 func createObsAddon(c client.Client, namespace string) error {
+	if namespace == config.GetDefaultNamespace() {
```
thibaultmg, line 62: I would appreciate a cooment here also explaining the reason for this check.

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -93,6 +94,25 @@ func (r *PlacementRuleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	if _, ok := managedClusterList["local-cluster"]; !ok {
```
thibaultmg, line 102: Do we need this check? I understand we just removed the key before.

PR #1352, opened 2024-02-28
[ACM-6112] Fix invalid uwl service monitors for hypershift hosted etcd and api-server

thibaultmg:
Fixes https://issues.redhat.com/browse/ACM-6112

New strategy is to retrieve service monitors created by Hypershift itself and duplicate needed configuration like TLS to be more robust to changes. We now depend only on the name of the original serviceMonitors.
Some integration tests are added for the Hypershift case.

I took this opportunity to refactor and isolate certain functions from the `observabilityendpoint` package, enhancing the code's maintainability.


operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
```diff
@@ -153,6 +159,7 @@ func (r *ObservabilityAddonReconciler) Reconcile(ctx context.Context, req ctrl.R
 		hubSecret,
 	)
 	if err != nil {
+		log.Error(err, "Failed to get hub info secret")
```
philipgough, line 162: should we log the error here?

operators/endpointmetrics/pkg/hypershift/hypershift.go
```diff
@@ -0,0 +1,276 @@
...
+func getKubeServiceMonitor(ctx context.Context, c client.Client, namespace, clusterID, clusterName string) (*promv1.ServiceMonitor, error) {
+	// Get the hypershift's api-server service monitor and replicate some of its settings
+	hypershiftKubeSM := &promv1.ServiceMonitor{}
+	if err := c.Get(ctx, types.NamespacedName{Namespace: namespace, Name: ApiServerSmName}, hypershiftKubeSM); err != nil {
+		return nil, fmt.Errorf("failed to get hypershift's kube-apiserver ServiceMonitor: %w", err)
+	}
+
+	if len(hypershiftKubeSM.Spec.Endpoints) != 1 {
+		return nil, fmt.Errorf("hypershift's kube-apiserver ServiceMonitor has more than one endpoint") // safe check
+	}
```
douglascamata, line 202: What if it has zero endpoints? Same applies to other functions in this file that do a similar check.

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
philipgough: same comment throughout actually to be honest. If we don't intend to log the message as to why we failed perhaps we can just wrap it and return the errors and log at the top level?

operators/endpointmetrics/pkg/openshift/openshift.go
```diff
@@ -0,0 +1,191 @@
...
+const (
+	ClusterRoleBindingName = "metrics-collector-view"
+	CaConfigmapName        = "metrics-collector-serving-certs-ca-bundle"
+	OwnerLabelKey          = "owner"
+	OwnerLabelValue        = "observabilityaddon"
+)
+
+var (
+	log = ctrl.Log.WithName("controllers").WithName("ObservabilityAddon")
+)
```
douglascamata, line 32: Can we avoid this global? It has a pretty common name that is often used for imports (`log`) and could be passed via parameter to the functions (even via context).

operators/endpointmetrics/controllers/observabilityendpoint/testdata/crd/hostedclusters.hypershift.yaml
```diff
@@ -0,0 +1,7981 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    service.beta.openshift.io/inject-cabundle: "true"
+  creationTimestamp: "2023-09-20T18:09:38Z"
+  generation: 4
+  name: hostedclusters.hypershift.openshift.io
+  uid: cc3bc7a2-fa66-4bca-93b2-1643bfe3e61d
```
philipgough, line 9: Is there some context around why we need these CRDs and under what circumstance? 
In any case we likely would not want to hardcode a UUID in the resources themselves. Same with timestamp and generation info etc.

operators/endpointmetrics/controllers/observabilityendpoint/testdata/crd/clusterversions-crd.yaml
douglascamata, line 1: How will we maintain these CRDs and other manifests `testdata` whenever they change due to upgrades in cluster components updating them?

douglascamata (commented):
Mostly good, but I left a few questions and thoughts.

operators/endpointmetrics/controllers/observabilityendpoint/testdata/crd/hostedclusters.hypershift.yaml
philipgough: I see now its in the testdata directory which is fine and makes sense. Best to exclude the specifics that will be generated in any case

operators/endpointmetrics/pkg/hypershift/hypershift.go
douglascamata: I mean, the `if` will do its job, but the log message will be wrong.

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
thibaultmg: Yes good point. I was considering the reconcile function as the top level one. But errors are indeed logged by the controller runtime. Removing those.

operators/endpointmetrics/controllers/observabilityendpoint/testdata/crd/hostedclusters.hypershift.yaml
thibaultmg: yes I added them for the integration tests with envtest. I have encapsulated the CRDs dependencies here for the addon. Some are already declared at some other places in the codebase. There is some duplication, but it is easier to manage in my view. There are probably better solutions for this.

operators/endpointmetrics/controllers/observabilityendpoint/testdata/crd/clusterversions-crd.yaml
thibaultmg: Yes I agree this is not ideal. We should probably have a central directory storing all CRD dependencies (for integration and e2e tests). For this to work, each test must select only the needed CRDs and not import the full package. On my side I had issue importing CRDs one by one in envtest by reading the file and unmarshalling into the required format.
The same needs to be done on the e2e tests side. I would like to be able to clearly identify who depends on what. I could create a task for centralizing them. WDYT?
But I am not sure I adressed your comment?

operators/endpointmetrics/pkg/hypershift/hypershift.go
thibaultmg: indeed, changing that

operators/endpointmetrics/controllers/observabilityendpoint/testdata/crd/hostedclusters.hypershift.yaml
thibaultmg: cf https://github.com/stolostron/multicluster-observability-operator/pull/1352#discussion_r1539232751

operators/endpointmetrics/pkg/openshift/openshift.go
thibaultmg: As I was refactoring this, I was hopping to remove them completely. But it actually requires changing some functions signatures to reflect back some status info. Doing a second pass

philipgough (approved):
/lgtm

operators/endpointmetrics/controllers/observabilityendpoint/testdata/crd/clusterversions-crd.yaml
douglascamata: Good good, I think we can follow up with this later. My point is not about a central directory though, but more about control and awareness of the versions. I want to avoid having issues when in production the CRDs are different from our tests and because of this our tests pass but prod breaks.

douglascamata (approved):
/lgtm

PR #1359, opened 2024-02-29
Kind version

philipgough:
Updates the KinD version and refactors some of the e2e test scripts to removing duplication.

We now fetch all test dependencies early and copy them into the ec2-users PATH.

Prior, we were duplicating the installs as the ec2-user post ssh so there was a potential to fail later in the test if those downloads failed. This change avoids that and allows us to fail the e2e tests early if the scp fails for any reason.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,4 @@ results.xml
 
 # build/tests artifacts
 registration-operator
+tmp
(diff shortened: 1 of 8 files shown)


saswatamcode (approved):
/lgtm

PR #1360, opened 2024-02-29
Cleanup some unused code and docs tidy

philipgough:


saswatamcode (approved):
LGTM, one small nit could put all the output stuff in docs into code blocks

PR #1362, opened 2024-03-01
Bring the Route and Route RBAC up to date

philipgough:
The logs complain about the versions used here as they are deprecated. Brings them up to date with the newer KinD version.


moadz (approved):
lgtm! These are the router/route stubs for OCP on kind right?

PR #1365, opened 2024-03-04
e2e testing changes

philipgough:
The focus on this change which is a follow up to #1353 , #1359, #1360, #1361, #1362 is to allow running the KinD end-to-end tests in all environments and provide a sensible refactor to the underlying bash scripts.

It attempts to split the logic into the following:

1. Cluster bootstrap
2. ACM setup
3. Test run

as opposed to the previous target which setup and kicked off everything but was difficult to break down and assess during test runs.

It also adds an additional Dockerfile which does not require access to the Red Hat CI registry in order to develop on.

The following is out of scope here but will be added in follow up(s):

1. Docs explaining how to build and use local image in e2e tests
2. Targets, scripts and docs for developing the metrics collector
3. Clean up after test artifacts in automated fashion

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -9,11 +9,11 @@ This document describes the process for build, running and testing this applicat
 This project is built using [operator-sdk](https://github.com/operator-framework/operator-sdk).
 There is a hard dependency on version `1.4.2` at the time of writing.
 
-> [!NOTE]  
+> [!NOTE]
 > Due to this [controller-gen issue](https://github.com/kubernetes-sigs/controller-tools/issues/880)
 > the steps below cannot be built using Go 1.22.x.
 
-> [!NOTE]  
+> [!NOTE]
 > For macOS users on arm64 you may need to build the binary from source using the steps below.
 
 Firstly, remove any existing `operator-sdk` binary from your $PATH.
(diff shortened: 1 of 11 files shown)


DEVELOPMENT.md
```diff
@@ -59,3 +59,47 @@ To build the image, which requires access to Red Hat CI registry, run:
...
+```
+
+Likewise, we can also bypass the environment setup by running:
+
+```shell
+SKIP_INSTALL_STEP=true make e2e-tests-in-kind
+```
+
+After running the tests, there will be some changes made to your environment.
+We can clean up the environment by running:
```
philipgough, line 98: This is not ideal. It will be better to copy all the required manifests into a temp directory that we can exclude and act on the copies. 

Unfortunately these paths are deeply embedded within the tests so I will create a follow-up in order to not bloat this further than it already is!

operators/multiclusterobservability/Dockerfile.dev
```diff
@@ -0,0 +1,61 @@
...
+COPY ./operators/multiclusterobservability/manifests /usr/local/manifests
+
+# install the prestop script
+COPY ./operators/multiclusterobservability/prestop.sh /usr/local/bin/prestop.sh
+
+# install operator binary
+COPY --from=builder /workspace/bin/manager ${OPERATOR}
+USER ${USER_UID}
+
+ENTRYPOINT ["/usr/local/bin/mco-operator"]
```
philipgough, line 61: see the commit message for justification on adding this!

scripts/test-utils.sh
```diff
@@ -7,12 +7,14 @@
 get_latest_snapshot() {
   BRANCH=""
   LATEST_SNAPSHOT=""
+  SNAPSHOT_RELEASE=${SNAPSHOT_RELEASE:=2.10}
```
philipgough, line 10: we can bump this per release, but essentially this allows you to add a particular snapshot to pick up, 2.9 , 2.10 etc

scripts/test-utils.sh
```diff
@@ -7,12 +7,14 @@
...
+  MATCH=$SNAPSHOT_RELEASE".*-SNAPSHOT"
   if [[ ${PULL_BASE_REF} == "release-"* ]]; then
     BRANCH=${PULL_BASE_REF#"release-"}
     LATEST_SNAPSHOT=$(curl https://quay.io//api/v1/repository/open-cluster-management/multicluster-observability-operator | jq '.tags|with_entries(select(.key|test("'${BRANCH}'.*-SNAPSHOT-*")))|keys[length-1]')
   fi
   if [[ ${LATEST_SNAPSHOT} == "null" ]] || [[ ${LATEST_SNAPSHOT} == "" ]]; then
-    LATEST_SNAPSHOT=$(curl https://quay.io/api/v1/repository/stolostron/multicluster-observability-operator | jq '.tags|with_entries(select((.key|contains("SNAPSHOT"))and(.key|contains("9.9.0")|not)))|keys[length-1]')
```
philipgough, line 15: this was not guaranteed to pick up the latest snapshot from local testing!

tests/pkg/tests/observability_install_test.go
```diff
@@ -242,9 +242,15 @@ func installMCO() {
...
+	Eventually(func() error {
+		BearerToken, err = utils.FetchBearerToken(testOptions)
+		if err != nil {
+			klog.Errorf("fetch bearer token error: %v", err)
+		}
+		if BearerToken == "" {
+			return fmt.Errorf("failed to get bearer token: %w", err)
+		}
+		return nil
+
```
philipgough, line 254: it takes some time to populate the token within the Secret hence we don't want to fail here in later versions of kube!

coleenquadros:
lgtm

PR #1368, opened 2024-03-05
Create a Secret for Grafana Accessor SA if not auto-generated

philipgough:
This change creates a Secret for the Grafana Accessor ServiceAccount if there was not previously one generated by the platform.

As per [OCP documentation](https://docs.openshift.com/container-platform/4.11/release_notes/ocp-4-11-release-notes.html#ocp-4-11-legacy-service-account) - we should not rely on Secrets generated by the underlying control plane since these may go away.

This behaviour is protected with LegacyServiceAccountTokenNoAutoGeneration feature gate and hence it makes sense to ensure that we force the generation of said Secret and token which we depend on.


coleenquadros (approved):

coleenquadros:
lgtm

thibaultmg (approved):

thibaultmg (approved):
/lgtm

PR #1374, opened 2024-03-13
Fix CI lint

thibaultmg:


.github/workflows/lint.yaml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-    tags:
```
douglascamata, line 7: Won't this mean a change in behavior?

thibaultmg: It doesn't seem to be used as we use release branches 🤷 
Restored it.

philipgough (approved):
/lgtm

PR #1376, opened 2024-03-13
Fix comparison of AddOnDeploymentConfig

douglascamata:
We should not only compare the proxy config, but the whole spec, and trigger a reconciliation if anything has changed.

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
+++ b/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
@@ -599,7 +599,7 @@ func (r *PlacementRuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	clusterPred := getClusterPreds()
 
 	// Watch changes for AddonDeploymentConfig
-	AddonDeploymentPred := GetAddOnDeploymentPredicates()
+	addOnDeploymentConfigPred := GetAddOnDeploymentConfigPredicates()
 
 	// Watch changes to endpoint-operator deployment
 	hubEndpointOperatorPred := getHubEndpointOperatorPredicates()
@@ -900,7 +900,7 @@ func (r *PlacementRuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		// secondary watch for alertmanager accessor serviceaccount
 		Watches(&source.Kind{Type: &corev1.ServiceAccount{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(amAccessorSAPred))
 
-	// watch for AddonDeploymentConfig
+	// watch for AddOnDeploymentConfig
 	if _, err := r.RESTMapper.RESTMapping(schema.GroupKind{Group: addonv1alpha1.GroupVersion.Group, Kind: "AddOnDeploymentConfig"}, addonv1alpha1.GroupVersion.Version); err == nil {
 		ctrBuilder = ctrBuilder.Watches(
 			&source.Kind{Type: &addonv1alpha1.AddOnDeploymentConfig{}},
@@ -911,7 +911,7 @@ func (r *PlacementRuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					}},
 				}
 			}),
-			builder.WithPredicates(AddonDeploymentPred),
+			builder.WithPredicates(addOnDeploymentConfigPred),
 		)
 	}
 	manifestWorkGroupKind := schema.GroupKind{Group: workv1.GroupVersion.Group, Kind: "ManifestWork"}
--- a/operators/multiclusterobservability/controllers/placementrule/predicate_func.go
+++ b/operators/multiclusterobservability/controllers/placementrule/predicate_func.go
@@ -100,18 +100,19 @@ func getClusterPreds() predicate.Funcs {
 	}
 }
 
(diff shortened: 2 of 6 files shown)


douglascamata:
I have proof that the bulk of my changes have 100% test coverage from running `go test -cov` within Goland (see attached picture). SonarCloud is messed up, as always.

image

douglascamata:
Putting a hold to wait for the beginning of the z-stream to merge this.

operators/multiclusterobservability/controllers/placementrule/predicate_func.go
```diff
@@ -100,18 +100,19 @@ func getClusterPreds() predicate.Funcs {
...
 			return true
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			if !reflect.DeepEqual(e.ObjectNew.(*addonv1alpha1.AddOnDeploymentConfig).Spec.ProxyConfig,
-				e.ObjectOld.(*addonv1alpha1.AddOnDeploymentConfig).Spec.ProxyConfig) {
-				log.Info("AddonDeploymentConfig is updated", e.ObjectNew.GetName(), "name", e.ObjectNew.GetNamespace(), "namespace")
-				return true
+			newObj := e.ObjectNew.(*addonv1alpha1.AddOnDeploymentConfig)
+			oldObj := e.ObjectOld.(*addonv1alpha1.AddOnDeploymentConfig)
+			if reflect.DeepEqual(newObj, oldObj) {
```
philipgough, line 111: I wonder would it be better here to just do a DeepEqual check on the entire Spec as opposed to the entire resource.  
I am considering the possibility here where you get into an endless loop due to checking things like updatedAt timestamps and revision numbers when all we really care about is changes to the Spec itself?

coleenquadros:
I think the issue is not because AddonDeploymentConfig is not checked although we still need this change. From the issue it is stated that when the  AddonDeploymentConfig that has the node selector/tolerations in ClusterManagementAddon is removed the Node Selector still persists in the in endpoint operator deployment. This I believe is due to this code path [here](https://github.com/stolostron/multicluster-observability-operator/blob/0b6ae5ccd0ead56791937fe768a7b5786a0cd0f6/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go#L359) where `defaultAddonDeploymentConfig` is a global variable and when the addondeploymentconfig was first set it holds the old AddonDeploymentConfig and continues to keep setting it [here](https://github.com/stolostron/multicluster-observability-operator/blob/0b6ae5ccd0ead56791937fe768a7b5786a0cd0f6/operators/multiclusterobservability/controllers/placementrule/manifestwork.go#L314) because we did not clear it when the `addonDeploymentConfig` was removed from `ClusterManagementAddon`.

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -58,7 +59,7 @@ const (
 
 var (
 	log = logf.Log.WithName("controller_placementrule")
-	//watchNamespace                  = config.GetDefaultNamespace()
+	// watchNamespace                  = config.GetDefaultNamespace()
```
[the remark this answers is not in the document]
douglascamata: This is auto-formatting by the IDE, I didn't even touch it.

douglascamata:
@coleenquadros ohhh, I see. We have more than one issue then. 😄 

I'll fix both here.

operators/multiclusterobservability/controllers/placementrule/predicate_func.go
douglascamata: Good one, @philipgough. Fixed in https://github.com/stolostron/multicluster-observability-operator/pull/1376/commits/8429c0d9a21c654e90ad75dacab57a163a31b1fb

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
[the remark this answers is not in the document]
philipgough: yep, but we might as well clean it up since we are creating a diff anyway :)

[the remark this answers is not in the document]
douglascamata: Removed.

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -354,6 +354,11 @@ func createAllRelatedRes(
 					addonConfig,
 				)
 				if err != nil {
+					if k8serrors.IsNotFound(err) {
```
coleenquadros: AddOnDeploymentConfig resource will still be found right from Get? since it was not deleted? I think the check should be for if config.DefaultConfig is nil then empty out defaultAddonDeploymentConfig

coleenquadros: Just saw your one line change. I guess this will not apply :)

douglascamata: Yep. It's much more simple to standard with the default case value (use an empty config). Only if we find a config we update the reference. This way we don't need to handle all the possible scenario in which the config should be set to empty.

coleenquadros (approved):

PR #1379, opened 2024-03-14
[ACM-10342]update metrics collector client cert when client CA is deleted

coleenquadros:
https://issues.redhat.com/browse/ACM-10342

ClientCACertificateCN is used to sign in the Certificate Signing Request to generate the Client cert (`observability-controller-open-cluster-management.io-observability-signer-client-cert`)  so when ClientCACertificateCN is deleted or updated we need to also update the client cert.


saswatamcode (approved):
/lgtm

PR #1386, opened 2024-03-21
[ACM-10511] Update hub collector with the whole obs addon spec

douglascamata:
There's no other way to fetch the observability addon spec from the metrics collector in the Hub besides querying the kube API for it. 

This fixes the synchronization of the MCO's observability addon spec to the hub's collector.

Important note: with these changes, the Hub's metrics collector will abide to the `metricsEnabled` setting of the `ObservabilityAddon` spec.


philipgough (approved):
/lgtm

PR #1389, opened 2024-03-28
[2.10] [ACM-10406] Fix handling of updates on AddOnDeploymentConfig 

douglascamata:
This cherry picks #1376 into the 2.10 release stream.

Fix in the addon deployment config predicate: we should not only compare the proxy config, but the whole spec, and trigger a reconciliation if anything has changed there.

Fix in the placement rule controller: whenever starting the loop for fetching a new default addon deployment config, reassign the result var to an empty config. We only change it from empty if we find any.


thibaultmg (approved):
/lgtm

philipgough (approved):
/lgtm

PR #1395, opened 2024-04-05
ACM-10828: sort resources for microshift install

thibaultmg:
Enables the endpoint operator to deploy resources on a microshift spoke.
Changes:

- Add necessary permissions for setting finalizers on deployed resources (needs finalizer updates perm on oba which is the referenced owner)
- Sort deployed resources so that needed permissions are deployed first, then needed CRDs, etc.
- Add some other missing perm for prometheus
- Remove noisy logs from the IBM filteredcache by only watching MCO resource when on the hub
- Clean some badly formatted logs in deploy

### DIFF
--- a/operators/pkg/deploying/deployer.go
+++ b/operators/pkg/deploying/deployer.go
@@ -66,7 +66,7 @@ func (d *Deployer) Deploy(obj *unstructured.Unstructured) error {
 	)
 	if err != nil {
 		if errors.IsNotFound(err) {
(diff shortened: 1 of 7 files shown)


operators/pkg/deploying/deployer.go
douglascamata, line 1: Why not all the update logs in this file were replaced with the new function, but only a few of them?

thibaultmg: Good catch. I was tracking the ones including ":" in the key/values and missed the missing ones. Updating.

PR #1400, opened 2024-04-09
E2e tests

coleenquadros:

The changes to the test cases under this PR is to accommodate the  global hub scenario post ACM-8509 changes.  The tests currently are only meant for `local-cluster` as currently we do not create any additional managed clusters to run the tests against. Adding a normal spoke would be the next steps in this effort

Background 
- Hub/local-cluster's endpoint operator, metrics collector and related resources are in `open-cluster-management-observability` namespace
- normal spokes' endpoint operator, metrics collector  and related resources are in `open-cluster-management-addon-observability` namespace
- local-cluster metrics collection do not have manifestwork and Observability addons so these tests are skipped
- resources limits have been increased for certain components as they OOM kill with the previous limits set


operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
```diff
@@ -240,9 +240,13 @@ func (r *ObservabilityAddonReconciler) Reconcile(ctx context.Context, req ctrl.R
 		}
 		deployer := deploying.NewDeployer(r.Client)
 		for _, res := range toDeploy {
-			if err := controllerutil.SetControllerReference(obsAddon, res, r.Scheme); err != nil {
-				log.Info("Failed to set controller reference", "resource", res.GetName())
-				globalRes = append(globalRes, res)
+			if !isHubMetricsCollector {
+				// For kind tests we need to deploy prometheus in hub but do not need to set controller
+				// reference as there is no observabilityaddon
```
douglascamata, line 245: Do we **need not** or **can not**? Because if the case is that we don't need, but it's fine if we do it, we should just do it to simplify the code and avoid the `if`.

If we **can not**, then it would be worth to mention why in the comment fot future reference.

coleenquadros:
@philipgough these wont work for Kind, I dont have the namespace change here for the cluster role binding.  this is the PR with the changes related to kind https://github.com/stolostron/multicluster-observability-operator/pull/1404

operators/endpointmetrics/pkg/util/client.go
```diff
@@ -8,7 +8,9 @@ import (
 	"context"
 	"os"
 
+	oav1beta2 "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/api/v1beta2"
```
douglascamata, line 11: Why did this import move?

tests/pkg/utils/mco_deploy.go
```diff
@@ -349,6 +353,11 @@ func CheckAllPodsAffinity(opt TestOptions) error {
 	}
 
 	for _, pod := range podList {
+		if pod.Labels["name"] == "endpoint-observability-operator" || pod.Labels["component"] == "metrics-collector" ||
+			pod.Labels["component"] == "uwl-metrics-collector" {
+			// No affinity set for endpoint-operator and metrics-collector in the hub
+			continue
+		}
```
douglascamata, line 360: Why these changes? Why do we skip looking at affinities of these pods? They still exist and can have affinities, from what I know.

douglascamata (changes requested):

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
coleenquadros: There is no obsaddon for hub so we cannot. ill update it in the comments

tests/pkg/utils/mco_deploy.go
douglascamata: We should eventually have test coverage for this.

douglascamata (approved):
LGTM

PR #1402, opened 2024-04-10
[ACM-10987]Check SA imagePullSecret length to discard comparing against generate…

philipgough:
…d and injected values

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/protobuf v1.5.3
 	github.com/golang/snappy v0.0.4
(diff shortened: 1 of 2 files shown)


saswatamcode (approved):
/lgtm

coleenquadros:
Can you also add a watch logic for the endpoint-sa [ here](https://github.com/stolostron/multicluster-observability-operator/blob/1cb5bad8ffedfe038cc5dbebee399f915241ccf3/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go#L1033) for the case when its deleted.

philipgough:
Discussed about follow up improvements offline with @coleenquadros which we will push to a new PR

PR #1403, opened 2024-04-10
[ACM-10967] Stop regenerating Secrets from SA updates

philipgough:
…d and injected values

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
+++ b/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
@@ -15,6 +15,8 @@ import (
 	"strconv"
 	"strings"
 
+	"golang.org/x/exp/slices"
+
 	rbacv1 "k8s.io/api/rbac/v1"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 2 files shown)


saswatamcode (approved):
/lgtm

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -568,9 +572,32 @@ func createUpdateResourcesForHubMetricsCollection(c client.Client, manifests []w
 					needsUpdate = true
 				}
 			case *corev1.ServiceAccount:
-				currentServiceAccount := currentObj.(*corev1.ServiceAccount)
-				if !reflect.DeepEqual(obj.ImagePullSecrets, currentServiceAccount.ImagePullSecrets) {
-					needsUpdate = true
+				// https://issues.redhat.com/browse/ACM-10967
+				// Some of these ServiceAccounts will be read from static files so they will never contain
+				// the generated Secrets as part of their corev1.ServiceAccount.ImagePullSecrets field.
+				// This checks by way of slice length if this particular ServiceAccount can be one of those.
```
coleenquadros, line 578: SHould we special case it for endpoint-sa? just to make sure no other SA runs through this logic, even if there isnt any other one

philipgough: I don't think so since we would see this problematic behaviour with other SAs too

coleenquadros:
@philipgough Can you also add a watch logic for the endpoint-sa [here](https://github.com/stolostron/multicluster-observability-operator/blob/1cb5bad8ffedfe038cc5dbebee399f915241ccf3/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go#L1033) for the case when its deleted.

Also I had a look at the code flow once again. Manifestwork will only hold the SA without the dockcfg-xxx generated secret info and get created and updated accordingly. ([here](https://github.com/stolostron/multicluster-observability-operator/blob/1cb5bad8ffedfe038cc5dbebee399f915241ccf3/operators/multiclusterobservability/controllers/placementrule/manifestwork.go#L151) .

There is no watch logic to check for SA update in the endpoint controller or placement controller. So maybe we shouldn't check for the added secret logic and only check if the the obj from the template has been changed without the dockcfg secret info.

philipgough:
Hey @coleenquadros - since this is a backport/cherry-pick I think we should merge it as it exists on main. However it seems like you have identified some improvements to the flow so we can also land those on main and backport to 2.10 post testing. wdyt?

PR #1405, opened 2024-04-12
[RHOBS-987] Update Thanos Compact retention

douglascamata:
Doing the retention increase in this PR.

The changes in max compaction level and and concurrency will be done through the jsonnet source, as they aren't supported in the CRDs.


philipgough (approved):
/lgtm

PR #1406, opened 2024-04-16
e2e kind tests

coleenquadros:

For kind tests the following changes were made

- Since kind needs prometheus installed, an annotation is added to MCO to enable the `installPrometheus` flag. This allows the code path to take the installPrometheus flags route to install the necessary resources while running on kind. 
- We need to also ensure that the prometheus resources are installed in the correct namespace i.e. the `open-cluster-management-observability`. Even the clusterrolebindings should refer to the correct namespace. The scrape-targets has been modified in place to ensure the jobs in the correct namespace are picked.

Annotations logic was merged with e2e tests PR - https://github.com/stolostron/multicluster-observability-operator/pull/1400/files

### DIFF
--- a/cicd-scripts/customize-mco.sh
+++ b/cicd-scripts/customize-mco.sh
@@ -14,7 +14,7 @@ SED_COMMAND=${SED}' -i-e -e'
 
 # Set the latest snapshot if it is not set
 source ./scripts/test-utils.sh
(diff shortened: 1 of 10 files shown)


coleenquadros:
Screenshot 2024-04-18 at 10 41 09

PR #1411, opened 2024-04-23
E2e kind tests

coleenquadros:

For kind tests the following changes were made

- Since kind needs prometheus installed, an annotation is added to MCO to enable the installPrometheus flag. This allows the code path to take the installPrometheus flags route to install the necessary resources while running on kind.
- We need to also ensure that the prometheus resources are installed in the correct namespace i.e. the open-cluster-management-observability. Even the clusterrolebindings should refer to the correct namespace. The scrape-targets has been modified in place to ensure the jobs in the correct namespace are picked.
- Annotations logic was merged with e2e tests PR - https://github.com/stolostron/multicluster-observability-operator/pull/1400/files


operators/endpointmetrics/pkg/rendering/renderer.go
```diff
@@ -56,6 +56,13 @@ func Render(
 	hubInfo *operatorconfig.HubInfo,
 ) ([]*unstructured.Unstructured, error) {
 
+	isKindTest := false
+	if strings.Contains(hubInfo.ClusterName, "kind") {
+		//remove -kind from the cluster name
```
jacobbaungard, line 61: This is because the only way to detect whether we run in kind is via the cluster name? So i.e add the suffix in the `observability_addon` controller and remove it again here?

This kind of name mangling doesn't seem super nice to me.

It seems the `isHubMetricsCollector` and `installPrometheus` are determined via the environment, could we infer this information directly form the environment instead here, or alternatively pass a `isKindTest` as an argument to Render() ?

coleenquadros: Makes sense. I've added checking for the environment var in render instead.

jacobbaungard: Thanks, that looks much cleaner to me :+1:

jacobbaungard:
Looks good, I just wonder if you can explain a little further why this is specifically needed in the kind tests, vs a normal deployment?

> We need to also ensure that the prometheus resources are installed in the correct namespace i.e. the open-cluster-management-observability. Even the clusterrolebindings should refer to the correct namespace. The scrape-targets has been modified in place to ensure the jobs in the correct namespace are picked.

coleenquadros:
> Looks good, I just wonder if you can explain a little further why this is specifically needed in the kind tests, vs a normal deployment?
> 
> > We need to also ensure that the prometheus resources are installed in the correct namespace i.e. the open-cluster-management-observability. Even the clusterrolebindings should refer to the correct namespace. The scrape-targets has been modified in place to ensure the jobs in the correct namespace are picked.

In the normal deployments in the Hub which is always an OCP cluster, we have prometheus installed and the cluster role bindings subjects and scrape jobs point to the` open-cluster-management-addon-observability` namespace in the templates which is correct in the case of a normal non-OCP  spoke  when it needs to install prometheus because that namespace exists. In the case of a kind cluster behaving as a hub, we need to explicitly install prometheus and the role bindings subjects and scrape targets should point to open-cluster-management-observability namespace. Under normal deployments we will never install prometheus in the hub because its always an OCP cluster and will have prometheus. 

This is how its always been tested prior to global hub changes and now we are just ensuring the correct namespace is referred to in kind.

jacobbaungard (approved):
Thanks for the explanation @coleenquadros , makes sense!

/lgtm

philipgough (approved):

PR #1412, opened 2024-04-25
Fix openshift-virtualization dashboard

sradco:
Fix openshift-virtualization dashboard name,
updated capital letter to lower case and
add the dashboard to kustomization.yaml.

Signed-off-by: Shirly Radco <sradco@redhat.com>


ch-stark:
lgtm

PR #1415, opened 2024-04-26
Placementrule: reduce resource usage

jacobbaungard:
When we reconcile in the placementrule controller, one of things we do is to cleanup after managed clusters which has been deleted. This is done by comparing the list of clusters in `managedClusterList`, to the list in `obsAddonList.Items`. If there are additional clusters in the `obsAddonList` it means a spoke was removed, and we need to do some cleanup for that spoke.

Prior to this commit, this work was done in an highly inefficient way. To get the set difference, we would loop over the `managedClusterList`, and each time we would loop through the `obsAddonList`, search for the cluster from the `managedClusterList` loop iteration, and allocate a new `currentClusters` slice with all the remaining clusters. That means in each loop over the `managedClusterList` we did `obsAddonList-i` allocations and a loop of the same amount.

In this commit we change this, to only do a single loop over the `obsAddonList`, doing a constant time lookup to see if the item exist in the `managedClusterList` and only add items which don't to a slice. This means only do one allocation per deleted spoke. Further we save one allocation of size `obsAddonList` as there is no reason to save this to a separate slice.

This is many many orders of magnitudes more efficient, and will likely save significant resource usage on larger installations.


philipgough (approved):
/lgtm

@jacobbaungard thanks for this and the detailed explanation and benchmark work

jacobbaungard:
Rebased with E2E test fix.

saswatamcode (commented):
/lgtm

PR #1416, opened 2024-04-26
[ACM-10706] Add support for custom observability hub URL

douglascamata:
This is supposed to cover cases where there is some sort of intermediate component, like but not limited to a reverse proxy, between the spokes and the Observatorium API Route in the hub.

I decided to add a new field to the MCO CR's advanced config instead of using AddOnDeploymentConfig's custom variables because that's how we customize other things in the MCO. This way the configuration stays all in one place, avoiding any possible user confusion or extra cognitive load.

### DIFF
--- a/operators/multiclusterobservability/pkg/config/config.go
+++ b/operators/multiclusterobservability/pkg/config/config.go
@@ -481,7 +481,6 @@ func GetDefaultTenantName() string {
 }
 
 // GetObsAPIHost is used to get the URL for observartium api gateway.
-// TODO: grab custom Obs API host from somewhere?
 func GetObsAPIHost(client client.Client, namespace string) (string, error) {
 	mco := &observabilityv1beta2.MultiClusterObservability{}
 	err := client.Get(context.TODO(),
--- a/operators/multiclusterobservability/pkg/config/config.go
+++ b/operators/multiclusterobservability/pkg/config/config.go
@@ -492,6 +492,10 @@ func GetObsAPIHost(client client.Client, namespace string) (string, error) {
 	}
 	advancedConfig := mco.Spec.AdvancedConfig
 	if advancedConfig != nil && advancedConfig.CustomObservabilityHubURL != "" {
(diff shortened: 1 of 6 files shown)


operators/multiclusterobservability/pkg/config/config.go
```diff
@@ -482,6 +482,18 @@ func GetDefaultTenantName() string {
 
 // GetObsAPIHost is used to get the URL for observartium api gateway.
 func GetObsAPIHost(client client.Client, namespace string) (string, error) {
+	mco := &observabilityv1beta2.MultiClusterObservability{}
```
philipgough, line 485: nit, but I wonder if the `GetObsAPIHost` shouldn't be adapted to accept a context here itself?

operators/multiclusterobservability/pkg/config/config.go
```diff
@@ -482,6 +482,18 @@ func GetDefaultTenantName() string {
...
+	err := client.Get(context.TODO(),
+		types.NamespacedName{
+			Name: GetMonitoringCRName(),
+		}, mco)
+	if err != nil && !errors.IsNotFound(err) {
+		return "", err
+	}
+	advancedConfig := mco.Spec.AdvancedConfig
+	if advancedConfig != nil && advancedConfig.CustomObservabilityHubURL != "" {
+		return string(advancedConfig.CustomObservabilityHubURL), nil
```
philipgough, line 495: I know there is a kubebuilder annotation for the regex but would it be worth calling https://pkg.go.dev/net/url#ParseRequestURI here and returning an error if we hit one?

philipgough (approved):
/lgtm

operators/multiclusterobservability/pkg/config/config.go
douglascamata: I'll add this in a follow up PR where I will do the same as here but for the alertmanager URL, then both functions will get the context. 👍

philipgough (approved):
/lgtm

PR #1425, opened 2024-05-02
Add required permissions for simulated cluster & various fixes for metrics generation

jacobbaungard:
This PR contains a few various fixes that was needed to run the metric collector simulator and generate new metrics for it. See details in individual commits messages.

### DIFF
--- a/tools/simulator/metrics-collector/generate-metrics-data.sh
+++ b/tools/simulator/metrics-collector/generate-metrics-data.sh
@@ -11,6 +11,8 @@ WORKDIR="$(
   pwd -P
 )"
 
+source ${WORKDIR}/../../../scripts/install-binaries.sh
+
 # Create bin directory and add it to PATH
 mkdir -p ${WORKDIR}/bin
 export PATH=${PATH}:${WORKDIR}/bin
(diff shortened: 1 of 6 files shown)


tools/simulator/metrics-collector/generate-metrics-data.sh
```diff
@@ -40,47 +41,63 @@ if ! command -v kubectl &>/dev/null; then
     if [[ "$(uname)" == "Linux" ]]; then
       curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.21.0/bin/linux/amd64/kubectl
     elif [[ "$(uname)" == "Darwin" ]]; then
-      curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.21.0/bin/darwin/amd64/kubectl
```
philipgough, line 43: I previously created and consolidated the tooling into https://github.com/stolostron/multicluster-observability-operator/blob/main/scripts/install-binaries.sh

Would it help to extend that script with the requirements and call as needed?

jacobbaungard: Yes, that seems sensible, will take a look :+1:

jacobbaungard: Ok done!

jacobbaungard:
/hold

The `$*` in the install-scrips causes some issues (probably just a warning) when calling these script with cmd line args. Trying to fix...

philipgough (approved):
/lgtm

PR #1427, opened 2024-05-03
[ACM-10812]: retry status update on conflict

thibaultmg:
In some cases, while the metrics collector is well deployed in the spoke, the addon status remains in progressing state on the hub. After having modified the part that reflects the local observability addon state to the hub in https://github.com/stolostron/multicluster-observability-operator/pull/1420, this PR modifies the code that updates the local observability addon on the spoke:

- Adds retry logic in case of conflicts
- Avoids overriding the whole the status.Conditions list with the latest one
- Limits the number of status.Conditions to 10.

It also removes some weird stuff like the `reportStatus bool` function parameter in the reportStatus function 😅

### DIFF
--- a/operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
+++ b/operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
@@ -202,7 +202,9 @@ func (r *ObservabilityAddonReconciler) Reconcile(ctx context.Context, req ctrl.R
 				// ACM 8509: Special case for hub/local cluster metrics collection
 				// We do not report status for hub endpoint operator
 				if !isHubMetricsCollector {
-					util.ReportStatus(ctx, r.Client, util.NotSupportedStatus, obsAddon.Name, obsAddon.Namespace)
+					if err := util.ReportStatus(ctx, r.Client, util.NotSupportedStatus, obsAddon.Name, obsAddon.Namespace); err != nil {
+						log.Error(err, "Failed to report status")
+					}
 				}
 
 				return ctrl.Result{}, nil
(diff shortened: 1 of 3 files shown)


operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
```diff
@@ -201,7 +201,12 @@ func (r *ObservabilityAddonReconciler) Reconcile(ctx context.Context, req ctrl.R
 				log.Error(err, "OCP prometheus service does not exist")
 				// ACM 8509: Special case for hub/local cluster metrics collection
 				// We do not report status for hub endpoint operator
-				util.ReportStatus(ctx, r.Client, obsAddon, "NotSupported", !isHubMetricsCollector)
+				if !isHubMetricsCollector {
```
jacobbaungard, line 204: We don't have an ObservabilityAddon in the local cluster?

operators/endpointmetrics/pkg/util/status.go
```diff
@@ -10,45 +10,92 @@ import (
...
-	err := client.Status().Update(ctx, i)
-	if err != nil {
-		log.Error(err, "Failed to update status for observabilityaddon")
+
+	return nil
+}
+
+// shouldAppendCondition checks if the new condition should be appended to the status conditions
+// based on the last condition in the slice.
+func shouldAppendCondition(conditions []oav1beta1.StatusCondition, newCondition *oav1beta1.StatusCondition) bool {
```
jacobbaungard, line 90: So it stops from changing the condition, if the condition is the same?

operators/endpointmetrics/pkg/util/status.go
```diff
@@ -10,45 +10,92 @@ import (
...
+			return err
+		}
+
+		if !shouldAppendCondition(obsAddon.Status.Conditions, newCondition) {
+			return nil
+		}
+
+		obsAddon.Status.Conditions = append(obsAddon.Status.Conditions, *newCondition)
+
+		if len(obsAddon.Status.Conditions) > MaxStatusConditionsCount {
```
jacobbaungard, line 75: Is this just for convenience, or why do we want only 10?

thibaultmg: It's an arbitrary number to limit the list size... You think we need more than 10? Previously we were overriding the whole list with the last element, so 10 is better than before already.

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
thibaultmg: Good question. The local cluster is not working as an addon, it's probably because of that.
In any case, I haven't changed the existing logic.

operators/endpointmetrics/pkg/util/status.go
thibaultmg: The idea is to avoid adding duplicated status. If it is available, I don't need to report available again. 
But I assume here that statuses are sorted by date, which is not necessarily the case IIRC. Probably have to change this! Having a look.

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
jacobbaungard: Oh, missed it passed through the `isHubMetricsCollector` as args. Thanks

operators/endpointmetrics/pkg/util/status.go
jacobbaungard: I mostly just wondered if there was a specific reason for such a limit. Couldn't we just keep adding? Not sure if there is some best practice on this area.

Anything, this should be fine for now.

operators/endpointmetrics/pkg/util/status.go
thibaultmg: updated.

operators/endpointmetrics/pkg/util/status.go
thibaultmg: I want to avoid an ever increasing list... And potential associated bugs... Maybe it's over engineered

jacobbaungard (approved):
/lgtm

PR #1433, opened 2024-05-14
Tests: Don't check remote_write_requests on spokes 

jacobbaungard:
Previously the test "Should have acm_remote_write_requests_total metrics with correct labels/value" looked for the metric for both the hub cluster and any spokes. However, **per my understanding** the spokes isn't supposed to expose this metric, hence the test would always fail if any managed clusters were added to the test setup.


thibaultmg (approved):
/lgtm

PR #1435, opened 2024-05-14
Improve e2e logs

thibaultmg:


thibaultmg:
I am waiting to make e2e pass and check the logging results before merging. Yesterday e2e did not work...

PR #1436, opened 2024-05-15
Add integration tests to github workflows

thibaultmg:
Adds envtest based integration tests to github workflows.


.github/workflows/integration_tests.yaml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/checkout@v4
+
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.21.x
+          cache-dependency-path: "**/*.sum"
+
+      - name: Install deps
+        run: make deps
```
saswatamcode, line 28: Worth adding short go build step before? So we don't run tests if things don't properly build?

thibaultmg: yes, done!

PR #1437, opened 2024-05-15
Relocate kubeconfig to /workspace 

subbarao-meduri:
The canary builds run e2e tests in a containerized env, and the user 1001 does not have access to `/opt` path.

Change kubeconfig path to `/workspaces` instead so user can access it.


douglascamata (approved):
/lgtm

PR #1438, opened 2024-05-15
[release-2.10] Relocate kubeconfig for non-root user access

subbarao-meduri:
The canary builds run e2e tests in a containerized env, and the user 1001 does not have access to /opt path.

Change kubeconfig path to /workspaces instead so user can access it.

Complimentary [PR 300](https://github.com/stolostron/canary-scripts/pull/300) in canary-scripts repo that invokes the containerized image


subbarao-meduri:
@coleenquadros are the e2e tests fixed in `release-2.10` stream? or only in `main` branch only?

douglascamata (approved):
/lgtm

PR #1441, opened 2024-05-17
[ACM-11543] Do not add custom obs api url to certs SAN

douglascamata:
When using a custom observability hub URL, the only SSL termination mode supported is **passthrough**, which means the certificates must still be signed with the Route's host in the SAN field. 

We **never** want the custom URL in any SAN.

**Implementation detail**

I separated the original code into two functions: one that gets the URL for access and another one that gets the URL for the cert's SAN field.

### DIFF
--- a/operators/multiclusterobservability/api/shared/multiclusterobservability_shared.go
+++ b/operators/multiclusterobservability/api/shared/multiclusterobservability_shared.go
@@ -20,12 +20,15 @@ import (
 // +kubebuilder:validation:MaxLength=2083
 type URL string
 
+// Validate validates the underlying URL.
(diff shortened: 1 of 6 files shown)


operators/multiclusterobservability/api/shared/multiclusterobservability_shared.go
```diff
@@ -25,6 +25,14 @@ func (u URL) Validate() error {
 	return err
 }
 
+func (u URL) Host() (string, error) {
```
philipgough, line 28: nit but naming is confusing here since the function returns both the host and the parsed path!

PR #1442, opened 2024-05-17
[ACM-11722] Add support for alertmanager path prefix and validate URL

douglascamata:
Previously the URL's path was being ignored in the code that built the additional cluster monitoring configuration. 

I also made the code handle URLs without protocol as being `https://` by default. This has been the only supported protocol anyway as it was previously prepended to all the hosts when the protocol was omitted.

### DIFF
--- a/operators/multiclusterobservability/api/shared/multiclusterobservability_shared.go
+++ b/operators/multiclusterobservability/api/shared/multiclusterobservability_shared.go
@@ -15,8 +15,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// URL is kubebuilder type that validates the containing string is an URL.
-// +kubebuilder:validation:Pattern=`^https?:\/\/`
+// URL is kubebuilder type that validates the containing string is an HTTPS URL.
+// +kubebuilder:validation:Pattern=`^https:\/\/`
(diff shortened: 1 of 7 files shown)


operators/multiclusterobservability/api/shared/multiclusterobservability_shared.go
```diff
@@ -15,8 +15,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// URL is kubebuilder type that validates the containing string is an URL.
-// +kubebuilder:validation:Pattern=`^https?:\/\/`
+// URL is kubebuilder type that validates the containing string is an HTTPS URL.
+// +kubebuilder:validation:Pattern=`^https:\/\/`
```
philipgough, line 19: is there any risk here of this breaking anyone?

douglascamata: No, because the creation of the additional config for cluster monitoring has been always hardcoded for HTTPS here: https://github.com/stolostron/multicluster-observability-operator/blob/21c0fdac879c3993a6ff1c1a7708c6897a2e4926/operators/endpointmetrics/controllers/observabilityendpoint/ocp_monitoring_config.go#L258

philipgough (approved):
/lgtm

PR #1449, opened 2024-05-22
Update prometheus operator crd after version update

coleenquadros:
Update prometheus CRD after  update to stolostron/prometheus-operator version v0.73.2

https://github.com/stolostron/prometheus-operator/pull/40


saswatamcode (approved):
/lgtm

PR #1451, opened 2024-05-23
Update e2e with managed

coleenquadros:


philipgough:
@coleenquadros is this still needed?

PR #1452, opened 2024-05-23
Tests: Use CI images, and improve getting snapshot from Quay

jacobbaungard:
- Always use the CI images we built during tests, don't rely on trying to get a resonable snapshot from quay
- Improve our chances of getting a resonable recent snapshot on quay for images we don't build as part of MCO (like grafana).

More details in individual commits


philipgough (approved):
/lgtm

PR #1456, opened 2024-05-28
[ACM-11543] Check if URL already contains remote write path before appending it

douglascamata:
For the reverse proxy/LB support, the remote write endpoint could be under a custom path prefix (i.e. `host:port/custom-path//api/metrics/v1/default/api/v1/receive`). 

We can't hardcoded the path directly to `/api/metrics/v1/default/api/v1/receive`. So the best way forward if to append this path to the url as a string, unless it's already present, then parse it into an `url.URL`.

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/hub_info_secret.go
+++ b/operators/multiclusterobservability/controllers/placementrule/hub_info_secret.go
@@ -69,14 +69,18 @@ func generateHubInfoSecret(client client.Client, obsNamespace string,
 		}
 	}
 
-	if !strings.HasSuffix(obsAPIHost, operatorconfig.ObservatoriumAPIRemoteWritePath) {
-		obsAPIHost += operatorconfig.ObservatoriumAPIRemoteWritePath
-	}
-
 	obsApiURL, err := url.Parse(obsAPIHost)
 	if err != nil {
 		return nil, err
 	}
+
+	// We have to *append* the Remote Write path, and not hardcode it, because it could include
+	// a custom sub-path required by intermediate components running between spokes and the hub (i.e. reverse proxies
+	// or load balancers).
+	if !strings.HasSuffix(obsApiURL.Path, operatorconfig.ObservatoriumAPIRemoteWritePath) {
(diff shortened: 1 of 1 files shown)


operators/multiclusterobservability/controllers/placementrule/hub_info_secret.go
```diff
@@ -69,9 +69,13 @@ func generateHubInfoSecret(client client.Client, obsNamespace string,
 		}
 	}
 
-	obsApiURL := url.URL{
-		Host: obsAPIHost,
-		Path: operatorconfig.ObservatoriumAPIRemoteWritePath,
+	if !strings.HasSuffix(obsAPIHost, operatorconfig.ObservatoriumAPIRemoteWritePath) {
+		obsAPIHost += operatorconfig.ObservatoriumAPIRemoteWritePath
```
thibaultmg, line 73: We could use path.Join in case there is trailing slash to handle in obsAPIHost?

operators/multiclusterobservability/controllers/placementrule/hub_info_secret.go
```diff
@@ -69,9 +69,13 @@ func generateHubInfoSecret(client client.Client, obsNamespace string,
 		}
 	}
 
-	obsApiURL := url.URL{
-		Host: obsAPIHost,
-		Path: operatorconfig.ObservatoriumAPIRemoteWritePath,
+	if !strings.HasSuffix(obsAPIHost, operatorconfig.ObservatoriumAPIRemoteWritePath) {
```
thibaultmg, line 72: I would add your PR description here to explain this logic. The code is already crippled with this kind of things and you don't know the reason.

operators/multiclusterobservability/controllers/placementrule/hub_info_secret.go
douglascamata: I decided to go with `net/url.JoinPath` (which is also present in `url.URL.JoinPath`) instead. `path.Join` is made for filesystem paths and has filesystem specific behaviors that are not a good idea for URL path joining.

thibaultmg (approved):
/lgtm

thibaultmg (approved):
/lgtm

PR #1458, opened 2024-05-29
Fix e2e test metrics presence

thibaultmg:
The updated test is never triggered because the ginko filter needed for it is never set. So it is only triggered if there is no ginko focus. Which almost never happens. Unless you modify only some files that are not monitored for setting ginko focus, such as `cicd-scripts`... 😬 

When it runs, it checks that the expected list of metrics is being collected correctly from the spokes. I have updated the list to reflect what is currently being collected in our e2e environment. Additionally, I have added some logging to track discrepancies between what is actually collected and what we filter on.

With this test, we can monitor deviations regarding what metrics we collect in this stable environment.


jacobbaungard (approved):
Awesome, thanks!

/lgtm

philipgough (approved):

PR #1461, opened 2024-05-31
Tests: Only set ginkgo focus in kind tests

jacobbaungard:
Previously, we tried to cleverly decide which tests to run on PRs. However this is error prone, and has recently allowed code to be checked in which broke a test for almost 4 months. This due the said test not being included in any of the Ginkgo focus tags selected.

While this will use more resources, it's less error prone, and we ensure we cannot checkin code that breaks tests.


thibaultmg (approved):
/lgtm

PR #1472, opened 2024-06-07
ACM-12251: New addon status design document

thibaultmg:
The addon status update is currently causing issues such as:

- Flapping states
- Inconsistent states

This document aims to specify the expected behavior. It is a target for achieving a consistent management of the spoke addon status being reported to the hub.
It does not describe the current implementation. 

Main differences are:

- **Clarification of State Transitions**: The document clarifies state transitions with added restrictions. Specifically, transitions between degraded and available states are restricted to avoid flapping.
- **Progressing Status on Metrics-Collector Update**: On each metrics-collector update, the status goes back to progressing state.
- Addition of new condition types for each collector that are aggregated by the status controller.

### DIFF
--- a/operators/endpointmetrics/doc/design/addon-status.md
+++ b/operators/endpointmetrics/doc/design/addon-status.md
@@ -13,22 +13,15 @@ The status is reported through the standard `status.Conditions` structure in the
 - **reason**: A short, machine-readable string that gives the reason for the condition's last transition.
 - **message**: A human-readable message providing details about the condition.
 
-In our case, we use the following types of conditions:
+[Three standard types](https://open-cluster-management.io/concepts/addon#add-on-healthiness) of conditions are used to report the status of the addon:
 
 - **Available**: The addon is available and running, including the UWL metrics collector if configured.
 - **Degraded**: The addon is not working as expected. Either metrics are not being forwarded or the addon update has failed.
-- **Disabled**: The addon is disabled.
-- **NotSupported**: The addon is not supported on the current platform. This happens when prometheus is not being installed but the cluster is not OCP.
 - **Progressing**: The addon is being installed or has been updated.
 
-Finally, the types can be associated to the following reasons:
+Only one of these standard types can have the `True` status at a time. They are mutually exclusive. 
 
-- **DisabledMetrics**: The metrics collector is disabled. It leads to the `Disabled` status. 
-- **ForwardFailure**: The metrics collector failed to forward metrics. It leads to the `Degraded` status.
-- **ForwardSuccessful**: The metrics collector successfully forwarded metrics. It leads to the `Available` status.
-- **NotSupported**: The platform is not supported. It leads to the `NotSupported` status.
-- **UpdateFailure**: The endpoint operator failed to update the metrics collector. It leads to the `Degraded` status.
-- **UpdateSuccesful**: The endpoint operator successfully updated the metrics collector. It leads to the `Progressing` status.
+Those types are supported by standard reasons explaining the reason for the condition's last transition. The reasons are defined later in this document.
 
 ## Condition List Management
 
@@ -38,7 +31,9 @@ If the most recent condition is identical to the new condition, the status contr
 
 ## Status Synchronization
 
-The following diagram shows the actors involved in the status synchronization process and their high level interactions:
+The diagram below shows the actors involved in the status synchronization process and their high level interactions. 
+
+The hub cluster also deploys the metrics-collector, but it does not rely on the addon feature. Thus, the addon status reporting described here does not apply for the hub cluster addon.
 
 ```mermaid
 sequenceDiagram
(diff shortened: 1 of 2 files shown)


operators/endpointmetrics/doc/design/addon-status.md
```diff
@@ -0,0 +1,92 @@
...
+    Progressing --> Disabled: Disabled metrics
+    Available --> Progressing: Update succesful
+    Degraded --> Disabled: Disabled metrics
+    Degraded --> Progressing: Update succesful
+    Progressing --> NotSupported: No Prometheus install and not ocp
+    Available --> Degraded: Update failure
+    Available --> Disabled: Disabled metrics
+```
+
+Open questions:
```
saswatamcode, line 64: Since the questions already have answers, maybe this can just be FAQ?

operators/endpointmetrics/doc/design/addon-status.md
```diff
@@ -0,0 +1,92 @@
...
+
+- The transition `Degraded -> Available` can only happen if the metrics-collector is responsible for the degraded state.
+- The transition `Available -> Degraded` only happens if the metrics collector fails to forward metrics over a certain period of time to avoid flapping state.
+
+### UWL metrics collector
+
+Questions:
+
+- Do nothing? 
+- Append to the conditions list?
```
saswatamcode, line 91: IMO we really should be adding the status for UWL too here.

operators/endpointmetrics/doc/design/addon-status.md
thibaultmg: Removing the section in new version

operators/endpointmetrics/doc/design/addon-status.md
thibaultmg: Yes, will do

douglascamata:
I'm missing an analogy here to help visualize what's the objective of this new design. Overall, I think we need a solution similar to the Node status conditions. I think this design might be inspired by them, but I'm not sure. 

Node's in k8s can have a bunch of conditions with different types:

* Ready.
* DiskPressure.
* MemoryPressure.
* PIDPressure.
* NetworkUnavailable.

All these conditions are totally independent and more than one of them can be present at the same time.

Note that these conditions follow the conventions that "abnormal is true". Example: if the "normal" state is to not have disk pressure, we know that the default value has to be `DiskPressure=False`. We need to decide which polarity we will adopt: positive `true` is good/healthy, while negative means `false` is good/healthy. 

Considering we **must** have those conditions and can't have super custom ones, I think the proposal is good. 

When aggregating the spokes addon status into the hub's,  will we increase a custom "message" that will allow a reader to quickly identify the spoke(s) with issues?

A small note: I don't see a way for the Endpoint Metrics Operator to transition into "Available". That stats only has outgoing transitions (update successful/failure, disabled metrics), but no incoming transitions.

thibaultmg:
> Overall, I think we need a solution similar to the Node status conditions. I think this design might be inspired by them, but I'm not sure

No, I did not know about it, but I guess it's good news.

> All these conditions are totally independent and more than one of them can be present at the same time.

Yes, this is the spirit of conditions. They can be treated as simple observations. But in our case, IIUC, we are reporting to the hub wether we are in one of the following states: `Available`, `Progressing`, `Degraded`.  See [addon documentation](https://open-cluster-management.io/concepts/addon#add-on-healthiness).
These are mutually exclusive states. Hence the state machine approach. Also, as both the operator and collector are updating the same observations, we need strict rules to avoid flapping and inconsistencies.
For the other states, we can treat them as independent observations.

> Note that these conditions follow the conventions that "abnormal is true". 

I haven't described this part, will do some additions.

> When aggregating the spokes addon status into the hub's, will we increase a custom "message" that will allow a reader to quickly identify the spoke(s) with issues?

This is the idea. I didn't want to go too deep into details so that this document remains easy to maintain. But I'll add a word about it.

> I don't see a way for the Endpoint Metrics Operator to transition into "Available". That stats only has outgoing transitions (update successful/failure, disabled metrics), but no incoming transitions.

The operator only installs and updates the collectors. It can't state wether it is functioning properly, only the collector can make this "observation".

Given our conversation, I wonder wether we can reduce the condition types to only the 3 standard ones, plus a unique one by collector. Because to me, `NotSupported` type could just be `Degraded`, and maybe the same for `Disabled`? I must check what is reported for the addon when it is in that state.

douglascamata:
@thibaultmg I'm aware of the specific statuses that the addons should have according to their documentation, that's why I wrote `Considering we must have those conditions and can't have super custom ones (we don't have decision power on the status of the addons), I think the proposal is good.`. 

I'm not saying we need extra status conditions like the Node object has. They would definitely be useful though, as they can indicate the problem directly in the addon resource without requiring a deeper investigation. Paired with what I mentioned about the message of the status aggregating the names of the spokes with issues, this could easily tell anyone `clusters X, Y, and Z are failing to send metrics`, `clusters A, B, C are failing to upgrade`, `clusters Q, W, E are disabled` and so on.

douglascamata:
On a separate note, we have to be careful with the state and their interaction with upgrades. For instance, we could end up with an upgrade blocked because fo the status of the monitoring addons.

thibaultmg:
I should have addressed most comments with the latest update. 

* I now focus on reasons for better clarity, avoiding mixing condition types
* Added a reasons managers section that should clarify responsibilities
* Simplified state transitions diagram
* And a few other minor updates

> On a separate note, we have to be careful with the state and their interaction with upgrades. For instance, we could end up with an upgrade blocked because fo the status of the monitoring addons.

@douglascamata I agree, let me know if you think of a scenario where this can happen.

PR #1473, opened 2024-06-10
Prevent unnecessary update of observatorium in MCO

coleenquadros:
Everytime we create an Observatorium CR we create a new tenantID for `default` tenant and then ensure that the tenant ID is the same as the ID created the first time. Because of which the comparison in old and new spec differs and we update the spec again although nothing has changed.


jacobbaungard (approved):
Nice find!

/lgtm

jacobbaungard:
Worth a backport perhaps?

PR #1475, opened 2024-06-11
docker: Update sha for ubi image in dev dockerfile

philipgough:
Updating since the bump in prom versions means newer prometheus 0.51 requires go 1.21


periklis (approved):
LGTM

jacobbaungard (approved):
/lgtm

PR #1476, opened 2024-06-11
Wait for vendor label on managedcluster object

coleenquadros:
We want to ensure that vendor label is set on a cluster before proceeding to decide if we need to install Prometheus on a cluster
https://issues.redhat.com/browse/ACM-11673

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/predicate_func.go
+++ b/operators/multiclusterobservability/controllers/placementrule/predicate_func.go
@@ -35,6 +35,9 @@ func getClusterPreds() predicate.Funcs {
 			return false
 		}
 
+		if !areManagedClusterLabelsReady(e.Object) {
+			return false
+		}
 		updateManagedClusterList(e.Object)
 		updateManagedClusterImageRegistry(e.Object)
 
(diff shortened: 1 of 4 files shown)


operators/multiclusterobservability/controllers/placementrule/predicate_func.go
```diff
@@ -34,9 +34,22 @@ func getClusterPreds() predicate.Funcs {
 		if isAutomaticAddonInstallationDisabled(e.Object) {
 			return false
 		}
-		if e.Object.GetName() != "local-cluster" {
-			updateManagedClusterList(e.Object)
+
+		label := "vendor"
+		if err := waitForLabel(e.Object, label); err != nil {
```
elgnay, line 39: Will this function call blocks the whole prediction? An alternative way is to handle the wait in the reconciliation function of your controller. for example,

1. If the cluster has no the vendor label, requeue the managed cluster after an interval, like 5 seconds;
2. otherwise if the label value is not OpenShift, run into the non-OpenShift logic;
3. if the openShiftVersion label does not exist yet, requeue the managed cluster after an interval;
4. once openShiftVersion label is in place, you are able to move forward to the OpenShift logic;

subbarao-meduri:
@coleenquadros This fix also need to be backported to service releases.

subbarao-meduri (approved):
/lgtm

PR #1434, opened 2024-05-14
Retry manifest work update on conflict

thibaultmg:
Adds backoff retry logic on calls updating manifest works in the hope it removes many errors we see in the mco operator logs from the placement_rule controller.

Exemple:

```
2024-05-06T09:12:01.020Z        ERROR   controller_placementrule        Failed to update monitoring-endpoint-monitoring-work work       {"error": "Operation cannot be fulfilled on manifestworks.work.open-cluster-ma
nagement.io \"miroshift-clusterbot6-observability\": the object has been modified; please apply your changes to the latest version and try again"}
```

PR #1477, opened 2024-06-11
Fix manifestwork update due to incorrect obj compare

coleenquadros:
The wrong object was passed to the obj_compare function during refactor [PR](https://github.com/stolostron/multicluster-observability-operator/pull/1434) causing the manifestwork to keep updating in MCO


thibaultmg (approved):
/lgtm

PR #1478, opened 2024-06-11
Automate go tool binary installations across scripts and Makefiles

periklis:
The following PR provides a couple hygiene improvements to the Makefile targets used to locally develop and test the MCO operator, i.e.:
- Automates all go tool binaries (e.g. kustomize, operator-sdk, etc) to be automatically installed in `GOBIN` via [bingo](https://github.com/bwplotka/bingo/) once a Makefile target is executed. In turn this eliminates local binary installations on system/home paths.
- Refactors using the native `$(MAKE) -C` syntax to call makefile targets in subdirectories e.g. `operators/multiclusterobservability`
- Moves all operator related targets from the top-level directory Makefile to the corresponding subdirectories, i.e. `docker-build`, `docker-build-local`, `docker-push`. However it keeps top-level targets that call the subdirectory-level ones for compatibility.
- Unifies all `*-tests` targets to one `tools` target for the remaining non-bingo-managed tools in particular `kubectl` and `jq`.

### Notes for the reviewer

The main motivation of this PR is to cleanup the local development environment for newcoming contributors. However moving the operator-relevant Makefile targets from the top-level directory to the according subdirectories is the main preparation for migrating the operator-sdk from `v1.4.2` to the latest available version.


.bingo/Variables.mk
```diff
@@ -29,18 +35,48 @@ $(GOIMPORTS): $(BINGO_DIR)/goimports.mod
...
+	@echo "(re)installing $(GOBIN)/gojsontoyaml-v0.1.0"
+	@cd $(BINGO_DIR) && GOWORK=off $(GO) build -mod=mod -modfile=gojsontoyaml.mod -o=$(GOBIN)/gojsontoyaml-v0.1.0 "github.com/brancz/gojsontoyaml"
+
 GOLANGCI_LINT := $(GOBIN)/golangci-lint-v1.54.2
 $(GOLANGCI_LINT): $(BINGO_DIR)/golangci-lint.mod
 	@# Install binary/ries using Go 1.14+ build command. This is using bwplotka/bingo-controlled, separate go module with pinned dependencies.
 	@echo "(re)installing $(GOBIN)/golangci-lint-v1.54.2"
 	@cd $(BINGO_DIR) && GOWORK=off $(GO) build -mod=mod -modfile=golangci-lint.mod -o=$(GOBIN)/golangci-lint-v1.54.2 "github.com/golangci/golangci-lint/cmd/golangci-lint"
 
+KIND := $(GOBIN)/kind-v0.22.0
```
saswatamcode, line 50: Do we need to pin `kind` via bingo here? Usually, most people have it installed already via brew or just binary, and installing via go might cause issues there. Maybe we can just install it using binaries like we did before?

saswatamcode (commented):
Thanks for unifying!

saswatamcode (approved):
/lgtm

PR #1479, opened 2024-06-11
Add Observatorium API's external host to tls certs SAN

douglascamata:
If we have an intermediate component, like a load balancer, between Observatorium API and the metrics-collector, we need that component's URL to be part of the SANs.


philipgough (approved):
/lgtm

PR #1484, opened 2024-06-13
Fix code setting global var with the MCO CR name

douglascamata:
`config.SetMonitoringCRName` and `config.GetMonitoringCRName` are the functions used to get and set, respectively, the global with the MCO CR name.

I noticed though the name was only set in the `CreateFunc` of the MCO predicate. This meant that if the MCO controller was not running when the MCO CR was created, nothing would set the CR name. 

I detected this because it would cause an issue in the `config.GetObsAPIExternalHost` function.

Considering that `config.GetMonitoringCRName` is used in many places throughout the code, it must be causing lots of problems.


philipgough (approved):
/lgtm

PR #1485, opened 2024-06-13
ACM-11985: fix addon status update

thibaultmg:
I previously made a change to use the conditions as an history log, but this is not possible.
This PR reverts to in-place update of status conditions for the addon on the spoke.


philipgough (approved):
/lgtm

PR #1486, opened 2024-06-13
Make URL handling safer and less wasteful

douglascamata:
Basically updated `GetObsAPIExternalHost` and `GetAlertmanagerEndopint` to return a `url.URL` instead of string and updated the code accordingly.


philipgough (approved):
/lgtm

PR #1489, opened 2024-06-17
backport status fix

thibaultmg:


jacobbaungard:
Could you please add some more information? No PR description or proper commit messages makes it hard to understand. Which main PRs are this supposed to backport specifically?

I see https://github.com/stolostron/multicluster-observability-operator/commit/bdf7c3c0c101959c5f28d187145007adde6eb236 should be a backport of https://github.com/stolostron/multicluster-observability-operator/pull/1487 but where does all the extra changes to the test come from in the 2.10 version?

operators/endpointmetrics/pkg/util/status.go
```diff
@@ -16,47 +16,47 @@ import (
...
 const (
-	DeployedStatus           StatusConditionName = "Deployed"
-	DisabledStatus           StatusConditionName = "Disabled"
-	DegradedStatus           StatusConditionName = "Degraded"
-	NotSupportedStatus       StatusConditionName = "NotSupported"
-	MaxStatusConditionsCount                     = 10
+	Deployed                 ConditionReason = "Deployed"
+	Disabled                 ConditionReason = "Disabled"
+	Degraded                 ConditionReason = "Degraded"
+	NotSupported             ConditionReason = "NotSupported"
```
douglascamata, line 25: I didn't see the original PR,  but I still prefer the old names. It's a quite common pattern in Go to const values of a typed-string after the original type's name, i.e. `ConditionReasonDeployed` and so on. Particularly helpful when this lives in an `util` package.

douglascamata (approved):
/lgtm

operators/endpointmetrics/pkg/util/status.go
thibaultmg: I agree. I could have wait.
I've changed it because I'm putting these status management files into their own `status` package in a subsequent PR...

PR #1496, opened 2024-06-17
Backport MCO reconcile issues to 2.10

coleenquadros:


operators/multiclusterobservability/controllers/multiclusterobservability/observatorium.go
```diff
@@ -159,20 +159,22 @@ func GenerateObservatoriumCR(
...
 		for _, oldTenant := range oldSpec.API.Tenants {
 			updateTenantID(&newSpec, newTenant, oldTenant, i)
 		}
 	}
 
+	oldSpecBytes, _ := yaml.Marshal(oldSpec)
+	newSpecBytes, _ := yaml.Marshal(newSpec)
+	if bytes.Equal(newSpecBytes, oldSpecBytes) &&
+		labels[obsCRConfigHashLabelName] == observatoriumCRFound.Labels[obsCRConfigHashLabelName] {
+		return nil, nil
```
moadz, line 175: Through no fault of your own, but we're inconsistent with error handling in this reconcile loop. We alternate between returning an empty string `""` and the `nil` type. Later we pass the actual value to the `obsCRConfigHashLabelName`, do we know how this `nil` is handled in this case?

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -395,7 +396,7 @@ func createManifestWorks(
 	log.Info(fmt.Sprintf("Cluster: %+v, Spec.NodeSelector (after): %+v", clusterName, spec.NodeSelector))
 	log.Info(fmt.Sprintf("Cluster: %+v, Spec.Tolerations (after): %+v", clusterName, spec.Tolerations))
 
-	if clusterName != clusterNamespace {
+	if clusterName == localClusterName {
```
moadz, line 399: 👍🏽 this is much more explicit

moadz (approved):
LGTM thanks for backporting

PR #1497, opened 2024-06-18
Wait for OpenshiftVersion label when managed cluster is added

coleenquadros:
Added a condition to ensure that when vendor label is set to `openshift` and OpenShiftVersion is not we should not reconcile MCO to add observability stack in managed cluster
https://issues.redhat.com/browse/ACM-12249

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
+++ b/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
@@ -100,12 +100,12 @@ func (r *PlacementRuleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, err
 	}
 	if err == nil {
-		vendor, ok := cluster.Labels["vendor"]
+		vendor, ok := cluster.GetLabels()["vendor"]
 		if !ok {
 			return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 		}
 		if vendor == "OpenShift" {
-			_, ok := cluster.Labels["openshiftVersion"]
+			_, ok := cluster.GetLabels()["openshiftVersion"]
 			if !ok {
 				return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 			}
--- a/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
+++ b/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
@@ -7,7 +7,6 @@ package placementrule
 import (
 	"context"
 	"errors"
-	"fmt"
 	"reflect"
 	"strings"
 	"sync"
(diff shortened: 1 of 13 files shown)


operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -591,7 +591,8 @@ func areManagedClusterLabelsReady(obj client.Object) bool {
 	if _, openshiftVersionPresent := obj.GetLabels()["openshiftVersion"]; openshiftVersionPresent && openshiftVendor {
 		return true
 	}
-	if vendorOk && vendor != "auto-detect" {
+
+	if vendorOk && vendor != "auto-detect" && vendor != "OpenShift" {
```
douglascamata, line 595: Could we document what this function has to do in a comment/doc above it? It's written in a way that's very difficult to understand. I also suggest a small refactor: 

```go
func areManagedClusterLabelsReady(obj client.Object) bool {
	vendor, foundVendor := obj.GetLabels()["vendor"]
	if !foundVendor || vendor == "" {
		log.Info("ManagedCluster labels are not ready", "cluster", obj.GetName())
		return false
	}

	if vendor == "OpenShift" {
		_, foundOpenshiftVersion := obj.GetLabels()["openshiftVersion"]
		return foundOpenshiftVersion
	}

	if foundVendor && vendor != "auto-detect" {
		return true
	}


	log.Info("ManagedCluster labels are not ready", "cluster", obj.GetName())
	return false
}
```

Please double check that the logic I wrote in the function comment is correct.

douglascamata: Hum, I ate a log line when there's no `openshiftVersion` label. Also could simplify even more:

```go
// areManagedClusterLabelsReady check if labels automatically set in the managed cluster
// are ready to be accessed. These labels are: "vendor" and "openshiftVersion".
// Labels are considered not ready when:
//
// - The "vendor" label isn't found.
// - The "vendor" label has value "auto-detect".
// - The "vendor" has value "openshift" but there is no "openshiftVersion".
//
func areManagedClusterLabelsReady(obj client.Object) bool {
	labels := obj.GetLabels()
	vendor, foundVendor := labels["vendor"]

	if !foundVendor || vendor == "" || vendor == "auto-detect" {
		log.Info("ManagedCluster labels are not ready", "cluster", obj.GetName())
		return false
	}

	if vendor == "OpenShift" {
		_, foundOpenshiftVersion := labels["openshiftVersion"]
		if !foundOpenshiftVersion {
			log.Info("ManagedCluster labels are not ready", "cluster", obj.GetName())
                        return false
		}
	}

	return true
}
```

thibaultmg (approved):
/lgtm

PR #1500, opened 2024-06-19
ACM-11987: Add self-managed to ClusterManagementAddOn

jacobbaungard:
Since we're currently not correctly set "supportedConfigs" and "desiredConfig->specHash" on our ManagedClusterAdddon objects, these will never get past "Progressing" in 2.11. This is because the "addon-management-controller" is now specifically looking for these fields to be set.

As a workaround we set `addon.open-cluster-management.io/lifecycle: self` to ensure this does not happen. This should mirror the current behaviour of 2.10.

### DIFF
--- a/operators/multiclusterobservability/pkg/util/clustermanagementaddon.go
+++ b/operators/multiclusterobservability/pkg/util/clustermanagementaddon.go
@@ -49,6 +49,22 @@ func CreateClusterManagementAddon(c client.Client) (
 	} else if err != nil {
 		log.Error(err, "Cannot create observability-controller clustermanagementaddon")
 		return nil, err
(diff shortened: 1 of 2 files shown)


jacobbaungard:
Potentially new flaky test? Not sure why this functionality would cause that though.. Order of the labels are occasionally the other way around.

```
 === RUN   TestUpdateAllManagedClusterLabelNames/Add_new_labels
I0619 16:27:17.877657   23955 util.go:125] added managedcluster label: label1
I0619 16:27:17.877668   23955 util.go:125] added managedcluster label: label2
I0619 16:27:17.877730   23955 util.go:189] managed ignore list is empty
    util_test.go:751: syncLabelList.RegexLabelList = [label2 label1], want [label1 label2] 
```

douglascamata (approved):
/lgtm

PR #1504, opened 2024-06-24
Metrics collector refactor

thibaultmg:
As part of my work on microshift, I've done some refactoring that I am extracting here in its own PR. It includes:

- Move status update from `util` package into its own.
- Isolating `metrics collector` related functions into its own package. It encapsulates all the logis, including status update. Refactoring tests to table tests, keeping same coverage.
- Moving IsHypershift into its own package.
- Removing global variables as much as possible to improve testability.
- Moving env var consumptions into main to improve packages testability.
- Simplifying `match_evaluator`
- Refactoring integration tests to enable multiple tests reusing the same etcd and api-server
- I'm also now requeuing the reconcile request if `metricsCollector.Update` fails with some types of errors. This might not be worth it. Let me know what you think.


operators/endpointmetrics/pkg/util/error.go
douglascamata, line 1: Maybe this function should be in a package called `error` instead? `util` package is usually an anti-pattern.

douglascamata (approved):
I have a minor comment that can be handled later. LGTM anyway.

/lgtm

operators/endpointmetrics/pkg/util/error.go
thibaultmg: Yes good point. I'll try to embed this in another PR

PR #1507, opened 2024-06-24
ACM-10812: fix status update on collector

thibaultmg:
Fixes:

- The status update on the collector does not happen if it finds an existing condition of the same type even when it is not the latest one. For example, if the most recent condition is `Progressing` but the last condition in the list is `Available`, it will not update this last condition if the new one is also `Available`, keeping the state in `Progressing`.
- Also it ensures that only one of the main conditions expected by ACM (Progressing, Available, Degraded) is set to `true`. The status display in the console seems to pick the first condition set to `true`, even if not the most recent one.


saswatamcode (approved):

thibaultmg:
e2e is flaky, was passing earlier with current state. Merging

PR #1518, opened 2024-06-26
ACM-12406 - 2.10: Remove custom allValue for optimized dash

jacobbaungard:
The custom all value, `.+` was meant to remove very long list of clusters in (default) queries when all cluster was selected, i.e one would see queries like these:

```
{cluster=~"(local-cluster|spoke-1|spoke-2|spoke-3|spoke-4....)"
```

Unfortunately adding this custom all value seem to have completely broken the ability to filter cluster on the dashboard, hence removing this for now.


thibaultmg (approved):
/lgtm

PR #1520, opened 2024-07-01
*ks: correctly define cluster:node_cpu:ratio rule

jacobbaungard:
There was a bug in the rule imported from kube-prometheus causing the rule not to correctly return any results. This caused the data to be missing on the optimized dashboard from *ks clusters.

Upstream fix was here:
https://github.com/prometheus-operator/kube-prometheus/pull/1628


thibaultmg (approved):
/lgtm

PR #1522, opened 2024-07-01
Fix `dnf install` to prevent interactive confirmation

douglascamata:
This should unblock https://github.com/openshift/release/pull/53852, which gets stuck on the build process because of the interactive confirmation prompt that these `dnf install` commands were triggering.


jacobbaungard (approved):
/lgtm

PR #1523, opened 2024-07-02
Upgrade packages

thibaultmg:
For scraping ETCD metrics on Microshift spokes, I need to upgrade the `prometheus-operator`. Achieving that required upgrading many other packages. 
I took this occasion to remove almost completely the `replace` directives.
To be noted: this now requires Go 1.22.


philipgough (approved):
/lgtm

PR #1526, opened 2024-07-08
ACM-11757: refactor status update management to avoid flapping status

thibaultmg:
This PR changes how the addon status reporting is managed, based on [this specification](https://github.com/stolostron/multicluster-observability-operator/pull/1472). 

It first implements a status package that enables atomic update of each component's status  by adding two specific condition types in the addon conditions list: `MetricsCollector` and `UwlMetricsCollector`. 
This package is used by the endpoint operator to set the `UpdateSuccessful` or `UpdateFailed` reasons (plus some others), and by the metrics collector to set the `ForwardSuccessful` or `ForwardFailed` reasons for each collector.

In order to avoid the flapping behavior described in the linked issue, it forbids some transitions such as `UpdateFailed` -> `ForwardXxx`. The idea is that to report a consistent status, the collector must first be in the state required by the endpoint operator.

These specific condition types are then aggregated by the status controller into the expected condition types by ACM (`Available`, `Degraded` and `Progressing`).

One edge case that is not handled here is if the collector pod stops after having set the reason to `ForwardSuccessful`. In such a case, we get an `Available` status while it is not working. We could add more logic, requiring the collector to do regular updates on its condition timestamp and the controller would monitor this. Let's keep it simple while nobody complains about it 😄


operators/endpointmetrics/controllers/status/status_controller.go
```diff
@@ -9,22 +9,82 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"sort"
 	"time"
 
+	"github.com/go-logr/logr"
```
philipgough, line 15: nit on the imports getting messed up here!

thibaultmg: what do you mean? Looks good to me.

collectors/metrics/pkg/status/status.go
```diff
@@ -7,39 +7,35 @@ package status
 import (
 	"context"
 	"errors"
-	"fmt"
+	"log/slog"
```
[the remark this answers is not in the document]
thibaultmg: The collector and operator are using different loggers. go-kit logger can be adapted to the logr.Logger interface, but it requires yet another dependency.
So we have two solutions:

- either we instantiate a logr logger
- or we can transform to logr easily as logr provides an adaptor. I went for this one, but I don't remember why. Probably because it is in the standard library.

I can implement the other solution if you want. Let me know.

philipgough (approved):
/lgtm

PR #1533, opened 2024-07-18
[ACM-12773] Fix flaky test

thibaultmg:
CI upgrade to go 1.21 fails due to unsorted list from a map: https://github.com/openshift/release/pull/54478
This PR fixes this.

### DIFF
--- a/proxy/pkg/util/util.go
+++ b/proxy/pkg/util/util.go
@@ -747,6 +747,7 @@ func getClusterList(userMetricsAccess map[string][]string) []string {
 	for clusterName := range userMetricsAccess {
 		clusterList = append(clusterList, clusterName)
 	}
+	slices.Sort(clusterList)
 	return clusterList
 }
 


saswatamcode (approved):

proxy/pkg/util/util.go
```diff
@@ -747,6 +747,7 @@ func getClusterList(userMetricsAccess map[string][]string) []string {
 	for clusterName := range userMetricsAccess {
 		clusterList = append(clusterList, clusterName)
 	}
+	slices.Sort(clusterList)
```
moadz, line 750: Why would this be tripped by go1.21 update? Is there a test that asserts a sorted slice?

thibaultmg: I don't know how this is triggered. But I ran it two times and it failed in each. I remember that go removed the random keys ordering when looping through maps, maybe this is it 🤷 . In any case, I don't think it hurts having a. consistent ordering.
There is a test that expects the cluster list to be in order. Although, it is not its purpose, but it checks the list of clusters as "c1,c2"

PR #1535, opened 2024-07-18
ACM-12689: Adding rw mutexes for RBAC query proxy enforcement

moadz:
# Why
RBAC Query proxy utils.go watches for managed clusters asynchronously, whilst also facilitating query RBAC enforcement requests that both touch the same maps. 

E.g. 

1. Change happens to `clusterA` 
2. User queries data with perms to access `clusterA` simultaneously 

Resulting in the following error: 

```
I0710 15:50:22.603117       1 util.go:97] managedcluster label names update required: false
I0710 15:50:22.685471       1 util.go:263] changed a managedcluster: namgcbswd41p 
fatal error: concurrent map iteration and map writegoroutine 14780 [running]:
github.com/stolostron/multicluster-observability-operator/proxy/pkg/util.canAccessAllClusters(...)
```

I've added a RWmutex so that reads/writes can happen concurrently without panic.

### DIFF
--- a/proxy/pkg/util/util.go
+++ b/proxy/pkg/util/util.go
@@ -19,6 +19,7 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/go-co-op/gocron"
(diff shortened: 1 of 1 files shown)


philipgough (approved):

proxy/pkg/util/util.go
```diff
@@ -53,9 +54,13 @@ type AccessReviewer interface {
...
 
 var (
-	allManagedClusterNames      map[string]string
-	allManagedClusterLabelNames map[string]bool
-	accessReviewer              AccessReviewer
+	allManagedClusterNames    map[string]string
+	allManagedClusterNamesMtx sync.RWMutex
+
+	allManagedClusterLabelNames    map[string]bool
+	allManagedClusterLabelNamesMtx sync.RWMutex
```
douglascamata, line 61: What if we used [sync.Map](https://pkg.go.dev/sync#Map) from the stdlib?

PR #1547, opened 2024-07-24
Swap out cmo dependency for local structs

philipgough:


pkg/cmo/api.go
douglascamata, line 1: Maybe this could live in the already existing `operators/pkg` folder instead of a new `pkg` folder as the repo root?

douglascamata: Also it could be a good idea to have a comment documenting the original source of all this code (so that we know where to look for updates to it) and its own license & copyright notice (if needed).

douglascamata (approved):
/lgtm

PR #1551, opened 2024-07-24
[RHOBS-1044] Bump the Compactor's default CPU and memory requests

douglascamata:
Now the Compactor's `default` t-shirt size will be matching the small one. This should help to the Compactor get access to an adequate amount of CPU time and memory by default when it ends up in an overcommitted node.


philipgough (approved):
/lgtm

saswatamcode (approved):

PR #1553, opened 2024-07-25
Tests: Query grafana datasource via UID

jacobbaungard:
Set UID of datasources manually and use that UID to query using tests. The older API is deprecated and soon to be removed.


jacobbaungard:
We'd want to wait with this till after Grafana is actually updated, to make this is actually all we need. This might be broken on our older version of Grafana.

coleenquadros (approved):

PR #1554, opened 2024-07-30
Add missing X-Requested-With header to access quay.io

periklis:


periklis:
Looks like the `X-Requested-With` header is not needed on CI :thinking:

PR #1562, opened 2024-08-02
Update Prometheus crd after prom-operator upgrade to 0.75.2

coleenquadros:
Upgrade - https://github.com/stolostron/prometheus-operator/pull/62


philipgough (approved):
/lgtm

PR #1565, opened 2024-08-07
Fixing ACM-12888 by supplying a UID for fixed URL ref in the URL

moadz:
Grafana should accept this, provided there is no clash, and the linkref will now reference a fixed dashboard URL instead of a grafana generated UID which changes per-installation.

Also fixed the query param templating for the severity and health impact.

The value of `${severity:queryparam}` already includes the query param tag in it, so the previous template would substitute the tag with `var-health_impact=$var-health_impact=All` instead of `var-health_impact=All`.


saswatamcode (approved):
LGTM

PR #1585, opened 2024-08-20
Don't allow PRs to latest `release-` branch

jacobbaungard:
This automation checks whether a PR is merging to the latest release branch, if so the check fails.

Ensures our fast-foward logic doesn't break and main doesn't diverge.

### DIFF
--- a/.github/workflows/latest-branch.yml
+++ b/.github/workflows/latest-branch.yml
@@ -1,0 +1,23 @@
+name: Check pull request against latest release branch
+on:
+  pull_request_target:
+    types:
+      - opened
+      - reopened
+      - synchronize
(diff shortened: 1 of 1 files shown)


saswatamcode (approved):
Nice!

coleenquadros:
wont this block Konflux from creating PRs for 2.12? They are not set up to create against main

jacobbaungard:
> wont this block Konflux from creating PRs for 2.12? They are not set up to create against main

It will make a test/check fail. I _hope_ that when updating the PR to point to `main` instead the check will re-run and then pass.

Not sure what the best long-term approach for this is (i.e handling the Konflux PRs)

PR #1589, opened 2024-08-21
ACM-12794: forward auth header to datasource

jacobbaungard:
This allows rbac-query-proxy to authentication against obs-api. This is needed as Grafana no longer automatically forwards the header (and for now only does so in our patched Grafana version).


saswatamcode (approved):

philipgough (approved):
/lgtm

PR #1587, opened 2024-08-21
Get Oauth proxy image from ImageStreams resource

coleenquadros:
https://issues.redhat.com/browse/ACM-12538

 Global Hub Ref PR -  https://github.com/stolostron/multicluster-global-hub/pull/1010
  https://github.com/stolostron/multicluster-global-hub/pull/1042

PR #1575, opened 2024-08-15
endpointmetrics: Refactor manipulations on cluster monitoring config

philipgough:

PR #1591, opened 2024-08-26
Expose cmo config revert binary

philipgough:
This change creates a new binary inside the endpoint metrics operator to enable manually/programatically revert the config in the managed cluster if the Hub for some reason is not able to clean it up like https://issues.redhat.com/browse/ACM-13355.

This change is driven by a customer request in https://issues.redhat.com/browse/ACM-13149 where they are using the cleanup script https://github.com/stolostron/deploy/blob/master/hack/cleanup-managed-cluster.sh.

This change can be paired with https://github.com/stolostron/deploy/pull/276

### DIFF
--- a/operators/endpointmetrics/cmd/cmo-config-revert/main.go
+++ b/operators/endpointmetrics/cmd/cmo-config-revert/main.go
@@ -1,0 +1,40 @@
+// Copyright (c) Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+// Licensed under the Apache License 2.0
+
+package main
+
+import (
+	"context"
+	"flag"
+	"log"
+	"time"
+
+	"github.com/stolostron/multicluster-observability-operator/operators/endpointmetrics/controllers/observabilityendpoint"
+
+	"go.uber.org/zap/zapcore"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+func main() {
+	kubeClient, err := client.New(ctrl.GetConfigOrDie(), client.Options{})
+	if err != nil {
+		log.Fatalf("unable to create client: %v", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	opts := zap.Options{Development: true, TimeEncoder: zapcore.ISO8601TimeEncoder}
+	opts.BindFlags(flag.CommandLine)
+	flag.Parse()
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	if err := observabilityendpoint.RevertClusterMonitoringConfig(ctx, kubeClient); err != nil {
+		log.Fatalf("unable to revert cluster monitoring config: %v", err)
+	}
+	log.Println("reverted cluster monitoring config")
+}
--- a/operators/endpointmetrics/controllers/observabilityendpoint/ocp_monitoring_config.go
+++ b/operators/endpointmetrics/controllers/observabilityendpoint/ocp_monitoring_config.go
@@ -320,7 +320,7 @@ func createOrUpdateClusterMonitoringConfig(
 			return err
 		}
 		if !revertedAlready {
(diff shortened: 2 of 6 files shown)


coleenquadros:
My PR is failing due to the same reason -> 
https://github.com/stolostron/multicluster-observability-operator/pull/1587

`  Normal   Pulled                           90s (x3 over 107s)  kubelet            Container image "quay.io/stolostron/grafana:2.12.0-SNAPSHOT-2024-08-26-15-47-08" already present on machine
  Normal   Created                          90s (x3 over 107s)  kubelet            Created container grafana
  Normal   Started                          90s (x3 over 106s)  kubelet            Started container grafana
  Warning  BackOff                          89s (x3 over 104s)  kubelet            Back-off restarting failed container grafana in pod observability-grafana-7b5d4cfbcf-lwdbm_open-cluster-management-observability(50cfa7ee-2480-434f-b38a-b23724b4383c)
  Warning  FailedToRetrieveImagePullSec
  Warning  BackOff                          89s (x3 over 104s)  kubelet            Back-off restarting failed container grafana in pod observability-grafana-7b5d4cfbcf-lwdbm_open-cluster-management-observability(50cfa7ee-2480-434f-b38a-b23724b4383c)
  Warning  FailedToRetrieveImagePullSecret  77s (x7 over 107s)  kubelet            Unable to retrieve some image pull secrets (multiclusterhub-operator-pull-secret); attempting to pull the image may not succeed.`

operators/endpointmetrics/controllers/observabilityendpoint/ocp_monitoring_config.go
```diff
@@ -494,10 +494,10 @@ func unset(ctx context.Context, client client.Client, ns string) error {
...
-// revertClusterMonitoringConfig reverts the configmap cluster-monitoring-config and relevant resources
+// RevertClusterMonitoringConfig reverts the configmap cluster-monitoring-config and relevant resources
 // (observability-alertmanager-accessor and hub-alertmanager-router-ca) for the openshift cluster monitoring stack.
-func revertClusterMonitoringConfig(ctx context.Context, client client.Client) error {
-	log.Info("revertClusterMonitoringConfig called")
+func RevertClusterMonitoringConfig(ctx context.Context, client client.Client) error {
+	log.Info("RevertClusterMonitoringConfig called")
 
 	// try to retrieve the current configmap in the cluster
 	found := &corev1.ConfigMap{}
```
moadz, line 503: This is perhaps a more involved change, but instead of relying on a backup of the entire object, making a change and reverting to that - we only ever add `additionalAlertmanagerConfig`, should we change the revert logic to find the index in the config slice and just delete it instead? This means if the backup doesn't exist, but somehow endpointoperator has made the change it will always remove it if it finds it? 

We have the hub alertmanager svc hardcoded, so we could always use that to find which entry this endpoint operator is responsible for? 

As an extension, to avoid double alerting, we could also error if that list is >0 (user editing of this config map is discouraged by CMO, so we can be certain that array should only be one item long).

operators/endpointmetrics/cmd/cmo-config-revert/main.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/stolostron/multicluster-observability-operator/operators/endpointmetrics/controllers/observabilityendpoint"
+
+	"go.uber.org/zap/zapcore"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+func main() {
```
moadz, line 22: I like this idea! Good job :)

moadz (approved):

operators/endpointmetrics/controllers/observabilityendpoint/ocp_monitoring_config.go
philipgough: Yes, I think the function itself needs some work and improvement but I think it would be extending the scope of this jira/problem to do so here.

I have started a basic refactor in https://github.com/stolostron/multicluster-observability-operator/pull/1575 which reduces complexity in the first place and should essentially be a no-op right now for behaviour. I would lean towards us merging both of these PRs first before making any radical changes to how we handle the config.

saswatamcode (approved):
/lgtm

PR #1470, opened 2024-06-05
[release-2.12] Add support for mcoa-managed observability capabilities

periklis:
The following PR adds support to reconcile further observability capabilities via the [multicluster-observability-addon](https://github.com/rhobs/multicluster-observability-addon). The present implementation adds:
1. Support of the addon as separate MCO operand.
2. Extends the `MultiClusterObservability` CRD spec by a new section named `capabilities` as discussed here (TODO: Add link to DDR here).
3. For logs data collection it adds support for platform and user workloads.
4. For tracing data collection/instrumentation it adds support for user workloads.

**Special notes for your reviewer**:
To the test the following PR one needs to publish an image of the referenced dependency PR and applying the following instructions:
1. Install the `Advanced Cluster Management for Kubernetes` from the OperatorHub.
2. Create a `MultiClusterHub` resource.
3. Set the `MultiClusterHub` resource to pause reconciliation via: 
`kubectl -n open-cluster-management annotate mch multiclusterhub --overwrite mch-pause=true`
4. Build and publish the MCO operator image from this branch:
```
make docker-build-local IMAGE_TAG_BASE=quay.io/$YOUR_ORG/multicluster-observability-operator

docker push quay.io/$YOUR_ORG/multicluster-observability-operator:latest
```
5. Edit the image for the multicluster-observability-operator deployment
`kubectl -n open-cluster-management edit deployment multicluster-observability-operator`
6. Update the ClusterRole to add support for `create/update/delete/patch` for `AddonDeploymentConfig` resources:
`kubectl edit clusterrole open-cluster-management:multicluster-observability-operator:multicluster-observability-operator`
7. Update the CRD using the one in this branch to add support for the `capabilities` section.
8. Finally create an `MultiClusterObservability` custom resource as the following:
```yaml
apiVersion: observability.open-cluster-management.io/v1beta2
kind: MultiClusterObservability
metadata:
  annotations:
    mco-multicluster_observability_addon-image: quay.io/YOUR_ORG_HERE/multicluster-observability-addon:v0.0.1
  name: observability
spec:
  capabilities:
    platform:
      collection:
        logs:
          enabled: true
    userWorkloads:
      logs:
        collection:
          clusterLogForwarder:
            enabled: true
      traces:
        collection:
          instrumentation:
            enabled: true
          openTelemetryCollector:
            enabled: false
  observabilityAddonSpec: {}
  storageConfig:
    metricObjectStorage:
      name: thanos-object-storage
      key: thanos.yaml
```

[...]

PR #1592, opened 2024-08-27
update to deployer functions for AddOnDeploymentConfig and ClusterManagementAddOn

JoaoBraveCoding:
In https://github.com/stolostron/multicluster-observability-operator/pull/1470 we added support for deploying and configuring MCOA with the capabilities field. However, if we enabled a field in capabilities and then disabled it MCO would not reconcile the AddOnDeploymentConfig correctly due to DeepDerivative, which would return true instead of false.


philipgough (approved):
/lgtm

periklis (commented):
/lgtm

periklis (approved):
/lgtm

PR #1593, opened 2024-08-28
[ACM-13266] - Enable mounting of arbitrary Secrets by alertmanager

philipgough:
https://issues.redhat.com/browse/ACM-13266 - allows a user to mount Secrets in alertmanager from the `advancedConfig` option in MCO. 

Following the pattern of prometheus-operator, Secrets will be mounted as a volume named secret-<secret-name>. Secrets are mounted into /etc/alertmanager/secrets/<secret-name> in the 'alertmanager' container.


operators/multiclusterobservability/pkg/rendering/renderer_alertmanager.go
```diff
@@ -40,88 +40,105 @@ func (r *MCORenderer) newAlertManagerRenderer() {
...
+	alertManagerContainer := &spec.Containers[0]
+	configReloaderContainer := &spec.Containers[1]
+	oauthProxyContainer := &spec.Containers[2]
+	kubeRbacProxyContainer := &spec.Containers[3]
 
+	alertManagerContainer.ImagePullPolicy = imagePullPolicy
 	if *dep.Spec.Replicas > 1 {
 		for i := int32(0); i < *dep.Spec.Replicas; i++ {
-			args = append(args, "--cluster.peer="+
+			alertManagerContainer.Args = append(alertManagerContainer.Args, "--cluster.peer="+
```
saswatamcode, line 82: Can just use fmt.Sprint here instead of adding strings

saswatamcode (commented):
Thanks!

operators/multiclusterobservability/pkg/rendering/renderer_alertmanager.go
philipgough: yep. I didn't add this btw, just moved things around. Needs a rebase now anyway so will take care of it

saswatamcode (approved):

PR #1596, opened 2024-09-02
CI/Kind: Use latest snapshot for OCM version

jacobbaungard:
Instead of hardcoding a specific snapshot, we just use the latest release instead.


philipgough (approved):
/lgtm

PR #1605, opened 2024-09-06
Add a test for CMO Config watch

philipgough:


operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
```diff
@@ -555,7 +555,7 @@ func (r *ObservabilityAddonReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Watches(
 			&corev1.ConfigMap{},
 			&handler.EnqueueRequestForObject{},
-			builder.WithPredicates(getPred(clusterMonitoringConfigName, promNamespace, true, true, false)),
+			builder.WithPredicates(getPred(clusterMonitoringConfigName, promNamespace, true, true, true)),
```
saswatamcode, line 558: But I thought it was a delete operation?

philipgough: its not, this made no difference to the test, i was just covering our bases whilst testing. works with or without. In theory we only care about Update

PR #1610, opened 2024-09-06
[ACM-14076]: Ensure CMO ConfigMap is reconciled on any event

philipgough:
1. Adds an integration test for merging of CMO ConfigMap with other clients
2. Adds the Delete event to the Watch for CMO ConfigMap
3. Explicitly adds the CMO ConfigMap reference to the filtered cache

### DIFF
--- a/operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
+++ b/operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
@@ -49,14 +49,16 @@ var (
 const (
 	obAddonName                     = "observability-addon"
 	obsAddonFinalizer               = "observability.open-cluster-management.io/addon-cleanup"
-	promSvcName                     = "prometheus-k8s"
-	promNamespace                   = "openshift-monitoring"
 	openShiftClusterMonitoringlabel = "openshift.io/cluster-monitoring"
 	mtlsCertName                    = "observability-controller-open-cluster-management.io-observability-signer-client-cert"
 	mtlsCaName                      = "observability-managed-cluster-certs"
 	metricsCollectorName            = "metrics-collector-deployment"
 	uwlMetricsCollectorName         = "uwl-metrics-collector-deployment"
 	uwlNamespace                    = "openshift-user-workload-monitoring"
+)
+const (
+	promSvcName   = operatorconfig.OCPClusterMonitoringPrometheusService
+	promNamespace = operatorconfig.OCPClusterMonitoringNamespace
 )
 
 // ObservabilityAddonReconciler reconciles a ObservabilityAddon object.
@@ -555,7 +557,7 @@ func (r *ObservabilityAddonReconciler) SetupWithManager(mgr ctrl.Manager) error 
 		Watches(
 			&corev1.ConfigMap{},
 			&handler.EnqueueRequestForObject{},
-			builder.WithPredicates(getPred(clusterMonitoringConfigName, promNamespace, true, true, false)),
+			builder.WithPredicates(getPred(clusterMonitoringConfigName, promNamespace, true, true, true)),
 		).
 		Watches(
 			&appsv1.Deployment{},
--- a/operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller_integration_test.go
+++ b/operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller_integration_test.go
@@ -14,6 +14,9 @@ import (
 	"testing"
 	"time"
 
+	yaml2 "github.com/ghodss/yaml"
+	ocinfrav1 "github.com/openshift/api/config/v1"
+	cmomanifests "github.com/openshift/cluster-monitoring-operator/pkg/manifests"
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1alpha1"
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/stolostron/multicluster-observability-operator/operators/endpointmetrics/pkg/hypershift"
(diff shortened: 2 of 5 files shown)


operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
```diff
@@ -555,7 +557,7 @@ func (r *ObservabilityAddonReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Watches(
 			&corev1.ConfigMap{},
 			&handler.EnqueueRequestForObject{},
-			builder.WithPredicates(getPred(clusterMonitoringConfigName, promNamespace, true, true, false)),
+			builder.WithPredicates(getPred(clusterMonitoringConfigName, promNamespace, true, true, true)),
```
moadz, line 560: We discussed this, it probably makes sense to create the config map when it's missing to insure that alerts are being forwarded.

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller_integration_test.go
```diff
@@ -47,6 +50,154 @@ var (
...
+			},
+		},
+	}
+	if err := createResources(k8sClient, resourcesDeps...); err != nil {
+		t.Fatalf("Failed to create resources: %v", err)
+	}
+
+	mgr, err := ctrl.NewManager(testEnvHub.Config, ctrl.Options{
+		Scheme:  k8sClient.Scheme(),
+		Metrics: metricsserver.Options{BindAddress: "0"}, // Avoids port conflict with the default port 8080
```
moadz, line 116: 👍🏽 wildcard port, but it is reserved and we lose the ability to discover the port if we need it. In this case that's not a problem.

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller_integration_test.go
```diff
@@ -47,6 +50,154 @@ var (
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	go func() {
+		err = mgr.Start(ctx)
+		assert.NoError(t, err)
+	}()
+
+	cm := &corev1.ConfigMap{}
+	err = wait.Poll(1*time.Second, time.Minute, func() (bool, error) {
```
moadz, line 147: This might make the unit suite take a bit longer to run. Is it worth checking less often? Perhaps every 10s.

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller_integration_test.go
```diff
@@ -47,6 +50,154 @@ var (
...
+			return false, nil
+		}
+
+		return true, err
+	})
+	assert.NoError(t, err)
+
+	foundClusterMonitoringConfiguration := &cmomanifests.ClusterMonitoringConfiguration{}
+	err = yaml2.Unmarshal([]byte(cm.Data[clusterMonitoringConfigDataKey]), foundClusterMonitoringConfiguration)
+	assert.NoError(t, err)
```
moadz, line 159: super nit: Readability, I'm not sure if i'm remembering incorrectly, but i think you already spun this into a util function in the ocp_monitoring_config.go refactor.

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller_integration_test.go
```diff
@@ -47,6 +50,154 @@ var (
...
+
+	assert.Len(t, foundClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs, 1)
+	assert.Equal(t, foundClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs[0].Scheme, "https")
+
+	foundClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs[0].Scheme = "http"
```
moadz, line 164: super nit: might be worth having a JSON->JSON comparison for these diffs to improve readability instead of field extracts/asserts but that would require a refactor of this entire test case which neither of us is willing to do 😆

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller_integration_test.go
```diff
@@ -47,6 +50,154 @@ var (
...
+		if err != nil {
+			return false, nil
+		}
+
+		if foundUpdatedClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs[0].Scheme != "https" {
+			return false, nil
+		}
+
+		if foundUpdatedClusterMonitoringConfiguration.PrometheusK8sConfig.Retention != "infinity-and-beyond" {
+			return false, nil
```
moadz, line 192: super nit: same as above comment, might be better for test readability to have this as a straight expected/actual comparison. But not a must for this PR

operators/endpointmetrics/controllers/observabilityendpoint/ocp_monitoring_config.go
```diff
@@ -394,16 +394,20 @@ func createOrUpdateClusterMonitoringConfig(
 		// check if alertmanagerConfigs exists
 		if foundClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs != nil {
 			additionalAlertmanagerConfigExists := false
-			for _, v := range foundClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs {
+			var atIndex int
+			for i, v := range foundClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs {
```
moadz, line 398: Would it be easier here to have a func that just checks for the alertmanager URL that we care about, and returns true if it finds the object. Or conversely if we always just add it when there's a reconcile. Is there a cost to updating this field? (Does alertmanager restart)

moadz (approved):
LGTM thanks for this Philip! Heroic work.

operators/endpointmetrics/main.go
```diff
@@ -95,6 +95,8 @@ func main() {
 			{FieldSelector: namespaceSelector},
 			{FieldSelector: fmt.Sprintf("metadata.name==%s,metadata.namespace!=%s",
 				operatorconfig.AllowlistCustomConfigMapName, "open-cluster-management-observability")},
+			{FieldSelector: fmt.Sprintf("metadata.name==%s,metadata.namespace==%s",
+				operatorconfig.OCPClusterMonitoringConfigMapName, operatorconfig.OCPClusterMonitoringNamespace)},
```
moadz, line 99: Is it worth an inline comment to describe the issue we ran into with not specifying the namespace on a watched resource? or is it neither here nor there.

PR #1614, opened 2024-09-10
mcoa: make mcoa deployment optional

JoaoBraveCoding:
Tested and it works ✅ 

Capabilities off:

image

Capabilities on:

image

### DIFF
--- a/operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
+++ b/operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
@@ -291,6 +291,22 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
 		}
 	}
 
+	if !rendering.MCOAEnabled(instance) {
+		namespace, labels := renderer.NamespaceAndLabels()
+		toDelete, err := renderer.MCOAResources(namespace, labels)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		for _, res := range toDelete {
+			resNS := res.GetNamespace()
+			if err := deployer.Undeploy(ctx, res); err != nil {
+				reqLogger.Error(err, fmt.Sprintf("Failed to undeploy %s %s/%s",
+					res.GetKind(), resNS, res.GetName()))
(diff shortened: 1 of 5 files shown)


operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -291,6 +291,22 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
...
+		}
+		for _, res := range toDelete {
+			resNS := res.GetNamespace()
+			if err := deployer.Undeploy(ctx, res); err != nil {
+				reqLogger.Error(err, fmt.Sprintf("Failed to undeploy %s %s/%s",
+					res.GetKind(), resNS, res.GetName()))
+				return ctrl.Result{}, err
+			}
+		}
+	}
```
periklis, line 308: Can we flip the logic in general here? Instead of undeploying just skipping deployment in the first place?

JoaoBraveCoding: Not deploying the resources will not remove them from the cluster unless I'm missing something...
This PR does two things when capabilities are not enabled: 
1. Stops MCO from rendering the manifests
2. It will attempt to remove existing deployments of MCOA (to clean up from having it deployed)

operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -291,6 +291,22 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
 		}
 	}
 
+	if !rendering.MCOAEnabled(instance) {
```
thibaultmg, line 294: We could add a filter on the request and run this only when the MCO config is modified?

JoaoBraveCoding: Do we have an example on the code base similar? I'm not quite sure how to accomplish this or I'm misunderstanding something

thibaultmg (approved):
/lgtm

PR #1622, opened 2024-09-12
mcoa: allow users to edit the resource ClusterManagementAddOn

JoaoBraveCoding:
Edited the `ClusterManagementAddOn` to instead of using `instance` to use `instance-default`

Screenshot 2024-09-12 at 15 52 18


thibaultmg (approved):
/lgtm

PR #1627, opened 2024-09-13
More debug tips

thibaultmg:


docs/debug.md
```diff
@@ -1,3 +1,60 @@
+## Grafana Issues
+
+### Grafana is loading but dashbords show no data
+
+Assuming that thanos is working correctly (it is able to ingest and query data). You can proceed with the following steps to debug the issue:
+
+* Check if the Grafana exlorer is working by querying a random metric like `up`.
+  * If it is not working, check communication between the components on the read path using the user token as Grafana does.
```
jacobbaungard, line 8: Maybe add a small reference (or even a link) to the section about the read path communications check?

docs/debug.md
```diff
@@ -1,3 +1,60 @@
...
+  Make sure that you find the metrics you are looking for.
+  * If it fails check the network policies.
+* Check if the `serviceMonitor` for the target exists and is correctly configured. Make sure that relabelling rules are not impacting your metrics.
+* Check the currently loaded configuration of the prometheus pod to see if you find the scraping job named after it, and if it is correctly configured:
+```bash
+curl http://localhost:9090/api/v1/status/config | sed 's/\\n/\n/g' | grep "job_name: serviceMonitor/NAMESPACE/SERVICE_MONITOR_NAME" -A100 
+# or from your local machine, pipe in less and search for the job name
+curl -s http://localhost:9090/api/v1/status/config | jq -r '.data.yaml' | yq '.scrape_configs' | less
+```
+* Check the state of the target:
```
jacobbaungard, line 35: I found out that this is also visible in the openshift console under Observe->targets (see the screenshot in https://issues.redhat.com/browse/ACM-14296), we could add a small reference to this here, as I found that was actually a pretty easy way to check if the target is OK. One can also use that page to check if the target is in the scrape config.

jacobbaungard (commented):
Nice, just two minor comments !

docs/debug.md
thibaultmg: Nice, adding that

PR #1628, opened 2024-09-15
Make federate sharding possible in metrics-collector by using workers

saswatamcode:
Attempting to implement federate sharding by sharding the match rules (both passed as file and flag) between a number of workers. 

If the code looks messy, that is because it is, there is a lot of unused and scattered logic throughout metrics-collector. I have tried to refactor a lot of this

### DIFF
--- a/collectors/metrics/cmd/metrics-collector/main.go
+++ b/collectors/metrics/cmd/metrics-collector/main.go
@@ -61,7 +61,7 @@ func main() {
 		&opt.WorkerNum,
 		"worker-number",
 		opt.WorkerNum,
-		"The number of workers that will be used to send metrics.")
+		"The number of workers that will work in parallel to send metrics.")
 	cmd.Flags().StringVar(
 		&opt.Listen,
 		"listen",
@@ -355,7 +355,7 @@ func (o *Options) Run() error {
 		g.Add(func() error {
 			for i, shardWorker := range shardWorkers {
 				go func(i int, shardWorker *forwarder.Worker) {
-					fmt.Printf("Starting shard worker %d\n", i)
+					logger.Log(o.Logger, logger.Info, "msg", "Starting shard worker", "worker", i)
 					shardWorker.Run(ctx)
 				}(i, shardWorker)
 			}
@@ -435,7 +435,7 @@ func splitMatchersIntoShards(matchers []string, shardCount int) [][]string {
 }
 
 // Agent is the type of the worker agent that will be running.
-// They are classed according to what they collect.
+// They are classified according to what they collect.
 type Agent string
 
 const (
--- a/collectors/metrics/cmd/metrics-collector/main.go
+++ b/collectors/metrics/cmd/metrics-collector/main.go
@@ -44,7 +44,7 @@ func main() {
 		From:                   "http://localhost:9090",
 		Listen:                 "localhost:9002",
 		LimitBytes:             200 * 1024,
-		Rules:                  []string{`{__name__="up"}`},
+		Matchers:               []string{`{__name__="up"}`},
 		Interval:               4*time.Minute + 30*time.Second,
 		EvaluateInterval:       30 * time.Second,
 		WorkerNum:              1,
@@ -133,37 +133,26 @@ func main() {
 		opt.LimitBytes,
 		"The maxiumum acceptable size of a response returned when scraping Prometheus.")
 
-	// TODO: more complex input definition, such as a JSON struct
 	cmd.Flags().StringArrayVar(
-		&opt.Rules,
+		&opt.Matchers,
 		"match",
-		opt.Rules,
+		opt.Matchers,
 		"Match rules to federate.")
 	cmd.Flags().StringVar(
-		&opt.RulesFile,
+		&opt.MatcherFile,
 		"match-file",
-		opt.RulesFile,
+		opt.MatcherFile,
 		"A file containing match rules to federate, one rule per line.")
 	cmd.Flags().StringArrayVar(
 		&opt.RecordingRules,
 		"recordingrule",
 		opt.RecordingRules,
 		"Define recording rule is to generate new metrics based on specified query expression.")
-	cmd.Flags().StringVar(
-		&opt.RecordingRulesFile,
-		"recording-file",
-		opt.RulesFile,
-		"A file containing recording rules.")
 	cmd.Flags().StringArrayVar(
 		&opt.CollectRules,
 		"collectrule",
 		opt.CollectRules,
 		"Define metrics collect rule is to collect additional metrics based on specified event.")
-	cmd.Flags().StringVar(
-		&opt.RecordingRulesFile,
-		"collect-file",
-		opt.RecordingRulesFile,
-		"A file containing collect rules.")
 
 	cmd.Flags().StringSliceVar(
 		&opt.LabelFlag,
@@ -209,13 +198,6 @@ func main() {
 		"log-level",
 		opt.LogLevel,
 		"Log filtering level. e.g info, debug, warn, error")
-
-	// deprecated opt
-	cmd.Flags().StringVar(
-		&opt.Identifier,
-		"id",
-		opt.Identifier,
-		"The unique identifier for metrics sent with this client.")
 
 	// simulation test
 	cmd.Flags().StringVar(
@@ -266,12 +248,10 @@ type Options struct {
 	AnonymizeSalt     string
 	AnonymizeSaltFile string
 
-	Rules              []string
-	RulesFile          string
-	RecordingRules     []string
-	RecordingRulesFile string
-	CollectRules       []string
-	CollectRulesFile   string
+	Matchers       []string
+	MatcherFile    string
+	RecordingRules []string
+	CollectRules   []string
 
 	LabelFlag []string
 	Labels    map[string]string
@@ -281,9 +261,6 @@ type Options struct {
 
 	LogLevel string
 	Logger   log.Logger
-
-	// deprecated
-	Identifier string
 
 	// simulation file
 	SimulatedTimeseriesFile string
@@ -432,8 +409,7 @@ func runMultiWorkers(o *Options, cfg *forwarder.Config) error {
 			ToUploadCA:              o.ToUploadCA,
 			ToUploadCert:            o.ToUploadCert,
 			ToUploadKey:             o.ToUploadKey,
-			Rules:                   o.Rules,
-			RenameFlag:              o.RenameFlag,
+			Matchers:                o.Matchers,
 			RecordingRules:          o.RecordingRules,
 			Interval:                o.Interval,
 			Labels:                  map[string]string{},
@@ -570,6 +546,10 @@ func initConfig(o *Options) (*forwarder.Config, error) {
 	transformer.With(metricfamily.TransformerFunc(metricfamily.PackMetrics))
 	transformer.With(metricfamily.TransformerFunc(metricfamily.SortMetrics))
 
+	// TODO(saswatamcode): Kill this feature.
+	// This is too messy of an approach, to get hypershift specific labels into metrics we send.
+	// There is much better way to do this, with relabel configs.
+	// A collection agent shouldn't be calling out to Kube API server just to add labels.
 	if !o.DisableHyperShift {
 		isHypershift, err := metricfamily.CheckCRDExist(o.Logger)
 		if err != nil {
@@ -599,16 +579,39 @@ func initConfig(o *Options) (*forwarder.Config, error) {
 		}
 	}
 
+	// Configure matchers.
+	matchers := o.Matchers
+	if len(o.MatcherFile) > 0 {
+		data, err := os.ReadFile(o.MatcherFile)
+		if err != nil {
+			return nil, fmt.Errorf("unable to read match-file: %w", err)
+		}
+		matchers = append(matchers, strings.Split(string(data), "\n")...)
+	}
+	for i := 0; i < len(matchers); {
+		s := strings.TrimSpace(matchers[i])
+		if len(s) == 0 {
+			matchers = append(matchers[:i], matchers[i+1:]...)
+			continue
+		}
+		matchers[i] = s
+		i++
+	}
+
 	f := forwarder.Config{
-		From:          from,
-		FromQuery:     fromQuery,
-		ToUpload:      toUpload,
-		FromToken:     o.FromToken,
-		FromTokenFile: o.FromTokenFile,
-		FromCAFile:    o.FromCAFile,
-		ToUploadCA:    o.ToUploadCA,
-		ToUploadCert:  o.ToUploadCert,
-		ToUploadKey:   o.ToUploadKey,
+		FromClientConfig: forwarder.FromClientConfig{
+			URL:       from,
+			QueryURL:  fromQuery,
+			Token:     o.FromToken,
+			TokenFile: o.FromTokenFile,
+			CAFile:    o.FromCAFile,
+		},
+		ToClientConfig: forwarder.ToClientConfig{
+			URL:      toUpload,
+			CAFile:   o.ToUploadCA,
+			CertFile: o.ToUploadCert,
+			KeyFile:  o.ToUploadKey,
+		},
 
 		AnonymizeLabels:   o.AnonymizeLabels,
 		AnonymizeSalt:     o.AnonymizeSalt,
@@ -617,8 +620,7 @@ func initConfig(o *Options) (*forwarder.Config, error) {
 		Interval:          o.Interval,
 		EvaluateInterval:  o.EvaluateInterval,
 		LimitBytes:        o.LimitBytes,
-		Rules:             o.Rules,
-		RulesFile:         o.RulesFile,
+		Matchers:          matchers,
 		RecordingRules:    o.RecordingRules,
 		CollectRules:      o.CollectRules,
 		Transformer:       transformer,
@@ -627,6 +629,10 @@ func initConfig(o *Options) (*forwarder.Config, error) {
 		SimulatedTimeseriesFile: o.SimulatedTimeseriesFile,
 	}
 
+	// TODO(saswatamcode): Evaluate better way for status reporting.
+	// Currently, it reports status for every 3 errors in forward request. This is too much of an overhead
+	// Every remote write request error does not need to be reported.
+	// Instead we can try to do this with metrics-collector meta-monitoring
 	if !o.DisableStatusReporting {
 		config, err := clientcmd.BuildConfigFromFlags("", "")
 		if err != nil {
--- a/collectors/metrics/cmd/metrics-collector/main.go
+++ b/collectors/metrics/cmd/metrics-collector/main.go
@@ -21,31 +21,38 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/oklog/run"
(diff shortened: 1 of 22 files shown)


operators/multiclusterobservability/api/shared/multiclusterobservability_shared.go
```diff
@@ -44,6 +44,14 @@ type ObservabilityAddonSpec struct {
 	// +kubebuilder:validation:Maximum=3600
 	Interval int32 `json:"interval,omitempty"`
 
+	// Workers is the number of workers that work parallelly to push metrics to hub server.
```
philipgough, line 47: nit: Workers is the number of workers that work in parallel ...   sounds more natural than the way this is written

collectors/metrics/cmd/metrics-collector/main.go
```diff
@@ -554,68 +537,241 @@ func initConfig(o *Options) (error, *forwarder.Config) {
 	transformer.With(metricfamily.TransformerFunc(metricfamily.PackMetrics))
 	transformer.With(metricfamily.TransformerFunc(metricfamily.SortMetrics))
 
-	isHypershift, err := metricfamily.CheckCRDExist(o.Logger)
-	if err != nil {
-		return err, nil
+	// TODO(saswatamcode): Kill this feature.
```
thibaultmg, line 540: I agree, most of what is done in the metrics collector is already setup in the endpoint-operator when generating the ServiceMonitors. The only missing labels are `managementcluster` and `managementclusterID`. They can be easily added.

collectors/metrics/cmd/metrics-collector/main.go
```diff
@@ -554,68 +537,241 @@ func initConfig(o *Options) (error, *forwarder.Config) {
...
+			continue
+		}
+		matchers[i] = s
+		i++
+	}
+
+	var statusClient client.Client
+
+	// TODO(saswatamcode): Evaluate better way for status reporting.
+	// Currently, it reports status for every 3 errors in forward request. This is too much of an overhead
```
thibaultmg, line 595: I'm not sure there is much to win here, given that it will be replaced by MCOA. In my view, it's not worth the effort. It's good enough.

collectors/metrics/cmd/metrics-collector/main.go
saswatamcode: Yup, we can think of a way to add it as a relabel config on Prom, using MCO. But probably in later PRs.

collectors/metrics/cmd/metrics-collector/main.go
saswatamcode: Hmm, MCOA would still use metrics-collector, right? Or are you saying we won't be reporting status from here in MCOA anymore?

collectors/metrics/cmd/metrics-collector/main.go
```diff
@@ -203,12 +196,17 @@ func main() {
 		opt.LogLevel,
 		"Log filtering level. e.g info, debug, warn, error")
 
-	// deprecated opt
-	cmd.Flags().StringVar(
-		&opt.Identifier,
-		"id",
-		opt.Identifier,
-		"The unique identifier for metrics sent with this client.")
+	cmd.Flags().BoolVar(
```
thibaultmg, line 199: What is the purpose for exposing these flags?

saswatamcode: e2e tests in the future. I already have one like so, but to add it here, would need some go.mod changes which we can tackle in later PRs.

```go
package e2e_test

import (
	"fmt"
	"os"
	"strconv"
	"testing"
	"time"

	"github.com/efficientgo/e2e"
	e2edb "github.com/efficientgo/e2e/db"
	e2einteractive "github.com/efficientgo/e2e/interactive"
	e2emon "github.com/efficientgo/e2e/monitoring"
	e2eobs "github.com/efficientgo/e2e/observable"
	"github.com/thanos-io/thanos/test/e2e/e2ethanos"

	"github.com/efficientgo/core/backoff"

	"github.com/efficientgo/core/testutil"
)

func TestMetricsCollectorE2E(t *testing.T) {
	e, err := e2e.NewDockerEnvironment("metrics-col")
	testutil.Ok(t, err)
	t.Cleanup(e.Close)

	m, err := e2emon.Start(e)
	testutil.Ok(t, err)

	prom := e2edb.NewPrometheus(e, "prom2")
	receive1 := e2ethanos.NewReceiveBuilder(e, "receiver-1").WithImage("quay.io/thanos/thanos:v0.35.1").WithIngestionEnabled().Init()

	testutil.Ok(t, e2e.StartAndWaitReady(prom, receive1))

	collector := NewMetricsCollector(e, "collector-1", MetricsCollectorOptions{
		From:     prom.InternalEndpoint("http"),
		To:       receive1.InternalEndpoint("remote-write"),
		Matchers: []string{`{__name__="up"}`, `{__name__="go_gc_duration_seconds"}`, `{__name__="go_goroutines"}`, `{__name__="go_memstats_alloc_bytes"}`, `{__name__="go_memstats_heap_alloc_bytes"}`, `{__name__="go_memstats_heap_idle_bytes"}`, `{__name__="go_memstats_heap_inuse_bytes"}`, `{__name__="go_memstats_heap_objects"}`, `{__name__="go_memstats_heap_released_bytes"}`, `{__name__="go_memstats_heap_sys_bytes"}`, `{__name__="go_memstats_last_gc_time_seconds"}`, `{__name__="go_memstats_lookups_total"}`, `{__name__="go_memstats_mallocs_total"}`, `{__name__="go_memstats_sys_bytes"}`, `{__name__="go_threads"}`, `{__name__="process_cpu_seconds_total"}`, `{__name__="process_open_fds"}`, `{__name__="process_resident_memory_bytes"}`, `{__name__="process_start_time_seconds"}`, `{__name__="process_virtual_memory_bytes"}`},
		// Matchers: []string{`{__name__="up"}`},
		LogLevel: "info",
		Interval: "2s",
		Workers:  4,
	})

	testutil.Ok(t, e2e.StartAndWaitReady(collector))

	time.Sleep(10 * time.Second)

	testutil.Ok(t, receive1.WaitSumMetricsWithOptions(e2emon.GreaterOrEqual(10), []string{"prometheus_tsdb_head_series"}, e2emon.WaitMissingMetrics()))

	query1 := e2edb.NewThanosQuerier(
		e,
		"query1",
		[]string{
			receive1.InternalEndpoint("grpc"),
		},
		e2edb.WithImage("quay.io/thanos/thanos:v0.35.1"),
	)
	testutil.Ok(t, e2e.StartAndWaitReady(query1))
	testutil.Ok(t, query1.WaitSumMetricsWithOptions(e2emon.Equals(1), []string{"thanos_store_nodes_grpc_connections"}, e2emon.WaitMissingMetrics()))
	testutil.Ok(t, e2einteractive.OpenInBrowser(fmt.Sprintf("http://%s", prom.Endpoint("http"))))
	testutil.Ok(t, e2einteractive.OpenInBrowser(fmt.Sprintf("http://%s", query1.Endpoint("http"))))
	testutil.Ok(t, m.OpenUserInterfaceInBrowser())
	testutil.Ok(t, e2einteractive.RunUntilEndpointHit())

}

type MetricsCollectorOptions struct {
	From     string
	To       string
	Matchers []string
	LogLevel string
	Interval string
	Workers  int
}

func NewMetricsCollector(e e2e.Environment, name string, o MetricsCollectorOptions) *e2eobs.Observable {
	f := e.Runnable(name).WithPorts(map[string]int{"metrics": 9002}).Future()

	args := e2e.BuildArgs(map[string]string{
		"--from":                     "http://" + o.From,
		"--to-upload":                "http://" + o.To + "/api/v1/receive",
		"--log-level":                o.LogLevel,
		"--interval":                 o.Interval,
		"--disable-status-reporting": "true",
		"--disable-hypershift":       "true",
		"--worker-number":            strconv.Itoa(o.Workers),
	})

	for _, matcher := range o.Matchers {
		args = append(args, "--match="+matcher)
	}

	return e2eobs.AsObservable(f.Init(wrapWithDefaults(e2e.StartOptions{
		Image:   "example.com/metrics-collector:v0.0.1",
		Command: e2e.NewCommandWithoutEntrypoint("metrics-collector", args...),
	})), "metrics")
}

var defaultBackoffConfig = backoff.Config{
	Min:        300 * time.Millisecond,
	Max:        600 * time.Millisecond,
	MaxRetries: 50,
}

func wrapWithDefaults(opt e2e.StartOptions) e2e.StartOptions {
	if opt.User == "" {
		opt.User = strconv.Itoa(os.Getuid())
	}
	if opt.WaitReadyBackoff == nil {
		opt.WaitReadyBackoff = &defaultBackoffConfig
	}
	return opt
}

```

collectors/metrics/cmd/metrics-collector/main.go
```diff
@@ -301,16 +299,36 @@ func (o *Options) Run() error {
 	// Some packages still use default Register. Replace to have those metrics.
 	prometheus.DefaultRegisterer = metricsReg
 
-	err, cfg := initConfig(o)
+	cfgRR, err := initShardedConfigs(o, AgentRecordingRule)
```
thibaultmg, line 302: There is probably some code we can extract in a specific function from `initShardedConfigs`. It mutates the options objects, doing the same processing 3 times.

collectors/metrics/cmd/metrics-collector/main.go
```diff
@@ -554,68 +537,241 @@ func initConfig(o *Options) (error, *forwarder.Config) {
...
-			if err := encoder.Encode(family); err != nil {
-				logger.Log(l, logger.Error, "msg", "unable to write metrics for family", "err", err)
-				break
+			if values[0] == "cluster" {
+				values[1] += "-" + fmt.Sprint(i)
+			}
+			if values[0] == "clusterID" {
+				values[1] = string(uuid.NewUUID())
 			}
+			opt.Labels[values[0]] = values[1]
```
thibaultmg, line 755: `opt.Labels` already exists at this stage no? It's been generated in `initShardedConfigs`. We could iterate directly on it. And why are we setting the `cluster` and `clusterID` at this stage and not in the other function?
I might be lost

saswatamcode: So this runMultiWorkers function isn't really used anywhere in main flow, this is part of the simulation testing flow. I just moved the function, so it diffs this way, but I haven't actually modified anything in this function,

https://github.com/stolostron/multicluster-observability-operator/blob/main/collectors/metrics/cmd/metrics-collector/main.go#L442

collectors/metrics/cmd/metrics-collector/main.go
saswatamcode: Yeah you're right, just didn't want to change even more things on sensitive code path

collectors/metrics/cmd/metrics-collector/main.go
saswatamcode: As for why it is the way it is, I don't really know, but I wanted to remove this simulation testing, it is mixed in with business logic and is not really a great simulation of anything. Can do in later PRs

thibaultmg (approved):
Awesome! Made some comments but no blocker.

/lgtm

saswatamcode:
/hold

Holding for a bit so others can take a look

collectors/metrics/cmd/metrics-collector/main.go
thibaultmg: The metrics collector should be replaced by the prometheusAgent. Nothing is set in stone though.

collectors/metrics/cmd/metrics-collector/main.go
thibaultmg: Ack no worries, I did not pay too much attention between what was already there, and what you added. We can leave it as is.

collectors/metrics/cmd/metrics-collector/main.go
saswatamcode: > The metrics collector should be replaced by the prometheusAgent. Nothing is set in stone though.

Yup me and @moadz discussed this, but I don't think it would be feasible yet, as you can't downsample, so would end up sending all raw 30s scrape interval data

collectors/metrics/pkg/metricsclient/metricsclient_test.go
```diff
@@ -22,7 +22,7 @@ func TestDefaultTransport(t *testing.T) {
 		TLSHandshakeTimeout: 10 * time.Second,
 		DisableKeepAlives:   true,
 	}
-	http := DefaultTransport(logger, true)
+	http := DefaultTransport(logger)
 	if http.Dial == nil || reflect.TypeOf(http) != reflect.TypeOf(want) {
```
moadz, line 26: `http.Dial` is deprecated, `http.DialContext` instead

collectors/metrics/pkg/status/status.go
```diff
@@ -93,3 +69,9 @@ func (s *StatusReport) UpdateStatus(ctx context.Context, reason status.Reason, m
 
 	return nil
 }
+
+type NoopReporter struct{}
```
moadz, line 73: `forwarder.go` we use noop reporter by default - why the need for a status reporter at all there?

operators/multiclusterobservability/api/shared/multiclusterobservability_shared.go
```diff
@@ -49,6 +49,14 @@ type ObservabilityAddonSpec struct {
 	// +kubebuilder:default:=1073741824
 	ScrapeSizeLimitBytes int `json:"scrapeSizeLimitBytes,omitempty"`
 
+	// Workers is the number of workers that work parallelly to push metrics to hub server.
+	// If set to > 1, metrics-collector will shard /federate calls to Prometheus, based on
+	// matcher rules provided by allowlist.
+	// +optional
+	// +kubebuilder:default:=1
+	// +kubebuilder:validation:Minimum=1
+	Workers int32 `json:"workers,omitempty"`
```
moadz, line 58: nit: Is it clear on the addon spec that workers corresponds to prometheus federate workers and not something else?

moadz (approved):
LGTM - awesome work Saswata! 

I don't, however, think it's appropriate to merge this day before code-freeze. Feels like there is substantial testing risk associated with this change and I wouldn't want to compromise the GA date. We can discuss this further in sync/offline.

collectors/metrics/cmd/metrics-collector/main.go
```diff
@@ -320,36 +338,31 @@ func (o *Options) Run() error {
...
-					if err := worker.Reconfigure(*cfg); err != nil {
-						logger.Log(o.Logger, logger.Error, "msg", "failed to reload config", "err", err)
-						return err
-					}
-				case <-cancel:
-					return nil
-				}
+			for i, shardWorker := range shardWorkers {
+				go func(i int, shardWorker *forwarder.Worker) {
+					fmt.Printf("Starting shard worker %d\n", i)
```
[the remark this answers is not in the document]
saswatamcode: Ah yes missed that one

collectors/metrics/cmd/metrics-collector/main.go
```diff
@@ -379,25 +388,26 @@ func (o *Options) Run() error {
...
+				err := s.Shutdown(ctx)
 				if err != nil {
 					logger.Log(o.Logger, logger.Error, "msg", "failed to close listener", "err", err)
 				}
 			})
 		}
 	}
 
-	err = runMultiWorkers(o, cfg)
+	// Run the simulation agent.
```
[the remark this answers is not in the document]
saswatamcode: This is perfscale simulations I believe, nothing I added, it existed originally. 
https://github.com/stolostron/multicluster-observability-operator/tree/main/tools/simulator/metrics-collector

We really should drop this

collectors/metrics/cmd/metrics-collector/main.go
```diff
@@ -409,68 +419,40 @@ func (o *Options) Run() error {
...
 
-func initConfig(o *Options) (error, *forwarder.Config) {
+// Agent is the type of the worker agent that will be running.
+// They are classed according to what they collect.
+type Agent string
+
+const (
+	AgentCollectRule      Agent = "collectrule"
+	AgentRecordingRule    Agent = "recordingrule"
+	AgentShardedForwarder Agent = "forwarder"
```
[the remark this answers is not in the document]
saswatamcode: Not exactly afaik.
Collectrules are rules that when evaluated and exceed a particular threshold, we just add a matcher to collect new metric
Recording Rules are actual recording rules that are evaluated in metrics-collector and sent as new series.

collectors/metrics/pkg/forwarder/forwarder.go
```diff
@@ -337,32 +395,34 @@ func (w *Worker) Reconfigure(cfg Config) error {
...
 
+	for {
 		select {
 		// If the context is canceled, then we're done.
 		case <-ctx.Done():
+			ticker.Stop()
 			return
-		case <-time.After(wait):
+		case <-ticker.C:
+			if err := w.forward(ctx); err != nil {
```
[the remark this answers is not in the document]
saswatamcode: I think this should be exponential backoff retry not quite what this is.
Right now, if a request fails, regardless of interval you set, it will just wait a full minute and try again.

collectors/metrics/pkg/forwarder/forwarder_test.go
```diff
@@ -111,38 +120,53 @@ func TestNew(t *testing.T) {
...
+				Logger: log.NewNopLogger(),
 			},
 			err: false,
 		},
 	}
 
 	for i := range tc {
 		tc[i].c.Metrics = NewWorkerMetrics(prometheus.NewRegistry())
-		if tc[i].c.ToUploadCA == customCA {
+		if i == 10 {
```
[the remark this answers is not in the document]
saswatamcode: This was set in kind of wrong way tbh, but got handled using a else branch that checks for the UNIT_TEST var directly in business logic, which is even more fishy. I don't think we are passing it as upload CA exactly.

I can instead just edit test struct instead

collectors/metrics/pkg/metricsclient/metricsclient_test.go
saswatamcode: Ack, will update, this was old code

collectors/metrics/pkg/status/status.go
saswatamcode: Forwarder is in charge of federate and remote writing that federate. So that is where we report status.

Earlier it always reported status using kubeAPI, but for local testing might not need that. So I made the status reporting kind of optional.

operators/multiclusterobservability/api/shared/multiclusterobservability_shared.go
saswatamcode: I'll add a better description

moadz (approved):
LGTM pending commnets and the hold

collectors/metrics/cmd/metrics-collector/main.go
saswatamcode: Ah yeah in that case it is, but still promagent would end up doing single large /federate call

collectors/metrics/pkg/forwarder/forwarder_test.go
```diff
@@ -214,75 +246,3 @@ func TestReconfigure(t *testing.T) {
 		}
 	}
 }
-
```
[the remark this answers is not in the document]
saswatamcode: This is the feature we dropped basically which is worker reconfigure. This test was testing exactly that

philipgough (approved):
/lgtm

PR #1629, opened 2024-09-15
Use prom port 9092 in metrics-collector to skip going through oauth-proxy

saswatamcode:
cc: @philipgough as we talked about last day
Let's check if this works!


saswatamcode:
From 4.16 onwards oauth-proxy is actually removed and replaced with kube-rbac-proxy, so this PR is no longer needed. The current code should work with 4.15 & 4.16

PR #1630, opened 2024-09-17
Grafana: Disable unified alerting

jacobbaungard:
This commit disables Grafanas unified alerting feature. We do this since it is not applicable for MCO, and since the upgrade to Grafana 11.1.5, we can disable it without errors.


thibaultmg (approved):
/lgtm

PR #1634, opened 2024-09-18
fix: [ACM-14189] identify healthy managed clusters during e2e

subbarao-meduri:
This is a test setup error.  The PR addresses two issues:

1. ~~Must exclude `local-cluster` from managed cluster list. This is because starting with ACM 2.10, observability is no longer running as a addon on the hub cluster, and always configured on the hub cluster. If a cluster is named "local-cluster" or has a label "local-cluster: true", it should be treated as a hub cluster.~~ Do not skip `local-cluster`.

2. Must exclude unhealthy clusters. From time-to-time, we have seen CICD environment with incomplete/unhealthy clusters in the hub, and the logic should skip testing Obs features on them. This is achieved by checking `ManagedClusterConditionAvailable` with `status="True"` in the managed cluster object.

### DIFF
--- a/tests/pkg/utils/mco_managedcluster.go
+++ b/tests/pkg/utils/mco_managedcluster.go
@@ -52,16 +52,10 @@ func ListManagedClusters(opt TestOptions) ([]string, error) {
 	for _, obj := range objs.Items {
 		metadata := obj.Object["metadata"].(map[string]interface{})
 		name := metadata["name"].(string)
-		labels := metadata["labels"].(map[string]interface{})
 
 		if os.Getenv("IS_KIND_ENV") == "true" {
 			// We do not have the obs add on label added in kind cluster
 			clusterNames = append(clusterNames, name)
-			continue
-		}
-
-		// Skip "local-cluster" by name or label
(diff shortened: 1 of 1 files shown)


coleenquadros (approved):

coleenquadros:
We only want to avoid tests on local cluster that are related to the observability addon. The other tests still can run on local cluster noh? We already blocked the local cluster from running obs addon tests.

subbarao-meduri:
> We only want to avoid tests on local cluster that are related to the observability addon. The other tests still can run on local cluster noh? We already blocked the local cluster from running obs addon and manifestwork tests.
> 
> If we skip local-cluster then no tests will run on it

@coleenquadros  I see your point. I did another update to the PR that does not skip `local-cluster`. 
It seems to pass the test in my local run, however I am seeing some weirdness with ManagedClusterAddon `observability-controller` getting into a `unknown state`, something we want to track separately from this PR. 
I opened a [slack thread](https://redhat-internal.slack.com/archives/CTZTHFQRH/p1726773608210059) in #forum-mce for this issue.

coleenquadros (approved):

PR #1637, opened 2024-09-24
Make sure local cluster is always in the managedcluster list

coleenquadros:
- Make sure local-cluster is always a part of  internal `managedcluster` list and not depend on the CRUD events of the Managedcluster  resource because when `disableHubSelfManagement: true`, the resource for local-cluster will not be created. 
- Use sync.Map for thread safe read and writes
- fix unit tests to run predicate tests against managedcluster resource and not deployment


subbarao-meduri (approved):
/lgtm

PR #1639, opened 2024-09-26
[ACM-14487] quote json variable so $variables are not expanded during shell echo

subbarao-meduri:
Quote shell variables so strings with `$variable` are not expanded during shell `echo`

### DIFF
--- a/tools/generate-dashboard-configmap-yaml.sh
+++ b/tools/generate-dashboard-configmap-yaml.sh
@@ -98,9 +98,9 @@ start() {
   fi
 
   # delete dashboard uid avoid conflict with old dashboard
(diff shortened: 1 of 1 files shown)


subbarao-meduri:
@saswatamcode here is how I verified the script works:
1. imported json provided in the defect
2. ran script to generate configmap yaml
3. slightly changed the name of dashboard and applied configmap
4. verified the dashboard loaded with the new name in `grafana-dev`

shane-snyder:
Just passing along, that I originally reported this issue and it seems to have fixed the behavior for me.

jacobbaungard (approved):
/lgtm

PR #1646, opened 2024-10-07
Dashboard loader: retry adding folder on errors

jacobbaungard:
Grafana sometimes fails to correctly set (default) permissions for folders added via the API. This seems to mostly affect the "Custom" folder. When this happens, while the folder is added, it's not possible for ACM users to view the folder, and its dashboards. This commit tries to detect that case, and retry adding the folder if it fails.

Additional minor changes:
- Configure grafana to retry queries
- Log messages from the component all starts in lower-case.


philipgough (approved):
/lgtm

jacobbaungard:
@moadz you want to contest the forceful exit?

I guess it's a matter of whether we want this to silently fail or not. With this approach there's a chance Grafana availability will be impacted until at least one pod has succeeded (Grafana replicas=2 by default), but at least it's clear something is wrong. Alternatively, we fail more silently and dashboards will just be missing (in this case maybe we want a bit more retries, but it's already pretty high as is).

moadz:
So my main question with forcing a crash is there is likely no intervention from a user that will fix this issue. And restarting the pod actually increases the likelihood this will fail again upon creation. So materially it's the same effect  (we keep retrying whether through crashloop or otherwise) and if it doesn't work they will have to create a support ticket to review. 

Restarting at least gives them a culprit when the dashboard doesn't show I suppose. But dashboard loader restarting grafana feels sketchy to me, so i would er on the side of just constantly retrying. If it doesn't work after N number of retries it's a big problem we will need to investigate regardless (i'm hoping the likelihood is remote).

moadz (approved):
LGTM % the restart discussion

loaders/dashboards/pkg/controller/dashboard_controller.go
```diff
@@ -40,24 +41,25 @@ const (
 var (
 	grafanaURI = "http://127.0.0.1:3001"
 	// Retry on errors.
-	retry = 10
+	httpRetry      = 10
+	dashboardRetry = 25
```
moadz, line 45: nit 
```suggestion
	maxDashboardRetry = 25
```

loaders/dashboards/pkg/controller/dashboard_controller.go
```diff
@@ -40,24 +41,25 @@ const (
 var (
 	grafanaURI = "http://127.0.0.1:3001"
 	// Retry on errors.
-	retry = 10
+	httpRetry      = 10
```
moadz, line 44: nit:
```suggestion
	maxHttpRetry      = 10
```

jacobbaungard:
> But dashboard loader restarting grafana

It doesn't actually restart the Grafana container (just the loader sidecar) but it will stop the service from directing requests to the failing pod if it's in crashloop backoff. The two Grafana replicas we have by default, don't use the same sqllite database. Not crashing the pod, could lead to some weird behavior where sometimes the dashboard show, sometimes it doesn't depending on which pod your request is routed to - Not a big fan of this this undetermism/hidden failures that this could cause.

However, I do agree that the impact of hitting this error, is potentially much larger with failing the pod, since if by bad luck we don't manage to ever actually create the dashboard, Grafana will not be available at all, whereas if we just silently fail, the worst that can happen is a few dashboards are missing. So maybe it is a little "safer" to not crash the pod, so the impact is minimal (albeit potentially more confusing for customers imo).

moadz (approved):
/lgtm

PR #1649, opened 2024-10-09
Add missing metrics to allowlist

sradco:
Added missing metrics to allowlist
for OpenShift Virtualization dashboards.

Signed-off-by: Shirly Radco <sradco@redhat.com>


sradco:
@moadz Hi, Our QE found that some panels were not showing data and I see we are missing some metrics in the allowlist unfortunately. Please review.

moadz (approved):
/lgtm

sradco:
@moadz Hi, I see this is still jot merged. It's important since without it we will not be able to view some of the panels.

PR #1652, opened 2024-10-24
Fix spec comparison for observatorium

coleenquadros:
https://issues.redhat.com/browse/ACM-14867
https://issues.redhat.com/browse/ACM-15050

Removed global var to check for custom rules config map as that could lead to a race condition to incorrectly intrepret the addition/removal of custom rules


saswatamcode (approved):
/lgtm

philipgough (approved):
/lgtm

PR #1659, opened 2024-11-07
ACM-14962: init MCOA metrics

thibaultmg:
This PR initialises metrics collection for MCOA. More precisely, it:

- Updates MCO CRD to support the metrics capability for both platform and user workload.
- Adds customised variables to the addon configuration.
- Adds permissions to MCOA for handling metrics resources.
- Generates only the configs for the enabled signals to avoid having to update it for getting a valid addon status

Relates to the MCOA PR: https://github.com/stolostron/multicluster-observability-addon/pull/77

Notes:

- Referenced default configurations for metrics are currently generated by the addon manager. It might make for sense to migrate it in MCO in a subsequent PR.
- The addon status is currently unhealthy because of the current HealthCHeck implementation on the addon manager which will soon get resolved (cf https://github.com/stolostron/multicluster-observability-addon/pull/94)


operators/multiclusterobservability/pkg/rendering/renderer_mcoa.go
```diff
@@ -212,6 +341,14 @@ func (r *MCORenderer) renderAddonDeploymentConfig(
 			}
 		}
 
+		if (cs.Platform != nil && cs.Platform.Metrics.Collection.Enabled) || (cs.UserWorkloads != nil && cs.UserWorkloads.Metrics.Collection.Enabled) {
+			obsAPIURL, err := mcoconfig.GetObsAPIExternalURL(context.TODO(), r.kubeClient, namespace)
```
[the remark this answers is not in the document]
JoaoBraveCoding: On the value itself I'm still not sure about the use of the key `signalsHubEndpoint` to represent it, mainly because this is still a very metrics oriented value. I'm thinking if we could split it so that on the short term would be more useful to more signals.

Taking the following two endpoints on the hub: 
- `https://obs-api-open-cluster-management-observability.apps.myhubcom/api/metrics/v1`
- `https://mcoa-managed-instance-openshift-logging.apps.myhub.foo.com/api/logs/v1/cluster-1/otlp/v1/logs`

Then:
- `hubHostname` (would be there if any Managed Storage is enabled): `apps.myhub.foo` or `myhub.foo`
- `hubMetricsRoute`: `obs-api-open-cluster-management-observability`

WDYT? I'm suggesting something like this because then logs&traces would be able to re-use `hubHostname`. Regardless if we do unified OBS Api or separated ones. (This could also be addressed in a follow up)

operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go
```diff
@@ -68,6 +68,16 @@ type PlatformLogsCollectionSpec struct {
...
 
+// PlatformMetricsCollectionSpec defines the spec for the addon to collect and forward metrics
+// from fleet managed clusters.
+type PlatformMetricsCollectionSpec struct {
+	// Enabled defines a flag to enable/disable the platform metrics collection.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	Enabled bool `json:"enabled,omitempty"`
+}
```
JoaoBraveCoding, line 79: In this initial version since metrics will be taking a managed approach to metric collection I wonder if these values are the best ones to use since the equivalent for logs & traces currently correlates to an unmanaged approach where the user fully controls the root resources. I would be in favor of deliberating a bit how we want these different scenarios to be exposed to users. This correlates with  https://github.com/stolostron/multicluster-observability-addon/pull/88#issue-2658695381 maybe a discussion for the nexus syncs

JoaoBraveCoding (commented):
Some thoughts

operators/multiclusterobservability/pkg/rendering/renderer_mcoa.go
```diff
@@ -164,6 +170,121 @@ func (r *MCORenderer) renderClusterManagementAddOn(
...
+				{
+					ConfigGroupResource: addonapiv1alpha1.ConfigGroupResource{
+						Group:    prometheusv1.SchemeGroupVersion.Group,
+						Resource: prometheusv1.PrometheusRuleName,
+					},
+					ConfigReferent: addonapiv1alpha1.ConfigReferent{
+						Name:      "platform-rules-default",
+						Namespace: mcoconfig.GetDefaultNamespace(),
+					},
+				},
```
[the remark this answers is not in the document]
thibaultmg: My aim is to have it working out of the box for the base case. It lists here the metrics we need to collect for having the platform dashboards working. What are you thinking about?

operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go
thibaultmg: You mean that you expect different fields in the MCO CRD to have either managed or unmanaged deployments?
It would be very interesting to discuss those concepts in the next nexus sync. 
On the metrics side, I am not sure there is demand for unmanaged type of deployments 🤷 . The expressed need is to replicate what we have now. We need something in between with some managed fields and some that we can leave it to the user for fine tuning.
Let's discuss all that in the nexus sync 👍

operators/multiclusterobservability/pkg/rendering/renderer_mcoa.go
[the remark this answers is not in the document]
thibaultmg: The issue is that we must let the user the ability to tune this list himself. As of now, he must be able to remove metrics from this list. However, the need for most or our users is to collect more metrics.
This is something I must discuss again with our PM. Maybe this will change later on.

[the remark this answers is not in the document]
thibaultmg: I could probably remove the rules more easily however. I'll think about it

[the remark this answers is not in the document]
JoaoBraveCoding: If there I a need I'm not opposed to having them here I was just double checking 👍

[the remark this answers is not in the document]
thibaultmg: I'm leaving it like this for now. I might remove some of them after reviewing the list of needed metrics, in another PR.

operators/multiclusterobservability/pkg/rendering/renderer_mcoa_test.go
```diff
@@ -156,7 +169,20 @@ func TestRenderAddonDeploymentConfig(t *testing.T) {
...
+	fakeClient := fake.NewClientBuilder().
+		WithScheme(scheme).
+		WithObjects(mco).
+		Build()
+
+	renderer := &MCORenderer{cr: mco, kubeClient: fakeClient, rendererOptions: &RendererOptions{
+		MCOAOptions: MCOARendererOptions{
+			MetricsHubHostname: "observability-hub",
+		},
+	}}
```
JoaoBraveCoding, line 185: We can simplify here
```suggestion
	renderer := &MCORenderer{cr: mco, rendererOptions: &RendererOptions{
		MCOAOptions: MCOARendererOptions{
			MetricsHubHostname: "observability-hub",
		},
	}}
```

operators/multiclusterobservability/pkg/rendering/renderer_mcoa_test.go
```diff
@@ -169,12 +195,214 @@ func TestRenderAddonDeploymentConfig(t *testing.T) {
...
+					},
+				},
+			}
+			// to kustomize resource
+			baseCMAUnstructured, err := runtime.DefaultUnstructuredConverter.ToUnstructured(baseCMA)
+			assert.NoError(t, err)
+			kres := &kustomizeres.Resource{}
+			err = runtime.DefaultUnstructuredConverter.FromUnstructured(baseCMAUnstructured, kres)
+			assert.NoError(t, err)
+
```
JoaoBraveCoding, line 380: Why do it this way VS the way we did it for `TestRenderAddonDeploymentConfig`?

operators/multiclusterobservability/pkg/rendering/renderer_mcoa_test.go
```diff
@@ -169,12 +195,214 @@ func TestRenderAddonDeploymentConfig(t *testing.T) {
...
+			// Check duplicated supportedConfigs
+			dups := make(map[string]struct{})
+			for _, cfg := range cma.Spec.SupportedConfigs {
+				key := cfg.ConfigGroupResource.Group + "/" + cfg.ConfigGroupResource.Resource
+				if _, ok := dups[key]; ok {
+					t.Errorf("duplicated supportedConfigs %s", key)
+				}
+				dups[key] = struct{}{}
+			}
```
JoaoBraveCoding, line 403: Why are we checking for duplicates? This is actually something that is [supported by the addon-framework](https://github.com/open-cluster-management-io/addon-framework/issues/263)

JoaoBraveCoding (commented):
Disregarding the discussion how the capabilities section will look like, I still have to manually test it but apart from that I would put it in lgtm territory

operators/multiclusterobservability/pkg/rendering/renderer_mcoa_test.go
thibaultmg: What do you mean? I'm testing the ClusterManagementAddOn, not the AddOnDeploymentConfig.

operators/multiclusterobservability/pkg/rendering/renderer_mcoa_test.go
thibaultmg: The addon framework will fail at runtime, I'm just checking it earlier at unit-test time.

JoaoBraveCoding (approved):
/lgtm

thibaultmg:
Force merging as e2e-kind is failing for reasons unrelated to this PR (a snapshot of a dependancy is missing).

PR #1662, opened 2024-11-07
(2.12) Use multi-arch origin-oauth-proxy image

jacobbaungard:
During upgrade from ACM2.11->ACM2.12 for a (currently not fully understood reason) we use the proxy image from the base templates. This seem to happen only for a short while, on later reconciles we appear get the correct image from the OCP imagestream.

The image previously set in the base template were amd64 only, causing the pods to crash on any other architecture. From rbac-query-proxy and Grafana it eventually resolves, however alertmanager uses a stateful set and gets stuck due the known Kubernetes issue described here: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#forced-rollback

Ideally we'd never want to use these images, but meanwhile this should at least ensure we don't crash and leave a unhealthy stateful set for alertmanager.


thibaultmg (approved):
/lgtm

PR #1664, opened 2024-11-07
fix status update and cm predicate

thibaultmg:
This PR:
- Introduces a new predicate for the configmap, only triggering reconciles when the data changes.
- Only updates the status if the collector deployment was actually updated.

https://issues.redhat.com/browse/ACM-15458


jacobbaungard (approved):
/lgtm

PR #1668, opened 2024-11-12
Use semantic.DeepEqual for Kubernetes Obj comparison

coleenquadros:
https://issues.redhat.com/browse/ACM-13688


thibaultmg:
This is a risky change to validate so close to the new release. I've had issues in the past where using deepEqual could trigger indefinite reconciliations because some default fields were set on the runtime object while the desired object would not have them.

PR #1670, opened 2024-11-15
Fix status update regression

thibaultmg:
Fixing invalid condition check for updating the status when the metrics collector deployment is updated. The isUpdated condition must be nested within the isEnabled block instead of checking for a AND condition. Because otherwise it gets into the ELSE that notifies for a disabled status.


jacobbaungard (approved):
/lgtm

saswatamcode (approved):
/lgtm

thibaultmg:
Merging without a green e2e, it's too flaky.

PR #1673, opened 2024-11-18
doc: clarify users must login to grafana-dev console before swithching 

subbarao-meduri:
Update README doc to make the login step explicit.


jacobbaungard (approved):
/lgtm

PR #1674, opened 2024-11-19
Tests: Increase custom dashboard test timeout

jacobbaungard:
The retrying of adding custom dashboards lasts at least 7 minutes. So increase timeout slightly in the hopes that the success rate is of test passings are a bit higher untill we get a better solution in place.


philipgough (approved):
/lgtm

PR #1669, opened 2024-11-14
Render: fail if no oauth image is found

jacobbaungard:
For some unknown reason, during upgrades from ACM 2.11, the template image is briefly used, instead of the image from the openshift imagestream. The template image is not available in disconnected environments, and as a result causes the pods be unable to come up.

This is a problem especially for alertmanager. Because of it being a statefulset when it gets into the bad state, with the wrong image, it doesn't automatically recover on the next reconcile. This requires manual intervention to fix.

Instead with this PR, we make the reconcile fail if we do not find the oauth image. This will make it retry later, when the imagestream is able to be found.

PR #1677, opened 2024-11-22
[main] Render: fail if no oauth image is found+kind fix

jacobbaungard:
cherry-pick of https://github.com/stolostron/multicluster-observability-operator/pull/1669 +

Renderer: detect if imagestream is available
Only try to get the oauth image from the OCP imagestream, if the
image.openshift.io api type is available. Otherwise, it likely means
we're running a non-ocp system (for kind tests for example) and in this
case we accept using the base template image.


philipgough (approved):
/lgtm

PR #1698, opened 2024-12-04
remove 2.12 konflux files

coleenquadros:
These are files that reflect the n-1 release in the main branch. We have ACM 2.13 files merged so we don't need 2.12 in main


jacobbaungard (approved):
/lgtm

PR #1700, opened 2024-12-06
[ACM-13481] - Remove HA mode when AM has one replica

coleenquadros:
https://issues.redhat.com/browse/ACM-13481

This is a workaround/hack when AM replicas are set to 1 and are no more in HA mode. https://github.com/prometheus/alertmanager/pull/1971


philipgough (commented):
/lgtm

PR #1718, opened 2024-12-10
[ACM-16286]: Ensure MCO spec addon options are propagated correctly

saswatamcode:
Turns out there are a few issues with the behavior of the ObservabilityAddon CRD.

The CRD should operationally be doing the following,
- Have a sane default for the options, all of which it will inherit from the MCO obsAddon spec at the start
- Be per-cluster configurable, so that the user can modify the ObservabilityAddon CR in a specific spoke cluster namespace on the hub
- Get copied over onto the spoke cluster and configure endpoint-operator
- Allow user to switch back to MCO-based ObsservabilityAddon config if they wish to 
- All these changes are in sync in three places: ManifestWork on hub, spoke ns ObservabilityAddon CR on hub, ObservabilityAddon CR in the actual spoke cluster.

However it was not doing this correctly which led to a few issues and fields like the scrape limit or workers for metrics-collector just wouldn't work as well as couldn't be customized. So this PR addresses those issues. This PR,

- Updates all copies of the relevant ObservabilityAddon CRD in the repo (in both mco and endpoint-operator), to include all of the options that have been added to it
- Make sure Placementrule Controller reconciles the ObservabilityAddon CR between spoke ns and ManifestWork correctly on changes to fields and annotations as well as status changes.
- Add an annotation to the ObservabilityAddon CR `observability.open-cluster-management.io/addon-source` during the creation of the ObservabilityAddon CR and maintain it for the lifecycle. It will have two accepted values "mco" and "override". When set to "mco" it will always reflect spec from MCO CR. If you try to change it, it will reconcile back to those values. When set to "override", we will not update the ObservabilityAddon CRD and will only use found values. This basically means user can set it to whatever they want, for a particular cluster
- If user deletes ObservabilityAddon CR, it will revert back to default with "mco" addon-source again

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
+++ b/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
@@ -977,6 +977,11 @@ func getObservabilityAddon(c client.Client, namespace string,
 		},
 	}
 
+	// Handle cases where the addon doesn't have the annotation
+	if _, ok := found.Annotations["observability.open-cluster-management.io/addon-source"]; !ok {
+		found.Annotations["observability.open-cluster-management.io/addon-source"] = "mco"
+	}
+
 	if found.Annotations["observability.open-cluster-management.io/addon-source"] == "mco" {
 		addon.Spec.EnableMetrics = mco.Spec.ObservabilityAddonSpec.EnableMetrics
 		addon.Spec.Interval = mco.Spec.ObservabilityAddonSpec.Interval
--- a/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
+++ b/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
@@ -971,9 +971,11 @@ func getObservabilityAddon(c client.Client, namespace string,
 			Namespace: spokeNameSpace,
 		},
 		Spec: mcoshared.ObservabilityAddonSpec{
-			EnableMetrics: mco.Spec.ObservabilityAddonSpec.EnableMetrics,
-			Interval:      mco.Spec.ObservabilityAddonSpec.Interval,
-			Resources:     config.GetOBAResources(mco.Spec.ObservabilityAddonSpec, mco.Spec.InstanceSize),
+			EnableMetrics:        mco.Spec.ObservabilityAddonSpec.EnableMetrics,
+			Interval:             mco.Spec.ObservabilityAddonSpec.Interval,
+			ScrapeSizeLimitBytes: mco.Spec.ObservabilityAddonSpec.ScrapeSizeLimitBytes,
+			Workers:              mco.Spec.ObservabilityAddonSpec.Workers,
+			Resources:            config.GetOBAResources(mco.Spec.ObservabilityAddonSpec, mco.Spec.InstanceSize),
 		},
 	}, nil
 }
--- a/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
+++ b/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
@@ -941,10 +941,10 @@ func generateMetricsListCM(client client.Client) (*corev1.ConfigMap, *corev1.Con
 
 // getObservabilityAddon gets the ObservabilityAddon in the spoke namespace in the hub cluster.
 // This is then synced to the actual spoke, by injecting it into the manifestwork.
-// It will set the MCO spec by default, but if the existing spoke namespace ObservabilityAddon exists,
-// it will use that.
-// If the found ObservabilityAddon exists with the default values, it will always prefer the MCO spec.
-// However, if the found ObservabilityAddon exists with the non-default values, it will use the found values, even if the MCO spec is different.
+// We assume that an existing addon will always be found here as we create it initially.
+// If the addon is found with the mco source annotation, it will update the existing addon with the new values from MCO
+// If the addon is found with the override source annotation, it will not update the existing addon but it will use the existing values.
+// If the addon is found without any source annotation, it will add the mco source annotation and use the MCO values (upgrade case from ACM 2.12.2).
 func getObservabilityAddon(c client.Client, namespace string,
 	mco *mcov1beta2.MultiClusterObservability) (*mcov1beta1.ObservabilityAddon, error) {
 	if namespace == config.GetDefaultNamespace() {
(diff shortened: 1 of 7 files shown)


philipgough (approved):
/lgtm

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -952,30 +957,57 @@ func getObservabilityAddon(c client.Client, namespace string,
...
-	}, nil
+	}
+
+	// Handle cases where the addon doesn't have the annotation
+	if found.Annotations == nil {
+		found.Annotations = make(map[string]string)
+	}
+
+	if _, ok := found.Annotations[addonSourceAnnotation]; !ok {
+		found.Annotations[addonSourceAnnotation] = "mco"
```
philipgough, line 987: might be worth making these values a constant somewhere? `mco, override `

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -952,30 +957,57 @@ func getObservabilityAddon(c client.Client, namespace string,
...
+	}
+
+	if found.Annotations[addonSourceAnnotation] == "mco" {
+		addon.Spec.EnableMetrics = mco.Spec.ObservabilityAddonSpec.EnableMetrics
+		addon.Spec.Interval = mco.Spec.ObservabilityAddonSpec.Interval
+		addon.Spec.ScrapeSizeLimitBytes = mco.Spec.ObservabilityAddonSpec.ScrapeSizeLimitBytes
+		addon.Spec.Workers = mco.Spec.ObservabilityAddonSpec.Workers
+		addon.Spec.Resources = config.GetOBAResources(mco.Spec.ObservabilityAddonSpec, mco.Spec.InstanceSize)
+
+		addon.Annotations[addonSourceAnnotation] = "mco"
```
philipgough, line 997: why do we need to set this here again?

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -952,30 +957,57 @@ func getObservabilityAddon(c client.Client, namespace string,
...
+	}
+
+	if found.Annotations[addonSourceAnnotation] == "override" {
+		addon.Spec.EnableMetrics = found.Spec.EnableMetrics
+		addon.Spec.Interval = found.Spec.Interval
+		addon.Spec.ScrapeSizeLimitBytes = found.Spec.ScrapeSizeLimitBytes
+		addon.Spec.Workers = found.Spec.Workers
+		addon.Spec.Resources = found.Spec.Resources
+
+		addon.Annotations[addonSourceAnnotation] = "override"
```
philipgough, line 1007: same comment as above? we wouldnt fall in here unless this k,v was already set?

operators/multiclusterobservability/controllers/placementrule/obsaddon.go
```diff
@@ -70,11 +76,22 @@ func createObsAddon(c client.Client, namespace string) error {
...
 				ownerLabelKey: ownerLabelValue,
 			},
 		},
 	}
+
+	if mco.Spec.ObservabilityAddonSpec != nil {
+		ec.Spec.EnableMetrics = mco.Spec.ObservabilityAddonSpec.EnableMetrics
+		ec.Spec.Interval = mco.Spec.ObservabilityAddonSpec.Interval
+		ec.Spec.ScrapeSizeLimitBytes = mco.Spec.ObservabilityAddonSpec.ScrapeSizeLimitBytes
+		ec.Spec.Workers = mco.Spec.ObservabilityAddonSpec.Workers
```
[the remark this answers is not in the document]
philipgough: in either case, we might be best to extract the "shared setting logic" to a function and appended additional atop?

operators/multiclusterobservability/controllers/placementrule/obsaddon.go
```diff
@@ -96,6 +113,21 @@ func createObsAddon(c client.Client, namespace string) error {
 		return err
 	}
 
+	// Check if existing addon was created by MCO
+	if found.Annotations != nil && found.Annotations[addonSourceAnnotation] == "mco" {
+		// Only update if specs are different
+		if found.Spec != ec.Spec {
```
thibaultmg, line 119: I would do a deepEqual, resources is a pointer in the struct.

operators/multiclusterobservability/controllers/placementrule/obsaddon.go
```diff
@@ -96,6 +113,21 @@ func createObsAddon(c client.Client, namespace string) error {
...
+			found.Spec = ec.Spec
+			err = c.Update(context.TODO(), found)
+			if err != nil {
+				log.Error(err, "Failed to update observabilityaddon cr")
```
thibaultmg, line 123: We could just wrap the error, it is logged already in parent functions.

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -653,7 +653,11 @@ func (r *PlacementRuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				e.ObjectNew.GetLabels()[ownerLabelKey] == ownerLabelValue &&
 				e.ObjectNew.GetNamespace() != localClusterName &&
 				!reflect.DeepEqual(e.ObjectNew.(*mcov1beta1.ObservabilityAddon).Status.Conditions,
-					e.ObjectOld.(*mcov1beta1.ObservabilityAddon).Status.Conditions) {
+					e.ObjectOld.(*mcov1beta1.ObservabilityAddon).Status.Conditions) &&
+				!reflect.DeepEqual(e.ObjectNew.(*mcov1beta1.ObservabilityAddon).Spec,
```
thibaultmg, line 657: Should the whole block comparing the object spec, annotations and conditions encapsulated in a AND and contains OR?
... && (!equalSpec || !equalAnnotations || ...)

Or I don't understand something 🤷

thibaultmg (commented):
Looks good overall! It always surprises me that we discover those kind of bugs so late... Waiting for your replies

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
saswatamcode: This is a new copy, but yeah we can just copy over from found.

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
saswatamcode: Nice catch

thibaultmg (approved):
/lgtm

PR #1719, opened 2024-12-11
[ACM-16335] Reconcile MCO only when there is actual change in mch image manifest

coleenquadros:
MCO and Placement Controller react to creates/updates/deletes from MCH CR . But we only specifically care about the mch-image-manifest configmap changes. So if we continuously get mch updates, the current reconcile logic always returns true after reading the image configmap even if it didnt differ from the previous version. 

The update to mch image configmap has to reflect on all spokes and in scale environment this is going to be a very expensive operation and if we dont check if an actual update is needed, then this is going to noisy in terms of logs and also unnecessary API calls

### DIFF
--- a/operators/multiclusterobservability/controllers/multiclusterobservability/predicate_func.go
+++ b/operators/multiclusterobservability/controllers/multiclusterobservability/predicate_func.go
@@ -5,10 +5,11 @@
 package multiclusterobservability
 
 import (
+	"reflect"
+
 	mcov1beta2 "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/api/v1beta2"
 	"github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/pkg/config"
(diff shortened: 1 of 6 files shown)


jacobbaungard:
This generally makes sense to me. The only thing I wondered, was that I've seen past cases were we didn't reconcile/progress with MCO due to MCH being in a non-ok state, and I just wondered if this change might impact that logic.

Looking at the code, I don't see that this PR changes any such behavior though, so maybe this is done elsewhere?

thibaultmg (commented):
What don't we use a predicate on the configmap directly instead and remove the Update part of the MCH predicate?

coleenquadros:
> Why don't we use a predicate on the configmap directly instead and remove the Update part of the MCH predicate?

I am not aware of the complete dependency on MCH with MCO. Previously we reacted to changes on MCH CR and then just read the mch image manifest configmap and returned true and now I have blocked the reconcile from happening unless image manifest spec changed. @subbarao-meduri are you aware of any other fields or resources on the MCH that we might have to reconcile on?

PR #1733, opened 2024-12-18
Fix leaky goroutine on metrics-collector

saswatamcode:
Without this, it starts a new routine when workers are > 1 which sends requests to federate with 0 limit byte


philipgough (approved):
/lgtm

PR #1763, opened 2025-01-10
minor update of README and Makefile

tremes:
This is a minor update to the README link for installing kustomize and a minor update to the Makefile where the `undeploy` command does not seem to be correct.

I was also wondering if the mention of the `kustomize` goal in https://github.com/stolostron/multicluster-observability-operator/blob/main/operators/multiclusterobservability/Makefile should be removed, because it doesn't seem to be defined anywhere (and it's referenced in the `install`, `uninstall` and `deploy` goals)


jacobbaungard (approved):
/lgtm

PR #1777, opened 2025-01-14
Fix: Flapping addon staus due to 409 response on remote write to the hub

thibaultmg:
Status from the metrics collector now reports available state when receiving a 409 http error on remote write to the hub. 
This reflects more accurately the status of the addon as metrics are well ingested. The error is still logged to make it visible as this is most often the result of an invalid configuration.

Fixes https://issues.redhat.com/browse/ACM-17055


philipgough (approved):
/lgtm

saswatamcode (approved):
/lgtm

I would merge as is, not wait for CI

PR #1794, opened 2025-01-21
Fix grafana dashboards update when adding new folder

thibaultmg:
As per [documentation](https://www.sqlite.org/tempfiles.html#temporary_file_storage_locations), sqlite needs access to a temporary directory for some operations. As the root file system is readOnly and it picks by default the `/tmp` dir, adding a folder was failing since we upgraded Grafana.

This PR adds a tempdir env variable for sqlite that point to the same mounted directory as the database as it is writable.

This PR also includes an improvement to reduce e2e test flakiness. Previously, because `cert/time-restarted` only had minute-level accuracy, it was possible for two e2e tests to generate the same timestamp, leaving components unchanged and causing tests to fail (certrenew at least). The time accuracy is now improved, down to the second.


jacobbaungard (approved):
/lgtm

PR #1796, opened 2025-01-22
ACM-15764: fix access token renewal from the metrics collector

thibaultmg:
Introduces a new struct named TokenFile to automatically renew the token by re-reading the token file when it approaches the expiration date. The token is accessed by the RoundTripper injecting the bearer token calling the GetToken() method.

### DIFF
--- a/collectors/metrics/pkg/forwarder/token.go
+++ b/collectors/metrics/pkg/forwarder/token.go
@@ -1,3 +1,7 @@
+// Copyright (c) Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+// Licensed under the Apache License 2.0
+
 package forwarder
 
 import (
--- a/collectors/metrics/pkg/forwarder/token.go
+++ b/collectors/metrics/pkg/forwarder/token.go
@@ -27,12 +27,15 @@ var (
 type TokenFile struct {
 	filePath    string
 	logger      log.Logger
+	readBackoff time.Duration
 	token       string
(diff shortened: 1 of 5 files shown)


collectors/metrics/pkg/forwarder/token.go
```diff
@@ -0,0 +1,166 @@
...
+	t.tokenMu.RLock()
+	defer t.tokenMu.RUnlock()
+	return t.token
+}
+
+// autoRenew automatically re-read the token file to update its value when it approaches the expiration time.
+// The objective is to have a simple and robust strategy.
+// It assumes that kubernetes renews the token when it reaches 80% of its lifetime. Most lifetimes are 1y or 1h.
+// The strategy is to read the token file when we reach 85% of the remaining lifetime, and then every backoff interval
+// when the remaining time is below 4 times the read backoff duration.
```
saswatamcode, line 99: This feels a bit too complicated. How about renewing when you hit some constant margin like 5m instead?
I believe this is what we do for token-refresher https://github.com/observatorium/token-refresher/blob/master/main.go#L113

thibaultmg: By updating the token early, I assumed we are more robust to clock skews or potential zone aware time problems (even this this last one should not happen). Even if a bit more complicated, the described strategy is validated through unit tests.
But if you prefer the referenced strategy, it is fine for me. Let's wait and see if we have other opinions.

PR #1801, opened 2025-01-28
Update renovate.json to pick only main and actual release branches

coleenquadros:
We get unnecessary Konflux  updates to branches not related to the release or work we do. Eg https://github.com/stolostron/multicluster-observability-operator/pull/1799 
https://github.com/stolostron/multicluster-observability-operator/pull/1788


jacobbaungard (approved):
/lgtm

PR #1805, opened 2025-01-28
Fix e2e watchdog alert test

thibaultmg:
The watchdog alert is a Prometheus alert that is always on, and automatically deployed on openshift clusters with the in-cluster stack.

The e2e alert test, verifies that the watchdog alert is well forwarded to the hub alertmanager, validating the communication links. And as this alert is deployed everywhere, it is forwarded by alertmanagers deployed on both the hub and spokes.

The test fetches the watchdog alerts received by the ACM AlertManager, and verifies that is contains the ones from all managed clusters. However, the list of managed clusters is built by fetching the ManagedCluster resources and by checking the label `feature.open-cluster-management.io/addon-observability-controller` that is added by the [addon manager](https://github.com/open-cluster-management-io/enhancements/blob/87ef9534262ecd4d524759c1be41a25c4c61d437/enhancements/sig-architecture/17-cluster-features/README.md?plain=1#L42) is set to `available`. But the addon manager is not anymore deployed on the hub, and this label is not added to the local cluster's ManagedCluster resource.  As a result, the list does not contain the local-cluster ID.

**Why is it not always failing?**
Because the retried queries to the ACM hub alert manager are not consistently returning watchdog alerts from all managed clusters. Most of the time they contain alerts from all clusters. But sometimes none, or just one of them, making the test eventually pass. Sometimes.

This PR changes how the list of expected cluster IDs is built, to always include the local-cluster one. 
It also changes the way clusters are checked from the alermanager response by removing them as they are being found instead of expecting all of them in the same response.

### DIFF
--- a/tests/pkg/tests/observability_alert_test.go
+++ b/tests/pkg/tests/observability_alert_test.go
@@ -14,8 +14,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"reflect"
-	"sort"
+	"slices"
 	"strings"
 
 	. "github.com/onsi/ginkgo"
@@ -302,6 +301,8 @@ var _ = Describe("Observability:", func() {
 	})
 
 	It("[P2][Sev2][observability][Integration] Should have alert named Watchdog forwarded to alertmanager (alertforward/g0)", func() {
+		// Watchdog is an alert that is installed by default on OCP clusters by the in-cluster monitoring stack
+		// It thus exists by default on the hub and openshift spokes, and is always activated.
 
 		amURL := url.URL{
 			Scheme: "https",
@@ -334,11 +335,12 @@ var _ = Describe("Observability:", func() {
 			alertGetReq.Header.Set("Authorization", "Bearer "+BearerToken)
 		}
 
-		expectedOCPClusterIDs, err := utils.ListOCPManagedClusterIDs(testOptions, "4.8.0")
+		expectedOCPClusterIDs, err := utils.ListOCPManagedClusterIDs(testOptions)
 		Expect(err).NotTo(HaveOccurred())
 		expectedKSClusterNames, err := utils.ListKSManagedClusterNames(testOptions)
 		Expect(err).NotTo(HaveOccurred())
 		expectClusterIdentifiers := append(expectedOCPClusterIDs, expectedKSClusterNames...)
+		missingClusters := slices.Clone(expectClusterIdentifiers)
 
 		// install watchdog PrometheusRule to *KS clusters
 		watchDogRuleKustomizationPath := "../../../examples/alerts/watchdog_rule"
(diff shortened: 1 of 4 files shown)


jacobbaungard (commented):
Awesome job, and thanks for the very thorough PR description ❤️

The generic `ListManagedClusters` (i.e without OCP/*ks filter) contains the following. 

```
 if conditionMap["type"] == "ManagedClusterConditionAvailable" && conditionMap["status"] == "True" {
 ```
 
 We don't want to include this for the OCP/*ks ones? I though the `ListManagedClusters` one was fixed for the new local-cluster behavior?

File: tests/pkg/utils/mco_managedcluster.go [hunk 1]
```diff
@@ -154,24 +132,16 @@ func ListKSManagedClusterNames(opt TestOptions) ([]string, error) {
...
-				}
-				if len(clusterNameStr) > 0 {
-					clusterNames = append(clusterNames, clusterNameStr)
-				}
-			}
+		if labels == nil {
+			continue
+		}
+
+		if vendor, ok := labels["vendor"]; ok && vendor == "OpenShift" {
```
coleenquadros, line 139: the vendor wouldnt be Openshift for KS clusters noh

thibaultmg: yes, I am skipping them here.

thibaultmg:
> Awesome job, and thanks for the very thorough PR description ❤️
> 
> The generic `ListManagedClusters` (i.e without OCP/*ks filter) contains the following.
> 
> ```
>  if conditionMap["type"] == "ManagedClusterConditionAvailable" && conditionMap["status"] == "True" {
> ```
> 
> We don't want to include this for the OCP/*ks ones? I though the `ListManagedClusters` one was fixed for the new local-cluster behavior?

yep good idea, I'll align them.

tests/pkg/tests/observability_alert_test.go
```diff
@@ -394,10 +400,16 @@ var _ = Describe("Observability:", func() {
 				}
 			}
 
-			sort.Strings(clusterIDsInAlerts)
-			sort.Strings(expectClusterIdentifiers)
-			if !reflect.DeepEqual(clusterIDsInAlerts, expectClusterIdentifiers) {
-				return fmt.Errorf("Not all openshift managedclusters >=4.8.0 forward Watchdog alert to hub cluster")
+			// Returned alerts by the alert manager is not consistent. It does not always contain all the alerts.
+			// To make the test more reliable, we romove clusters found in the response continuously on each retry
```
jacobbaungard, line 404:
```suggestion
			// To make the test more reliable, we remove clusters found in the response continuously on each retry
```

jacobbaungard (approved):
/retest

PR #1806, opened 2025-01-29
OBSINTA-203: Grafana dashboard for Incidents overview

tremes:
This is ading a new Grafana dashboard for the Incident detection feature.I added two following metrics to be scraped and stored in Thanos:

- `cluster:health:components:map` - incidents data
- `console_url` - to have the Console URL of spoke clusters (so that we can link to the Incident detail in the future)


thibaultmg:
Could you please also add this dashboard to platform/mcoa along with the needed metrics? 
Unless you don't want them when MCOA is enabled?

tremes:
@thibaultmg do you mean to move the dashboard definition to `platform-mcoa` folder and the metrics then need to be in the [scrape-config.yaml](https://github.com/stolostron/multicluster-observability-operator/blob/main/operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/scrape-config.yaml) ?

thibaultmg:
I mean duplicate. For context, the platform dashboards are marked ass deprecated and replaced by platform-mcoa that contains optimised metrics for MCOA without breaking the existing dashboards. So it depends on your objective but I think it would be nice to have it in both directories.
For the MCOA ones, you need to update the scrape-config indeed. There is a special CI check that will scream if metrics are missing.

tremes:
@thibaultmg Thank you. This context is really helpful! I was not aware about it. I agree and I am going to update it.

thibaultmg (approved):
/lgtm

PR #1822, opened 2025-02-04
Fix Hermetic builds and Enterprise contract violations

coleenquadros:
Fix Hermetic builds and Enterprise contract violations


subbarao-meduri (approved):
/lgtm

PR #1849, opened 2025-02-18
[ACM-17437] Verify overwrite Thanos components CLI flags

coleenquadros:
https://issues.redhat.com/browse/ACM-17437


tests/pkg/tests/observability_config_test.go
```diff
@@ -348,6 +348,79 @@ var _ = Describe("Observability:", func() {
 		}
 	})
 
+	It("RHACM4K-43019 - Observability - Verify overwrite Thanos components CLI args in MCO CR - [P2][Sev2][Observability][Integration]@ocpInterop @non-ui-post-restore @non-ui-post-release @non-ui-pre-upgrade @non-ui-post-upgrade @post-upgrade @post-restore @e2e @post-release (config/g0)", func() {
+		By("Check the value is effect in the observability-thanos-compact and rule")
```
jacobbaungard, line 352: I think we should apply the actual MCO changes directly in the test here (i.e `../../../examples/updatemcocr/advancedmcoconfig`)  so that we do not rely on the other tests that applies this config are being executed first. iirc it seems like these tests execute in order but as far as I could see from Ginkgo docs it is not guaranteed (unless specifically set).

PR #1861, opened 2025-02-24
Update  go builder - 1.23

coleenquadros:
Update CMO and go builder

### DIFF
--- a/collectors/metrics/Containerfile.operator
+++ b/collectors/metrics/Containerfile.operator
@@ -1,7 +1,7 @@
 # Copyright Contributors to the Open Cluster Management project
 # Licensed under the Apache License 2.0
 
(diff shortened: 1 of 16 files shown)


jacobbaungard (commented):
Shouldn't we update the rest of the images too, such as https://github.com/stolostron/multicluster-observability-operator/blob/main/operators/multiclusterobservability/Dockerfile so that we consistently use 1.23 everywhere ?

coleenquadros:
> Shouldn't we update the rest of the images too, such as https://github.com/stolostron/multicluster-observability-operator/blob/main/operators/multiclusterobservability/Dockerfile so that we consistently use 1.23 everywhere ?

I will take care of that when the go.mod gets updated to 1.23. For now this only Konflux references.

jacobbaungard (approved):
/lgtm

PR #1881, opened 2025-02-27
Kind: do not copy incompatible binaries to VM & fallback to previous release if no snapshots found

jacobbaungard:
- Tooling changes required as we're now running later versions of ci-opertor which uses RHEL9, and the remote VM uses RHEL8 which means binaries are not compatible
- If we cannot find a snapshot for the current component version, fallback to a snapshot from the previous version.


jacobbaungard:
This should pass as soon as we have some 2.14 snapshots on quay.

coleenquadros (approved):

PR #1886, opened 2025-03-05
dashboard-loader: set home dashboard with UID

jacobbaungard:
Instead of using the name, as the same name was used for the MCOA equivalent.

Fixes: ACM-18355

/cherry-pick release-2.13


thibaultmg (approved):
/lgtm

PR #1888, opened 2025-03-05
Remove standard hub rules for MCOA

thibaultmg:
Stop relying on [rules](https://github.com/stolostron/multicluster-observability-operator/blob/bf2118fa4a5b644a9962eb63bc9a3714d40e1da6/operators/multiclusterobservability/manifests/base/alertmanager/alert_rules.yaml#L1) computed on the Hub Thanos Ruler. They are replaced by rules pushed to the spokes and computed there.
Resulting rules are added to the scrape config.
Sort alphabetically the scrape config list of metrics.


jacobbaungard (approved):
/lgtm

PR #1890, opened 2025-03-07
[ACM-15722] RBAC e2e only authorized user can query managed cluster metrics data

coleenquadros:
This PR includes the following changes 

- Remove MCO v1beta1 related tests and changes. This should probably help run/start the tests quicker.
- Replace deprecated Ginkgo fns for all test files for after spec runs
- Add rbac testing to query metrics from users with different binding roles that limit access to query metrics
- Not running the rbac test in kind because it requires oc cli to be installed and we use kind as a means of testing non-ocp clusters
- Reduced the number of replicas for thanos components when running on kind or locally for a quicker and lighter setup.


examples/mco/e2e/v1beta2/custom-certs-kind/observability.yaml
```diff
@@ -0,0 +1,127 @@
...
+      retentionInLocal: 5d
+      retentionResolution1h: 31d
+      retentionResolution5m: 15d
+      retentionResolutionRaw: 6d
+    observatoriumAPI:
+      resources:
+        limits:
+          cpu: 1
+          memory: 1Gi
+      replicas: 2
```
thibaultmg, line 20: Why did you keep some replicas to more than 1?

tests/pkg/tests/observability_rbac_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"os/exec"
+)
+
+var _ = Describe("Observability:", func() {
+	It("RHACM4K-1406 - Observability - RBAC - only authorized user could query managed cluster metrics data [Observability][Integration]@ocpInterop @non-ui-post-restore @non-ui-post-release @non-ui-pre-upgrade @non-ui-post-upgrade @post-upgrade @post-restore @e2e @post-release (requires-ocp/g0) (obs_rbac/g0)", func() {
+		By("Setting up users creation and rolebindings for RBAC", func() {
+			cmd := exec.Command("../../setup_rbac_test.sh")
+			var out bytes.Buffer
+			cmd.Stdout = &out
+			_ = cmd.Run()
```
thibaultmg, line 24: we could check this error. Like:
```
err := cmd.Run()
Expect(err).NotTo(HaveOccurred())
```

thibaultmg (approved):
/lgtm

examples/mco/e2e/v1beta2/custom-certs-kind/observability.yaml
coleenquadros, line 1: @saswatamcode This is the file for kind resource allocation

operators/multiclusterobservability/pkg/config/config.go
```diff
@@ -118,7 +118,7 @@ const (
 
 const (
 	DefaultImgRepository = "quay.io/stolostron"
-	DefaultImgTagSuffix  = "2.4.0-SNAPSHOT-2021-09-23-07-02-14"
+	DefaultImgTagSuffix  = "2.13.0-SNAPSHOT-2025-03-07-03-19-34"
```
saswatamcode, line 121: What is this tag for?

tests/pkg/tests/observability_alert_test.go
```diff
@@ -57,27 +56,18 @@ var _ = Describe("Observability:", func() {
...
+		Eventually(func() error {
+			_, err := dynClient.Resource(utils.NewPrometheusRuleGVR()).Namespace(MCO_NAMESPACE).
+				Get(context.TODO(), "acm-observability-alert-rules", metav1.GetOptions{})
+			if err != nil {
+				testFailed = true
+				return err
+			}
+			testFailed = false
+			return nil
+		}, EventuallyTimeoutMinute*5, EventuallyIntervalSecond*10).Should(Succeed())
```
saswatamcode, line 68: What is testFailed used for?

examples/mco/e2e/v1beta2/custom-certs-kind/observability.yaml
```diff
@@ -0,0 +1,127 @@
...
+          cpu: 1
+          memory: 1Gi
+      serviceAccountAnnotations:
+        test.com/role-arn: 's3_role'
+    receive:
+      resources:
+        limits:
+          cpu: 1
+          memory: 1Gi
+      replicas: 2
```
saswatamcode, line 47: Does changing this also modify replication factor?

examples/mco/e2e/v1beta2/custom-certs-kind/observability.yaml
saswatamcode: If 2 works for local, I'd keep it that way actually, good to test all multi rep features for everything

examples/mco/e2e/v1beta2/custom-certs-kind/observability.yaml
```diff
@@ -0,0 +1,127 @@
...
+      memoryLimitMb: 2048
+      maxItemSize: 2m
+      connectionLimit: 2048
+    grafana:
+      replicas: 2
+      resources:
+        limits:
+          cpu: 1
+          memory: 1Gi
```
saswatamcode, line 89: Do we have tests that check grafana? Just wondering if 1 Gi, 1 cpu is high

examples/mco/e2e/v1beta2/custom-certs-kind/observability.yaml
saswatamcode: To me most of it looks fine in general, some comments
I think you can do a total CPU, mem value by multiplying reps? If this works locally, I think we are fine anyway :)

saswatamcode (commented):
Thanks a lot for doing this, some comments!

examples/mco/e2e/v1beta2/custom-certs-kind/observability.yaml
coleenquadros: Yes, https://github.com/coleenquadros/multicluster-observability-operator/blob/78d5118a7c32809c59bc0bd399dcf97e9df2398f/operators/multiclusterobservability/controllers/multiclusterobservability/observatorium.go#L624

examples/mco/e2e/v1beta2/custom-certs-kind/observability.yaml
coleenquadros: We have those grafana dev tests. But I dont think these tests are extensive to require more resources

operators/multiclusterobservability/pkg/config/config.go
coleenquadros: If we dont end up picking the latest snapshot while running e2e, it just defaults to this.

tests/pkg/tests/observability_alert_test.go
coleenquadros: its a global var used in the e2e tests for the final reporting, shown below ->

testFailed = testFailed || CurrentSpecReport().Failed()

PR #1892, opened 2025-03-11
Add controller reference on resources deployed by the endpoint-operator

thibaultmg:
To ensure that resources created by the endpoint operator are deleted, especially on the hub, a controller reference is added to them. The owner is set to:

- The observability addon resource on the spokes
- MCO on the hub

https://issues.redhat.com/browse/ACM-18474

### DIFF
--- a/operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
+++ b/operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
@@ -294,6 +294,7 @@ func (r *ObservabilityAddonReconciler) Reconcile(ctx context.Context, req ctrl.R
 		return res, err
 	}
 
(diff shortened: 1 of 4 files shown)


jacobbaungard (approved):
/lgtm

We should probably backport at least to 2.13(.1) right? (think we should wait with the merge till final .0 build is created).

thibaultmg:
> We should probably backport at least to 2.13(.1) right? (think we should wait with the merge till final .0 build is created).

In theory we should backport but this is really a small bug. It only happens when you uninstall observability from the hub. And there are still other resources that are not cleaned... 
Ideally we should have a test ensuring that all resources created by all controllers have the owner field correctly configured so that they get garbage collected and that global resources are correctly cleaned.
I'll create a task to track that.

PR #1897, opened 2025-03-12
[ACM-17443] Check for node selector addondeployment config in endpoint operator

coleenquadros:
[ACM-17443](https://issues.redhat.com/browse/ACM-17443)

We don't need an e2e for this. We can just create/update an addondeployment config with the node selector configs and ensure that its propagated to the endpoint operator deployment when the manifestworks for the spokes are created. Testing actual node selection by the pods is not in the scope for mco testing


thibaultmg (approved):
/lgtm

PR #1899, opened 2025-03-12
Only activate cmowatcher on spokes

thibaultmg:
Currently it logs many status update errors on the hub. With no consequence as it does not impact the reconcile loop.


jacobbaungard (approved):
/lgtm

PR #1912, opened 2025-03-17
[ACM-15724] Observability - RBAC - Verify only cluster-manager-admin role can deploy MCO CR

coleenquadros:
https://issues.redhat.com/browse/ACM-15724
This test is only going to check the policy to be able to delete MCO resource with the users respective clusterrole bindings instead of actually deploying or trying to delete MCO and waiting for the deployment to come up to reduce e2e run time


thibaultmg (approved):
/lgtm

PR #1936, opened 2025-04-10
[ACM-17966] Namespace right sizing recommendation changes

rajusem:
This PR helps to enable `namespaceRightSizingRecommendation` feature with MCO. Right-Sizing Recommendation helps users identify over-provisioned or under-utilized resources across managed clusters, enabling more efficient use of infrastructure and reducing unnecessary costs.

- User can enable it by updating CRD as below. 
```
kind: MultiClusterObservability
metadata:
  name: observability
spec:
  capabilities:
    platform:
      analytics:
        namespaceRightSizingRecommendation:
          enabled: true
```
- Underlying it will create `Configmap`, `Policy`, `Placement`, `PlacementBinding` etc components.
- Policy contains PrometheusRule info for cpu/memory related data points, which will be used in grafana dashboard 
- Through configmap we are enabling users to modify placement and prometheusrule information so we have added predicate function to watch create/update events 
- We have created `operators/multiclusterobservability/controllers/multiclusterobservability/analytics/` folder and added all related controller logic inside the same.
- Added rightsizing related metrics into `metric_allowlist`
- Also added grafana dashboard under `operators/multiclusterobservability/manifests/base/grafana/analytics/` folder

Epic - https://issues.redhat.com/browse/ACM-17966

Signed-off-by: Raj Zalavadia <rzalavad@redhat.com>
Co-authored-by: Darshan Vandra <dvandra@redhat.com>


operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -398,6 +399,12 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
 		}
 	}
 
+	// create rightsizing component
+	result, err = analyticsctrl.CreateRightSizingComponent(r.Client, r.Scheme, instance, r.Manager)
```
tremes, line 403: You can pass the `ctx` here and use it in the new functions.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_prometheusrule.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		if len(l.InclusionCriteria) > 0 && len(l.ExclusionCriteria) > 0 {
+			return monitoringv1.PrometheusRule{}, fmt.Errorf("only one of inclusion or exclusion allowed for label_env")
+		}
+		if len(l.InclusionCriteria) > 0 {
+			selector := fmt.Sprintf(`kube_namespace_labels{label_env=~"%s"}`, strings.Join(l.InclusionCriteria, "|"))
+			labelJoin = fmt.Sprintf(`* on (namespace) group_left() (%s or kube_namespace_labels{label_env=""})`, selector)
+		} else if len(l.ExclusionCriteria) > 0 {
+			selector := fmt.Sprintf(`kube_namespace_labels{label_env!~"%s"}`, strings.Join(l.ExclusionCriteria, "|"))
+			labelJoin = fmt.Sprintf(`* on (namespace) group_left() (%s or kube_namespace_labels{label_env=""})`, selector)
```
tremes, line 45: nit: the value of the `labelJoin` is the same in both branches so it could be moved out of the if/else

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go
```diff
@@ -0,0 +1,192 @@
...
+	err = c.Update(context.TODO(), policy)
+	if err != nil {
+		log.Error(err, "Failed to update the modified policy")
+		return err
+	}
+
+	log.Info("Successfully updated ComplianceType in policy")
+
+	// Wait for 1 seconds
+	time.Sleep(1 * time.Second)
```
tremes, line 76: we can probably remove this line right?

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go
```diff
@@ -0,0 +1,192 @@
...
+			RemediationAction: "inform",
+			Severity:          "low",
+			NamespaceSelector: configpolicyv1.Target{
+				Include: []configpolicyv1.NonEmptyString{
+					configpolicyv1.NonEmptyString(rsMonitoringNamespace),
+				},
+			},
+			ObjectTemplates: []*configpolicyv1.ObjectTemplate{
+				{
+					ComplianceType: "MustOnlyHave",
```
tremes, line 130: nit: you can use `configpolicyv1.MustOnlyHave ` as the value

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go
```diff
@@ -0,0 +1,208 @@
...
+			Namespace: config.GetDefaultNamespace(),
+		},
+	}
+
+	// Fetch the ConfigMap
+	err := c.Get(context.TODO(), types.NamespacedName{
+		Name:      rsConfigMapName,
+		Namespace: config.GetDefaultNamespace(),
+	}, existingCM)
+	log.Info("RS - fetch configmap completed2")
```
tremes, line 36: I am not sure. Maybe you plan to revisit the logging messages, but I can see these numbers (here as `completed2`) and `rs8` strings and similar which are not probably necessary.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go
```diff
@@ -0,0 +1,192 @@
...
+		Namespace: policy.Namespace,
+	}, policy)
+
+	if errPolicy != nil && !errors.IsNotFound(errPolicy) {
+		fmt.Println("Error retriving Policy:", errPolicy)
+		return errPolicy
+	}
+
+	// Marshal the PrometheusRule object into JSON
+	promRuleJSON, err := addAPIVersionAndKind(prometheusRule, "monitoring.coreos.com/v1", "PrometheusRule")
```
tremes, line 106: Maybe we could use `runtime.Convert_runtime_Object_To_runtime_RawExtension` instead? :thinking:

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go
rajusem: Explicitly Kept it so that Policy will be applied and propagate to underlying spoke cluster before deleting Policy.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go
rajusem: Yes we will be working on cleaning up logs.

operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go
```diff
@@ -116,6 +140,12 @@ type PlatformCapabilitiesSpec struct {
 	// +optional
 	// +kubebuilder:validation:Optional
 	Metrics PlatformMetricsSpec `json:"metrics,omitempty"`
+	// Analytics defines the spec for the analytics capabilities.
+	// Currently it support Right Sizing recommendation feature.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	Analytics *AnalyticCapabilitiesSpec `json:"analytics,omitempty"`
```
tremes, line 148: nit: I would probably call it only `AnalyticsSpec` and would remove the:
>// Currently it support Right Sizing recommendation feature.

because it won't be excatly true once there's also the incident detection included

operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
rajusem: Now using `ctx`.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go
rajusem: Fixed

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go
```diff
@@ -0,0 +1,192 @@
...
+	configPolicy := configpolicyv1.ConfigurationPolicy{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "policy.open-cluster-management.io/v1",
+			Kind:       "ConfigurationPolicy",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: rsPrometheusRulePolicyConfigName,
+		},
+		Spec: &configpolicyv1.ConfigurationPolicySpec{
+			RemediationAction: "inform",
```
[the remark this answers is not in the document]
rajusem: Fixed

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go
rajusem: Corrected logs and removed unwanted logs.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go
rajusem: Good idea, implemented the code as suggested, which avoids the need for JSON marshaling/unmarshaling..

operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go
rajusem: Merged changes with latest as changes are already in place with incident detection PR.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go
rajusem: We have removed this codebase, now using `pruneObjectBehavior` property with the Policy.

operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go
```diff
@@ -132,6 +132,12 @@ type PlatformAnalyticsSpec struct {
 	// +optional
 	// +kubebuilder:validation:Optional
 	IncidentDetection PlatformIncidentDetectionSpec `json:"incidentDetection,omitempty"`
+
+	// Feature to enable namespace right-sizing recommendation capabilities for the Analytics.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	NamespaceRightSizingRecommendation NamespaceRightSizingRecommendationSpec `json:"namespaceRightSizingRecommendation,omitempty"`
```
tremes, line 140: nit: I remember I was asked to prefix the type name with `Platform` here (see above).

operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go
```diff
@@ -142,6 +148,24 @@ type PlatformIncidentDetectionSpec struct {
 	Enabled bool `json:"enabled,omitempty"`
 }
 
+// NamespaceRightSizingSpec defines the namespace right sizing capabilities.
+// +kubebuilder:object:generate=true
+// +kubebuilder:validation:Type=object
```
tremes, line 153: I think you don't need the kubebuilder annotations here

operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go
```diff
@@ -142,6 +148,24 @@ type PlatformIncidentDetectionSpec struct {
...
+type NamespaceRightSizingRecommendationSpec struct {
+	// Enabled defines a flag to enable/disable the namespace right-sizing feature for the Analytics.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	Enabled bool `json:"enabled,omitempty"`
+
+	// NamespaceBinding defines where we want to create all resources under specific namespace as well as used in
+	// PolicyBinding. If not supplied then we will use open-cluster-management-global-set.
```
tremes, line 162: I would try to rephrase this. Perhaps something like:
```
NamespaceBinding defines the namespace where all the required resources are created. The default value (or namespace?) is `open-cluster-management-global-set`
```

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_controller.go
```diff
@@ -0,0 +1,115 @@
...
+	PrometheusRuleConfig   RSPrometheusRuleConfig   `yaml:"prometheusRuleConfig"`
+	PlacementConfiguration clusterv1beta1.Placement `yaml:"placementConfiguration"`
+}
+
+func CreateRightSizingComponent(
+	ctx context.Context,
+	c client.Client,
+	scheme *runtime.Scheme,
+	mco *mcov1beta2.MultiClusterObservability,
+	mgr ctrl.Manager,
```
tremes, line 60: are the `mgr` and `scheme` params needed? It seems they are not

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_controller.go
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+// Licensed under the Apache License 2.0
+
+package analytics
+
+import (
+	"context"
+
+	"github.com/cloudflare/cfssl/log"
```
tremes, line 10: What is the reason for using this log? It looks like other parts of MCO uses `sigs.k8s.io/controller-runtime/pkg/log`

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_controller_test.go
```diff
@@ -0,0 +1,117 @@
...
+	mco := newTestMCO("", false)
+
+	k8sClient := fake.NewClientBuilder().
+		WithScheme(scheme).
+		WithObjects(mco).
+		Build()
+
+	result, err := analyticsctrl.CreateRightSizingComponent(context.TODO(), k8sClient, scheme, mco, nil)
+	require.NoError(t, err)
+	assert.Nil(t, result)
```
tremes, line 116: I personally prefer the [Go table driven tests](https://go.dev/wiki/TableDrivenTests) but it's just matter of preference.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_helper.go
```diff
@@ -0,0 +1,26 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+// FormatYAML converts a Go data structure to a YAML-formatted string
+func FormatYAML(data interface{}) string {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Error(nil, "Panic recovered during YAML marshaling: %v", r)
+		}
+	}()
```
tremes, line 18: Why is this required please?

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_helper.go
```diff
@@ -0,0 +1,26 @@
...
+
+package analytics
+
+import (
+	"github.com/cloudflare/cfssl/log"
+	"gopkg.in/yaml.v2"
+)
+
+// FormatYAML converts a Go data structure to a YAML-formatted string
+func FormatYAML(data interface{}) string {
```
tremes, line 13: maybe it can be `func FormatYAML[T RSPrometheusRuleConfig | clusterv1beta1.Placement](data T) string {`

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go
```diff
@@ -0,0 +1,137 @@
...
+func createOrUpdatePrometheusRulePolicy(
+	ctx context.Context,
+	c client.Client,
+	prometheusRule monitoringv1.PrometheusRule) error {
+
+	policy := &policyv1.Policy{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "policy.open-cluster-management.io/v1",
+			Kind:       "Policy",
+		},
```
tremes, line 32: Is this type metadata required? I guess it should not be required.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_placementbinding_test.go
```diff
@@ -0,0 +1,78 @@
...
+	policyv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+)
+
+func TestCreatePlacementBinding_CreatesWhenNotFound(t *testing.T) {
+	scheme := runtime.NewScheme()
+	_ = policyv1.AddToScheme(scheme)
+
+	client := fake.NewClientBuilder().WithScheme(scheme).Build()
+
+	err := createPlacementBinding(context.TODO(), client)
```
tremes, line 26: nit: you can create `ctx := context.Background()` and use it in the test repeatedly.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go
```diff
@@ -0,0 +1,194 @@
...
+
+	processConfigMap := func(cm *corev1.ConfigMap) bool {
+		configData, err := GetRightSizingConfigData(cm)
+		if err != nil {
+			log.Error(err, "Failed to extract RightSizingConfigData")
+			return false
+		}
+
+		// Apply changes based on the config map
+		if err := applyRSNamespaceConfigMapChanges(context.Background(), c, configData); err != nil {
```
tremes, line 130: Ideally we should pass the context to the `GetNamespaceRSConfigMapPredicateFunc`

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			if e.Object.GetName() == rsConfigMapName && e.Object.GetNamespace() == config.GetDefaultNamespace() {
+				if cm, ok := e.Object.(*corev1.ConfigMap); ok {
+					return processConfigMap(cm)
+				}
+			}
+			return true
```
tremes, line 144: Shouldn't be here `false`?

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go
```diff
@@ -0,0 +1,194 @@
...
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectNew.GetName() != rsConfigMapName || e.ObjectNew.GetNamespace() != config.GetDefaultNamespace() {
+				return true
```
tremes, line 148: Same here - shouldn't be `false` ?

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+
+			// Check if the ConfigMap `Data` has changed before proceeding
+			oldCM, oldOK := e.ObjectOld.(*corev1.ConfigMap)
+			newCM, newOK := e.ObjectNew.(*corev1.ConfigMap)
+
+			if oldOK && newOK && reflect.DeepEqual(oldCM.Data, newCM.Data) {
+				log.Info("No changes detected in ConfigMap data, skipping update")
+				return true
```
tremes, line 157: Same here - if there are no changes then we probably don't need to reconcile or am I wrong?

operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go
```diff
@@ -148,18 +148,15 @@ type PlatformIncidentDetectionSpec struct {
...
 	// Enabled defines a flag to enable/disable the namespace right-sizing feature for the Analytics.
 	//
 	// +optional
 	// +kubebuilder:validation:Optional
 	Enabled bool `json:"enabled,omitempty"`
 
-	// NamespaceBinding defines where we want to create all resources under specific namespace as well as used in
-	// PolicyBinding. If not supplied then we will use open-cluster-management-global-set.
+	// NamespaceBinding defines the namespace where all the required resources are created.
+	// The default value (or namespace?) is `open-cluster-management-global-set`
```
tremes, line 159: I meant it as either:
```The default value is `open-cluster-management-global-set```
or
```The default namespace is `open-cluster-management-global-set``` :-)

File: operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go (hunk 9 above)
Comment (dvandra) [reply operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go]: Fixed the type with prefix as `Platform`.

File: operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go (hunk 10 above)
Comment (dvandra) [reply operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go]: ACK

File: operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go (hunk 11 above)
Comment (dvandra) [reply operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go]: Updated as requested.

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_controller.go (hunk 13 above)
Comment (dvandra) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_controller.go]: Updated the logs.

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_helper.go (hunk 15 above)
Comment (dvandra) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_helper.go]: Removed as it is not required.

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_helper.go (hunk 16 above)
Comment (dvandra) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_helper.go]: Updated the function parameters.

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go (hunk 17 above)
Comment (dvandra) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go]: True, removed type metadata.

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_placementbinding_test.go (hunk 18 above)
Comment (dvandra) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_placementbinding_test.go]: ACK

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go (hunk 19 above)
Comment (dvandra) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go]: ACK

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go (hunk 20 above)
Comment (dvandra) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go]: Updated the return as false.

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go (hunk 21 above)
Comment (dvandra) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go]: Updated the return as false.

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go (hunk 22 above)
Comment (dvandra) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go]: Updated the return as false.

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go (hunk 20 above)
Comment (rajusem) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go:143]: As we have two predicate functions, we are taking approach of **Loosen Restrictions – Return true When Event Not Relevant**. 

Also `e2e-kind` is failing with returning `false`, so keeping it with `true`

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go (hunk 21 above)
Comment (rajusem) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go:147]: As we have two predicate functions, we are taking approach of **Loosen Restrictions – Return true When Event Not Relevant**. 

Also `e2e-kind` is failing with returning `false`, so keeping it with `true`

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go (hunk 22 above)
Comment (rajusem) [reply operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go:156]: As we have two predicate functions, we are taking approach of **Loosen Restrictions – Return true When Event Not Relevant**. 

Also `e2e-kind` is failing with returning `false`, so keeping it with `true`

**Comment by tremes** [comment]
I haven't tested it, but I have already reviewed it couple of times.
/lgtm

**Comment by thibaultmg** [comment]
Can you please add a scrapeConfig in the grafana analytics directory containing the new dashboard so that the needed metrics are also federated when MCOA for metrics is enabled? 
I'll add CI checks for this new analytics directory in another PR (won't be blocking this one)

**Review by jacobbaungard — COMMENTED** [review]
Looks pretty sensible to me on a first pass. Since this is a big PR I also asked another engineer to take a look.

As a general comment, I think the logging here is very verbose, and I also see some logs which are very generic like the below, whereas other has `RS` prefixed - would be good to have a common pattern.
```
log.Info("ConfigMap data has changed, processing update")
```

Could we consider doing a pass over the logs and see if there are some which could be removed?

File: operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go [hunk 24]
```diff
@@ -0,0 +1,193 @@
...
+		existingCM.Data = GetDefaultRSNamespaceConfig()
+
+		// Create the ConfigMap
+		err := c.Create(ctx, existingCM)
+		if err != nil {
+			log.Error(err, "RS - Failed to create ConfigMap", "ConfigMap", rsConfigMapName)
+			return err
+		}
+		log.Info("RS - Created configMap completed")
+	} else {
```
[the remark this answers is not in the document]
rajusem: Thanks for pointing out, updated code to handle some other error as well.

rajusem:
> Can you please add a scrapeConfig in the grafana analytics directory containing the new dashboard so that the needed metrics are also federated when MCOA for metrics is enabled? I'll add CI checks for this new analytics directory in another PR (won't be blocking this one)

@thibaultmg  Ack. Added scrapeConfig in analytics directory.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_controller.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Set to default namespace if not given
+	if newBinding == "" {
+		newBinding = rsDefaultNamespace
+	}
+
+	// Check if right-sizing namespace feature enabled or not
+	// If disabled then cleanup related resources
+	if !isRightSizingNamespaceEnabled {
+		log.Info("RS - NamespaceRightSizing feature not enabled")
```
thibaultmg, line 76: Could we set this one to debug level? It will show up on every reconcile loop otherwise.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_controller.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	if namespaceBindingUpdated {
+		// Clean up resources except config map to update NamespaceBinding
+		cleanupRSNamespaceResources(ctx, c, existingNamespace, true)
+
+		// Get configmap
+		cm := &corev1.ConfigMap{}
+		if err := c.Get(ctx, client.ObjectKey{Name: rsConfigMapName, Namespace: config.GetDefaultNamespace()}, cm); err != nil {
+			log.Error(err, "RS - Failed to get existing ConfigMap")
```
thibaultmg, line 105: We can remove these error logs. They will get logged by the controller runtime when the reconcile loop returns an error (which is the case here). We should rather wrap the error with these details to avoid double error logs.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_configmap.go
```diff
@@ -0,0 +1,191 @@
...
+			if e.ObjectNew.GetName() != rsConfigMapName || e.ObjectNew.GetNamespace() != config.GetDefaultNamespace() {
+				return true
+			}
+
+			// Check if the ConfigMap `Data` has changed before proceeding
+			oldCM, oldOK := e.ObjectOld.(*corev1.ConfigMap)
+			newCM, newOK := e.ObjectNew.(*corev1.ConfigMap)
+
+			if oldOK && newOK && reflect.DeepEqual(oldCM.Data, newCM.Data) {
+				log.Info("RS - No changes detected in ConfigMap data, skipping update")
```
thibaultmg, line 156: Let's use debug logging for noop.

operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -394,6 +395,12 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
 		}
 	}
 
+	// create rightsizing component
+	result, err = analyticsctrl.CreateRightSizingComponent(ctx, r.Client, instance)
+	if result != nil {
```
thibaultmg, line 400: I'm not sure I understand. When the result is not nil, we return right away from the reconcile loop? Why do we skip the rest of the processing?

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_controller.go
```diff
@@ -0,0 +1,122 @@
...
+	// Set isEnabled flag which will be used for checking namespaceBindingUpdated condition next time
+	isEnabled = true
+
+	// Retrieve the existing namespace as existingNamespace and update the new namespace in rsNamespace
+	existingNamespace := rsNamespace
+	rsNamespace = newBinding
+
+	// Creating configmap with default values
+	if err := EnsureRSNamespaceConfigMapExists(ctx, c); err != nil {
+		return nil, err
```
thibaultmg, line 95: It's good, we have reduced unnecessary logging, however I would wrap errors to add context. It might be difficult to debug otherwise. Just a recommendation, not a blocker for the review.

operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
rajusem: We followed existing logic in the codebase like below. everywhere we are retuning and next logic is not processing I believe. 

```
// create an Observatorium CR
result, err = GenerateObservatoriumCR(r.Client, r.Scheme, instance)
if result != nil {
   return *result, fmt.Errorf("failed to generate the observatorium CR: %w", err)
}

// generate grafana datasource to point to observatorium api gateway
result, err = GenerateGrafanaDataSource(r.Client, r.Scheme, instance)
if result != nil {
   return *result, fmt.Errorf("failed to generate Grafana data source: %w", err)
}
```

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_controller.go
rajusem: Ack @jacobbaungard, I miss read earlier comment for the error, thanks for bringing it up. Now wrapped error with context.

operators/multiclusterobservability/controllers/multiclusterobservability/analytics/rightsizing_namespace_policy.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	if errors.IsNotFound(errPolicy) {
+		if err = c.Create(ctx, policy); err != nil {
+			return err
+		}
+		log.Info("RS - Created PrometheusRulePolicy successfully", logCtx...)
+	} else {
+		if err := c.Update(ctx, policy); err != nil {
```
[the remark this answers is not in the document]
thibaultmg: Discussed in private, we can ignore my comment.

operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
rajusem: Also now returning only err from rightsizing controller, removed `result` which we were returning nil in all cases earlier.

PR #1941, opened 2025-04-15
[ACM-19970] Make sure detached cluster is cleaned up

coleenquadros:
…ring reconcile

When a cluster detaches, it will still be listed in the managedcluster resource but it will have a DeletionTimeStamp and this causes MCO to think it needs to install Observability. Make sure the managed cluster is not a part of the managedcluster list during delete event.

https://issues.redhat.com/browse/ACM-19970


thibaultmg (approved):
/lgtm

PR #1945, opened 2025-04-16
[Fix]: filter out non standard condition types when replicating addon status conditions

thibaultmg:
Cf https://issues.redhat.com/browse/ACM-19714


philipgough (approved):
/lgtm

PR #1949, opened 2025-04-17
Remove Bearer from authorization header when setting X-Forwarded-Access-Token

### LINKED ISSUE #1942: rbac-query-proxy: Token assignment includes `bearer`.
When forwarding Grafanas oauth identity to a datasource behind the proxy the token extraction includes `Bearer` and the token used is invalid.

If the token in the `Authorization:`header includes `Bearer <token>` the token assignment below should remove `Bearer`. 

https://github.com/stolostron/multicluster-observability-operator/blob/5f4a1d3ee077aac64fdcb512d3222ce3b9e150d4/proxy/pkg/proxy/proxy.go#L132

cldmnky:
Remove `Bearer ` from Authorization header when setting X-Forwarded-Access-Token.

This fixes #1942

### DIFF
--- a/proxy/pkg/proxy/proxy.go
+++ b/proxy/pkg/proxy/proxy.go
@@ -133,6 +133,8 @@ func preCheckRequest(req *http.Request) error {
 		if token == "" {
 			return errors.New("found unauthorized user")
 		} else {
+			// Remove Bearer from token if present
+			token = strings.TrimPrefix(token, "Bearer ")
 			req.Header.Set("X-Forwarded-Access-Token", token)
 		}
 	}
--- a/proxy/pkg/proxy/proxy.go
+++ b/proxy/pkg/proxy/proxy.go
@@ -143,7 +143,7 @@ func preCheckRequest(req *http.Request) error {
 	if userName == "" {
 		userName = util.GetUserName(token, config.GetConfigOrDie().Host+userAPIPath)
 		if userName == "" {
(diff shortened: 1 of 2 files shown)


jacobbaungard:
Can you expand a little on why you need this exactly? As far as I know there are no known issues with the datasources we setup currently (and I am not exactly aware that we support adding others).

iirc Grafana these days purposely filter out the `Authorization` header so I am not sure we would ever see it?

cldmnky:
I'm adding a DS to the mcoo setup in a separate Grafana instance managed by the Grafana Operator. I'm using Dex and the generic oauth in that instance. I'm trying to use the oauth forwarding for auth in the DS. Grafana sends an Authorization header with `Bearer ` prepended as in the oauth spec.

The code in the proxy breaks since the header value is just copied into X-Forwarded-Access-Token.

I'm aware that this is not supported but at least the current implementation in the code should be correct 😅

jacobbaungard:
Ack, I think it would be OK to merge, I'd double check a couple of things next week.

Actually, if we can use the `Authorization` header instead of the `X-Forwarded-Access-Token` in the general case, it would save us some headaches downstream.

jacobbaungard (approved):
/lgtm
/ok-to-test

jacobbaungard:
Not sure why Konflux doesn't run. Anyway, this should work. Merging.

PR #1952, opened 2025-04-17
Tests: Make JUnit report conform to QE standard

jacobbaungard:
All output should now align with the correct format:
```
<Polarion test case ID>: <Component name>: <Test case name>
```


philipgough (approved):
/lgtm

PR #1953, opened 2025-04-22
fix: use correct CRD values for MCOA

tremes:


operators/multiclusterobservability/manifests/base/multicluster-observability-addon/cluster_role.yaml
```diff
@@ -70,5 +70,9 @@
       verbs: ["get", "list", "watch"]
     # Role for addon to perform prometheus specific actions
     - apiGroups: ["monitoring.coreos.com"]
-      resources: ["prometheusagents", "scrapeconfigs", "prometheusrules"]
-      verbs: ["get", "list", "watch", "create", "update", "delete"]
+      resources: ["scrapeconfigs", "prometheusagents", "prometheusrules", "servicemonitors"]
+      verbs: ["get", "list", "watch"]
```
thibaultmg, line 74: Why are you removing some verbs here? They are needed for metrics collection.

tremes: ahh sorry my bad. I didn't notice that

thibaultmg (approved):
/lgtm

PR #1986, opened 2025-05-08
Grafana: Remove OCP 3.11 dashboards

jacobbaungard:
OCP 3.11 is no longer supported, so these dashboards can be removed.


thibaultmg (approved):
/lgtm

PR #1991, opened 2025-05-14
ACM-15008: Updated the descriptions in ACM HCP overview dashboard

rokej:
Updated the descriptions in ACM HCP overview dashboard


jacobbaungard (approved):
/lgtm

PR #2003, opened 2025-05-16
Tests: pin victoria metrics image

jacobbaungard:
The `stable` image appears to be removed as well. Pin it to a specific version so the tests can pass.


thibaultmg (approved):
/lgtm

PR #2009, opened 2025-05-19
Improve some e2e

thibaultmg:
- Avoid failing at the uninstall step if some resources are not found
- Check early if store secret has been reset in QE jenkins env
- Simplify metrics check with a fixed list, easier to maintain as project are adding they dashboards and updating the collected metrics


jacobbaungard (approved):
/lgtm

PR #2013, opened 2025-05-20
Correctly handle custom-certs in config tests

jacobbaungard:
The config tests, which applies CRs from `examples/updatemcocr` did not correctly handled the cases where the minio-tls-secret was not applied. This commit ensure have two versions of the mcoCR that handle both when the tls-secret is created and when it is not.

/cherry-pick release-2.13


thibaultmg (approved):
/lgtm

PR #2017, opened 2025-05-21
rbac_test: correct login to kube:admin+cleanup env

jacobbaungard:
- Fix kube:admin login in rbac tests to work when using username/pw instead of kubecontext
- Correctly clean up rbac env from any previous tests
- Make sure we logout from any non kube:admin user after each rbac test (so not to break any future tests in case of failures)


thibaultmg (approved):
/lgtm

PR #2018, opened 2025-05-22
Fix CI metrics check

thibaultmg:
- Adds the `--ignored-rules-missing-definition` flag to the rulescheck command.
- Adds the `cluster:health:components:map` metrics to the ignored rules as it is defined directly through COO for incident detection feature.


jacobbaungard (approved):
/lgtm

PR #2022, opened 2025-05-27
Tests: skip RHACM4K-7518 if no managed cluster

jacobbaungard:
This test depends on the managed cluster being correctly added to the testoptions. Therefore we skip the test completetly, if this is not the case.


jacobbaungard:
> But we will always have one managed cluster noh (local-cluster)
> 
> https://github.com/stolostron/multicluster-observability-operator/blob/7d47f2190f820c5d3bba160d4d93667b10b03593/cicd-scripts/run-e2e-tests.sh#L68

Not when running through Jenkins.

coleenquadros (approved):

PR #2033, opened 2025-06-04
Fix home grafana dashboard for MCOA

thibaultmg:


jacobbaungard (commented):
I wonder if we should use an annotation instead of labels? I think best practice (which I learned after I created the first commit) is that labels = for identifying the resource, annotation = info about the resource itself.

PR #2035, opened 2025-06-06
[MCOA]: Add metrics collector alerts

thibaultmg:
Adds some alerts for federation queries and remote write.

### DIFF
--- a/operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
+++ b/operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
@@ -209,3 +209,107 @@ spec:
         for: 1m
         labels:
           severity: "{{ $labels.severity }}"
+  - name: acm-metrics-collector-federation-alerts
+    rules:
+      - alert: HighFederationEndpointErrorRate5Percent
+        expr: |
+          (
+            sum(
+              rate(prometheus_http_requests_total{handler="/federate",namespace=~"openshift-monitoring|openshift-user-workload-monitoring",job=~"prometheus-k8s|prometheus-user-workload",code=~"5..|4.."}[5m])
+            ) by (job, pod) 
+            / 
+            sum(
+              rate(prometheus_http_requests_total{handler="/federate",namespace=~"openshift-monitoring|openshift-user-workload-monitoring",job=~"prometheus-k8s|prometheus-user-workload"}[5m])
+            ) by (job, pod)
+            * 100
+          ) > 5
+          AND 
+          sum(
+            rate(prometheus_http_requests_total{handler="/federate",namespace=~"openshift-monitoring|openshift-user-workload-monitoring",job=~"prometheus-k8s|prometheus-user-workload"}[5m])
+          ) by (job, pod) 
+          > 0.1
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          summary: High error rate on /federate endpoint for {{ $labels.job }}/{{ $labels.pod }}.
+          description: 'The /federate endpoint on pod {{ $labels.pod }} (job {{ $labels.job }}) is experiencing an error rate of {{ printf "%.2f" $value }}% over the last 10 minutes, which is above the 5% threshold. This could be caused by Prometheus being overloaded, network issues, or misconfiguration. Check the logs of the pod.'
+      - alert: HighFederationEndpointLatencyOver8s
+        expr: |
+          sum (
+            increase(prometheus_http_request_duration_seconds_bucket{handler="/federate",namespace=~"openshift-monitoring|openshift-user-workload-monitoring",job=~"prometheus-k8s|prometheus-user-workload", le="+Inf"}[10m])
+            - 
+            increase(prometheus_http_request_duration_seconds_bucket{handler="/federate",namespace=~"openshift-monitoring|openshift-user-workload-monitoring",job=~"prometheus-k8s|prometheus-user-workload", le="8"}[10m])
+          ) by (job, pod) 
+          >0
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          summary: High latency on /federate endpoint ({{ $labels.job }}/{{ $labels.pod }}).
+          description: '{{ printf "%.0f" $value }} federation request(s) on pod {{ $labels.pod }} (job {{ $labels.job }}) took longer than 8 seconds in the last 10 minutes. This indicates potential performance degradation or a scrape query federating too many metrics. Check for high CPU/memory usage or long-running queries on the Prometheus pod.'
+      - alert: HighFederationEndpointLatencyOver20s
+        expr: |
+          sum (
+            increase(prometheus_http_request_duration_seconds_bucket{handler="/federate",namespace=~"openshift-monitoring|openshift-user-workload-monitoring",job=~"prometheus-k8s|prometheus-user-workload", le="+Inf"}[10m])
+            - 
+            increase(prometheus_http_request_duration_seconds_bucket{handler="/federate",namespace=~"openshift-monitoring|openshift-user-workload-monitoring",job=~"prometheus-k8s|prometheus-user-workload", le="20"}[10m])
+          ) by (job, pod) 
+          > 0
+        for: 10m
+        labels:
+          severity: critical
+        annotations:
+          summary: Critically high latency on /federate endpoint ({{ $labels.job }}/{{ $labels.pod }}).
+          description: '{{ printf "%.0f" $value }} federation request(s) on pod {{ $labels.pod }} (job {{ $labels.job }}) took longer than 20 seconds in the last 10 minutes. This is critically close to the 30s proxy timeout and may lead to failed scrapes. Investigate performance on the Prometheus pod immediately.'
+      - alert: MetricsCollectorNotIngestingSamples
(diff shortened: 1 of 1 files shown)


operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
```diff
@@ -209,3 +209,107 @@ spec:
...
+        expr: |
+          max_over_time(prometheus_remote_storage_shards_desired{job=~"platform-metrics-collector|user-workload-metrics-collector"}[5m]) 
+          > 
+          max_over_time(prometheus_remote_storage_shards_max{job=~"platform-metrics-collector|user-workload-metrics-collector"}[5m])
+        for: 15m
+        labels:
+          severity: warning
+        annotations:
+          summary: Prometheus Agent {{ $labels.namespace }}/{{ $labels.pod }} remote write needs more shards.
+          description: 'The Prometheus Agent pod {{ $labels.pod }} (namespace {{ $labels.namespace }}) remote write wants to run {{ printf "%.1f" $value }} shards for remote_name {{ $labels.remote_name}}, exceeding its configured maximum. This indicates sustained backpressure. Chack the queueConfig.maxShards for this remote_name or use the metric prometheus_remote_storage_shards_max.'
```
tremes: nit: Typo in "Chack"

operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
```diff
@@ -209,3 +209,107 @@ spec:
...
+          ) by (job, pod) 
+          > 0.1
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          summary: High error rate on /federate endpoint for {{ $labels.job }}/{{ $labels.pod }}.
+          description: 'The /federate endpoint on pod {{ $labels.pod }} (job {{ $labels.job }}) is experiencing an error rate of {{ printf "%.2f" $value }}% over the last 10 minutes, which is above the 5% threshold. This could be caused by Prometheus being overloaded, network issues, or misconfiguration. Check the logs of the pod.'
+      - alert: HighFederationEndpointLatencyOver8s
+        expr: |
```
simonpasquier: (nit) why not a ratio?

operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
```diff
@@ -209,3 +209,107 @@ spec:
...
         labels:
           severity: "{{ $labels.severity }}"
+  - name: acm-metrics-collector-federation-alerts
+    rules:
+      - alert: HighFederationEndpointErrorRate5Percent
+        expr: |
+          (
+            sum(
+              rate(prometheus_http_requests_total{handler="/federate",namespace=~"openshift-monitoring|openshift-user-workload-monitoring",job=~"prometheus-k8s|prometheus-user-workload",code=~"5..|4.."}[5m])
+            ) by (job, pod)
```
simonpasquier: (nit) I have a personal preference for `sum by(...) (...)` over `sum(...) by(...)` (easier to read).

operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
```diff
@@ -209,3 +209,107 @@ spec:
...
+          sum(
+            rate(prometheus_http_requests_total{handler="/federate",namespace=~"openshift-monitoring|openshift-user-workload-monitoring",job=~"prometheus-k8s|prometheus-user-workload"}[5m])
+          ) by (job, pod) 
+          > 0.1
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          summary: High error rate on /federate endpoint for {{ $labels.job }}/{{ $labels.pod }}.
+          description: 'The /federate endpoint on pod {{ $labels.pod }} (job {{ $labels.job }}) is experiencing an error rate of {{ printf "%.2f" $value }}% over the last 10 minutes, which is above the 5% threshold. This could be caused by Prometheus being overloaded, network issues, or misconfiguration. Check the logs of the pod.'
```
simonpasquier: > Check the logs of the pod.

I'm not sure that the logs would show something?

operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
```diff
@@ -209,3 +209,107 @@ spec:
...
+          ) 
+          * 100 
+          > 10
+        for: 10m
+        labels:
+          severity: critical
+        annotations:
+          summary: Prometheus Agent {{ $labels.namespace }}/{{ $labels.pod }} fails to remote write to the Hub.
+          description: 'The Prometheus Agent pod {{ $labels.pod }} in namespace {{ $labels.namespace }} fails to send {{ printf "%.1f" $value }}% of its samples to remote_name {{ $labels.remote_name }}. Check Agent logs and Hub receiver.'
+      - alert: MetricsCollectorRemoteWriteBehind
```
simonpasquier, line 253: cc @machine424 we had some recent bug on these alerting rules.

simonpasquier (commented):
Could the alerts trigger if an external actor/service queries the /federate endpoint?

thibaultmg:
> Could the alerts trigger if an external actor/service queries the /federate endpoint?

Yes it will... I found no metric for alerting from the prometheus agent... Maybe I missed something?

operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
thibaultmg: For context, we are potentially only making one request every 5 minutes if only default metrics are scraped. This is why I am not using quantile functions.
Regarding the ratio I'm not sure. Clients will add their own scrape configs. They don't know when they overuse federation. The purpose is to warn them when they have a scrapeConfig that is federating too many metrics, before they hit support with missing data or timeouts. 
In that regard I feel that the ratio would be less efficient. WDYT?

operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
thibaultmg: ok, works for me, changing that if that's more conventional.

operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
thibaultmg: I think it does with debug mode, but I'm not sure, I don't remember. Removing it.

operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
thibaultmg: Ok, I'm interested into knowing more!

operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
simonpasquier: I have a personal preference for "x% of the requests failed over the last Y minutes" vs. "x requests failed over the last Y minutes" because the former is more precise IMHO.

simonpasquier:
> > Could the alerts trigger if an external actor/service queries the /federate endpoint?
> 
> Yes it will

We suspect that a significant fraction of customers (ab)use federation already so it might create noise for MCOA. Is there a possibility to alert from the client's perspective?

thibaultmg:
> Prometheus server and agent expose the same metrics so there are no specific alerting rules for agent.



> > > Could the alerts trigger if an external actor/service queries the /federate endpoint?
> > 
> > 
> > Yes it will
> 
> We suspect that a significant fraction of customers (ab)use federation already so it might create noise for MCOA. Is there a possibility to alert from the client's perspective?

Ok, good to know. Then I should probably remove this alert. I don't have client side metrics for the calls to the federation endpoint (from the prometheus agent perpective).

operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
thibaultmg: ok, sounds good. I'll change that.

simonpasquier:
> I don't have client side metrics for the calls to the federation endpoint (from the prometheus agent perpective).

well you should have the `up` metric of the target?

thibaultmg:
I've removed all federation related alerts as it will be too noisy (as discussed, we only have server side metrics).

operators/multiclusterobservability/manifests/base/grafana/platform-mcoa/prometheus-rule.yaml
machine424: yes, if the agent has nothing to send (because relabeling yields no series e.g.) for more than 12mins, the alert would fire, which could be undesirable, see https://issues.redhat.com/browse/OCPBUGS-56568.
I'll see how we can address this upstream, then we can backport here.

thibaultmg: ok, thanks. I don't think we could have such a case for ACM Observability, so I'll keep it for now.

jacobbaungard (approved):
/lgtm

PR #2046, opened 2025-06-10
fix: Corrected type as Instant for the cluster CPU data points

rajusem:
Corrected data type as instant for the CPU data points in the grafana dashboard 

Epic - https://issues.redhat.com/browse/ACM-17966

Signed-off-by: Raj Zalavadia [rzalavad@redhat.com](mailto:rzalavad@redhat.com)


operators/multiclusterobservability/manifests/base/grafana/analytics/dash-acm-right-sizing-namespace.yaml
```diff
@@ -58,7 +58,8 @@ data:
         },
         {
           "datasource": {
-            "uid": "${datasource}"
+            "type": "prometheus",
```
thibaultmg, line 61: Are you sure about this change? This is a leftover from dev no?

rajusem: Yes you are correct let me change that, also I see already merged code have this  dev left over, let me change it too

rajusem: @thibaultmg  Now replaced hardcoded datasource uid `"uid": "000000001"` with variable as`"uid": "${datasource}"`. Earlier code  will also work as it will pick 1st available prometheus datasource(Observatorium). But good to use `${datasource}` so we don't need to change code everywhere incase if we want to change datasource,  we can simply change configuration one place if required.

PR #2052, opened 2025-06-12
MCO Annotation  uwm-alert-forwarding to optionally disable UWM alert forwarding

subbarao-meduri:
Related to [ACM-21198](https://issues.redhat.com/browse/ACM-21198) 

This PR supports a new MCO annotation to optionally disable UWM alert forwarding. This would be useful in environments where customers want to manage UWM config map externally and avoid potential conflicts.

```
oc annotate MultiClusterObservability observability --overwrite mco-disable-uwl-alerting=true
```

While  setting `mco-disable-alerting: "true"` disables all alert forwarding from spoke clusters, `mco-disable-uwl-alerting: "true"` specifically disables UWM alert forwarding while continuing to forward platform alerts. 

Setup:   (Single hub + Spoke)
CMO config = `cluster-monitoring-config` config map in `openshift-monitoring` namespace.
User Workload config = `user-workload-monitoring-config` in `openshift-user-workload-monitoring` namespace
Enable User workload monitoring on the spoke. Deploy an application that generates user workload metrics. Deploy a prometheus rule and generate an alert on the spoke locally.

- [x] Default configuration (no annotations). Verify CMO Config and User workload config are  updated with hub alert manager configuration on both hub and spoke.
- [x] Set annotation `mco-disable-uwl-alerting=true` on MCO. verify CMO config is still present, User workload configuration is reverted to remove alert manager config removed on both hub and spoke.
- [x] Set annotation `mco-disable-uwl-alerting=false` on MCO. Verify CMO config is still present, User workload config is restored to include hub alert manager config on both hub and spoke.
- [x] Remove annotation  `mco-disable-uwl-alerting` on MCO. (This is same as setting `mco-disable-uwl-alerting=true`) Verify CMO config is still present, User workload config continues to specify alert manager config on both hub and spoke. (same as setting it as false)
- [x] Add annotation `mco-disable-alerting=true` on MCO. Verify CMO and User workload configurations are reverted on both hub and spoke.
- [x] Add back `mco-disable-uwl-alerting=true` on MCO while keeping `mco-disable-alerting=true`. Verify CMO config and User workload config do not incldue hub alert manager information
- [x] Remove `mco-disable-alerting=true` on MCO. CMO config incldues alert manager config, but not User workload config.

### DIFF
--- a/operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
+++ b/operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
@@ -587,6 +587,13 @@ func (r *ObservabilityAddonReconciler) SetupWithManager(mgr ctrl.Manager) error 
 			builder.WithPredicates(configMapDataChangedPredicate(clusterMonitoringConfigName, promNamespace)),
 		).
 		Watches(
+			&corev1.ConfigMap{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(configMapDataChangedPredicate(
+				operatorconfig.OCPUserWorkloadMonitoringConfigMap,
+				operatorconfig.OCPUserWorkloadMonitoringNamespace)),
+		).
+		Watches(
 			&appsv1.Deployment{},
 			&handler.EnqueueRequestForObject{},
 			builder.WithPredicates(getPred(metricsCollectorName, r.Namespace, true, true, true)),
--- a/operators/multiclusterobservability/pkg/config/config.go
+++ b/operators/multiclusterobservability/pkg/config/config.go
@@ -57,7 +57,7 @@ const (
 	AnnotationMCOPause           = "mco-pause"
 	AnnotationCertDuration       = "mco-cert-duration"
 	AnnotationDisableMCOAlerting = "mco-disable-alerting"
-	AnnotationDisableUWMAlerting = "mco-disable-uwm-alerting"
+	AnnotationDisableUWMAlerting = "mco-disable-uwl-alerting"
 
 	MCHUpdatedRequestName               = "mch-updated-request"
 	MCOUpdatedRequestName               = "mco-updated-request"
--- a/operators/multiclusterobservability/controllers/placementrule/hub_info_secret_test.go
+++ b/operators/multiclusterobservability/controllers/placementrule/hub_info_secret_test.go
@@ -196,6 +196,64 @@ func TestNewSecret(t *testing.T) {
 		t.Fatalf("Wrong content in hub info secret: \ngot: "+hub.ObservatoriumAPIEndpoint+" "+hub.AlertmanagerEndpoint+" "+hub.AlertmanagerRouterCA, clusterName+" "+"https://test-host"+" "+"test-host"+" "+routerCA)
 	}
 
+	// Test UWM alerting disabled
+	mco.Annotations = map[string]string{config.AnnotationDisableUWMAlerting: "true"}
+	hubInfo, err = generateHubInfoSecret(c, mcoNamespace, namespace, true, config.IsUWMAlertingDisabledInSpec(mco))
+	if err != nil {
+		t.Fatalf("Failed to generate hub info secret with UWM alerting disabled: %v", err)
+	}
+	if hubInfo == nil {
+		t.Fatal("Generated hub info secret is nil")
+	}
+	hub = &operatorconfig.HubInfo{}
+	err = yaml.Unmarshal(hubInfo.Data[operatorconfig.HubInfoSecretKey], &hub)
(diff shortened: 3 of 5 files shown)


operators/multiclusterobservability/pkg/config/config.go
```diff
@@ -860,6 +861,16 @@ func IsAlertingDisabledInSpec(mco *observabilityv1beta2.MultiClusterObservabilit
 	return annotations != nil && annotations[AnnotationDisableMCOAlerting] == "true"
 }
 
+// Get AnnotationUWMAlerting.
+func IsUWMAlertingDisabledInSpec(mco *observabilityv1beta2.MultiClusterObservability) bool {
+	if mco == nil {
+		return false
```
coleenquadros: This should be true right if MCO is nil?

subbarao-meduri: I returned `false` because that is the default behavior (alert forwarding is enabled by default) and is consistent with handling in `IsAlertingDisabledInSpec` defined right above it . We would not call this method with a `null` mco, unless its from a unit test.

subbarao-meduri: Scenarios mco is could be nil (other than a contrived unit test where nil could be passed):
1. mco resource is not deployed. In this case, endpoint operator doesn't exist
2. mco resource is deleted. 
In either case, it doesn't seem to matter what is returned.  I have flipped it to be `true` just so that it is consistent with the desired state.

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
```diff
@@ -365,6 +365,14 @@ func (r *ObservabilityAddonReconciler) initFinalization(
 		if err != nil {
 			return false, err
 		}
+
+		// revert the change to user workload monitoring stack
+		err = RevertUserWorkloadMonitoringConfig(ctx, r.Client)
+		if err != nil {
+			r.Logger.Error(err, "Failed to revert user workload monitoring config")
+			// Don't return error here as we want to continue with other cleanup
```
jacobbaungard: Wondering why you made this decision? In all other cases we seem to return an error here. Shouldn't we be consistent?

Is a finalizer re-tried if it fails?

operators/multiclusterobservability/controllers/placementrule/hub_info_secret.go
```diff
@@ -14,14 +14,15 @@ import (
...
+	mcov1beta2 "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/api/v1beta2"
 	"github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/pkg/config"
 	operatorconfig "github.com/stolostron/multicluster-observability-operator/operators/pkg/config"
 )
 
 // generateHubInfoSecret generates the secret that contains hubInfo.
 // this function should only called when the watched resources are created/updated.
 func generateHubInfoSecret(client client.Client, obsNamespace string,
-	namespace string, ingressCtlCrdExists bool) (*corev1.Secret, error) {
+	namespace string, ingressCtlCrdExists bool, mco *mcov1beta2.MultiClusterObservability) (*corev1.Secret, error) {
```
jacobbaungard: Why don't we follow the same logic as for the platform alerting disablement?

I.e instead of passing in the whole MCO, we create a variable in the config, and use that instead of having to pass the whole MCO object here.

jacobbaungard: Maybe this new way is a bit simpler (requires less logic maybe), but would be best to handle them the same.

Maybe we could create a task to refactor this code so it is the same for both the uwm and platform alerting in 2.15?

operators/endpointmetrics/controllers/observabilityendpoint/observabilityaddon_controller.go
subbarao-meduri: It seems the finalizer is re-tried if it fails. So, returning error is the right thing. I will fix this. Thanks for catching.

jacobbaungard: Great. Maybe we should use a similar approach to the errors as done here, but we can do it in a follow up https://github.com/stolostron/multicluster-observability-operator/pull/2055/files#diff-6cbb01e6ad5cde2e9c29c793b0d59f7587b14d5eb30d0bb6c801947f42940246R486

operators/multiclusterobservability/controllers/placementrule/hub_info_secret.go
subbarao-meduri: @jacobbaungard addressed this through refactor

PR #2055, opened 2025-06-12
ACM-21373: fix addon status update

thibaultmg:
- Fixes a reconcile loop triggered by invalid resources cleaning in the placement rule controller for the hub
- Improves multi error handling
- Retry status update on conflict
- Improve conditions update

### DIFF
--- a/operators/multiclusterobservability/pkg/util/managedclusteraddon.go
+++ b/operators/multiclusterobservability/pkg/util/managedclusteraddon.go
@@ -6,13 +6,15 @@ package util
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"time"
 
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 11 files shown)


operators/multiclusterobservability/pkg/util/managedclusteraddon.go
```diff
@@ -45,108 +80,69 @@ func CreateManagedClusterAddonCR(c client.Client, namespace, labelKey, labelValu
...
-			log.Error(
-				errPoll,
-				"failed to get the created managedclusteraddon",
-				"name",
-				ManagedClusterAddonName,
-				"namespace",
-				namespace,
-			)
-			return nil, errPoll
+			desiredStatus.Conditions = append(desiredStatus.Conditions, newCondition)
```
coleenquadros, line 114: Do we always append the Addon Installing status everytime we update when its not available nor degraded?

thibaultmg: Honestly, I think we can completely remove this. But I didn't want to change too much the behavior, so I left it almost as it was doing it before. We would have been earlier in the release cycle, I would I removed this status update.

jacobbaungard (approved):
Reviewed, couldn't find any issues. Thanks for handling this!

@coleenquadros , can you add a `lgtm` if you are happy with this, too?

jacobbaungard (approved):
/lgtm

PR #2118, opened 2025-07-17
fix observatorium spec comparison

coleenquadros:


coleenquadros:
Tested locally multiple times  and the stateful set is updated

PR #2164, opened 2025-07-31
[ACM-18082] Fix ACM alerts threshold

dbuchanaRH:
Changes the following alerts to be based on a percentage of failing requests:
* `ACMMetricsCollectorFederationError`
* `ACMMetricsCollectorForwardRemoteWriteError`
* `ACMUWLMetricsCollectorFederationError`
* `ACMUWLMetricsCollectorForwardRemoteWriteError`

Noticed that the metrics collector would not record the `acm_metrics_collector_federate_requests_total` on any network errors. Added fix in `metricsclient.go` to account for this. When not able to reach prometheus instances because prometheus is not present or network misconfiguration, will record metric with **status code 0** and cause the `ACMMetricsCollectorFederationError`  to fire.

### DIFF
--- a/operators/endpointmetrics/pkg/collector/metrics_collector.go
+++ b/operators/endpointmetrics/pkg/collector/metrics_collector.go
@@ -472,7 +472,7 @@ func (m *MetricsCollector) ensureAlertingRule(ctx context.Context, isUWL bool) e
 								"summary":     "Error federating from in-cluster Prometheus.",
 								"description": "There are errors when federating from platform Prometheus",
 							},
-							Expr: intstr.FromString(`(sum by (name) (rate(` + replace + `federate_requests_total{status_code!~"2.*"}[10m]))) / (sum by (name) (rate(` + replace + `federate_requests_total[10m]))) > .2`),
+							Expr: intstr.FromString(`(sum(rate(` + replace + `federate_requests_total{status_code!~"2.*"}[10m]))) / (sum(rate(` + replace + `federate_requests_total[10m]))) > .2`),
 							For:  &forDuration,
 							Labels: map[string]string{
 								"severity": "critical",
@@ -484,7 +484,7 @@ func (m *MetricsCollector) ensureAlertingRule(ctx context.Context, isUWL bool) e
 								"summary":     "Error forwarding to Hub Thanos.",
 								"description": "There are errors when remote writing to Hub hub Thanos",
 							},
-							Expr: intstr.FromString(`(sum by (name) (rate(` + replace + `forward_write_requests_total{status_code!~"2.*"}[10m]))) / (sum by (name) (rate(` + replace + `forward_write_requests_total[10m]))) > .2`),
+							Expr: intstr.FromString(`(sum(rate(` + replace + `forward_write_requests_total{status_code!~"2.*"}[10m]))) / (sum(rate(` + replace + `forward_write_requests_total[10m]))) > .2`),
 							For:  &forDuration,
 							Labels: map[string]string{
 								"severity": "critical",


collectors/metrics/pkg/metricsclient/metricsclient.go
```diff
@@ -108,6 +108,10 @@ func (c *Client) RetrieveRecordingMetrics(
 	defer cancel()
 	families := make([]*clientmodel.MetricFamily, 0, 100)
 	err := withCancel(ctx, c.client, req, func(resp *http.Response) error {
+		if resp == nil {
+			c.metrics.FederateRequests.WithLabelValues("recording", "0").Inc()
```
saswatamcode: When would resp and err both be nil? The underlying call is client.Do in withCancel

operators/endpointmetrics/pkg/collector/metrics_collector.go
```diff
@@ -472,7 +472,7 @@ func (m *MetricsCollector) ensureAlertingRule(ctx context.Context, isUWL bool) e
 								"summary":     "Error federating from in-cluster Prometheus.",
 								"description": "There are errors when federating from platform Prometheus",
 							},
-							Expr: intstr.FromString(`(sum by (status_code, type) (rate(` + replace + `federate_requests_total{status_code!~"2.*"}[10m]))) > 10`),
+							Expr: intstr.FromString(`(sum by (name) (rate(` + replace + `federate_requests_total{status_code!~"2.*"}[10m]))) / (sum by (name) (rate(` + replace + `federate_requests_total[10m]))) > .2`),
```
saswatamcode: Hmm, so a couple of things with this,

- Why change the aggregating label to `name`? We instrument with only type and status_code here
- So this is checking over 10m. Our requests fire every 5m. So within a 10m window, even if one request fails, it will be over 50%. Do we want the alert to fire on every single failure?

jacobbaungard: > Why change the aggregating label to name? We instrument with only type and status_code here

If we aggregate on `status_code`, we'd get multiple timeseries for each non200 code, and I don't think we'd want to differentiate between i.e 500, 501, 502 in this alert?

We do separate federate requests for recording rules and `normal`, which is put into the `type` label (not true for the `forward_write` below, where I don't see `type` as a label at all). Does it make sense to differentiate between these in the alert?

> So this is checking over 10m. Our requests fire every 5m. So within a 10m window, even if one request fails, it will be over 50%. Do we want the alert to fire on every single failure?

The `for` is set to `10min` in the alert definition though, so it wouldn't actually fire until that state persists for a little while.

We seem to do multiple requests every 5 min (in my cluster the value seem to increase by ~17 in 5min chunks), although I assume it's unlikely only some of them fail.

collectors/metrics/pkg/metricsclient/metricsclient.go
dbuchanaRH: I changed withCancel because it exited when `err != nil` without checking if there was a resp. Originally, if there was an error and no response, it would exit the withCancel function prior to calling the passed in function (that records the metric). So if there is an error from the `client.Do(req)` in withCancel(), the FederateRequests metric would not be recorded. I don't think there is a case here where both resp and err are nil?

operators/endpointmetrics/pkg/collector/metrics_collector.go
saswatamcode: > If we aggregate on status_code, we'd get multiple timeseries for each non200 code, and I don't think we'd want to differentiate between i.e 500, 501, 502 in this alert?

Yup get that, and we might want to differentiate between the type. However where does the `name` label come from was more my question. 🙂 

> The for is set to 10min in the alert definition though, so it wouldn't actually fire until that state persists for a little while.

Ack missed the for field. So ~3 continuous federate request failures.

> (in my cluster the value seem to increase by ~17 in 5min chunks)

You mean 17 `normal` requests? Hmm, any chance you have a workers field set on the CR?

collectors/metrics/pkg/metricsclient/metricsclient.go
saswatamcode: Right, sorry missed the withCancel change there. 

But in that case, where there was an err while making the request with some partial response, we don't really need to count that as a meaningful request. A partial federate response can't really be used in our case 🙂

wdyt?

dbuchanaRH: Yeah I see what you mean, because the change to withCancel is ignoring that error as long as there is a response. I think the only thing that I am confused on is how we determine what a partial federate response is in this context?

saswatamcode: I don't think we should consider any partial responses here, and discard response on error. 
Federate endpoint or its consumers don't really have partial semantics normally, and effectively for most Prometheus protocols, the accepted solution is to just retry the request instead of ingesting partial data.

dbuchanaRH: Agreed we should just discard the response on error. I believe that is what we were doing originally, so I will just revert the changes for this file. Only changes on this MR will be to the alerting rules.

operators/endpointmetrics/pkg/collector/metrics_collector.go
jacobbaungard: > Yup get that, and we might want to differentiate between the type. However where does the name label come from was more my question. 🙂

Turns out when you put in a bogus/non-existing label it's functionally equivalent to a normal `sum` (without `by`). So I think we should just rewrite to just use the `sum` operator, unless you really think we should i.e sum by the status code or the type? I think you'd want to be notified no-mater the non-ok status code or type.

> You mean 17 normal requests? Hmm, any chance you have a workers field set on the CR?

Looks like it's one normal and the rest is for recording rules.

dbuchanaRH: The `type` for `acm_metrics_collector_federate_requests_total` seems to have a difference in recording rate between `normal` and `recording` metrics, so it may be helpful to group by type there. For `acm_metrics_collector_forward_write_requests_total` I agree that just `sum(rate(...` should suffice.

Screenshot From 2025-09-02 15-59-26 (Edited)

PR #2173, opened 2025-08-20
Introduce rs-utility package for shared right-sizing logic

dvandra:
- As part of adding virtualization right-sizing recommendation functionality, the current codebase has been refactored, and some common functionality has been moved into the rs-utility package, which can be utilized by both namespace and virtualization right-sizing recommendations.


thibaultmg:
@dvandra if you rebase on main, e2e ci should work now

thibaultmg (approved):
/lgtm

PR #2174, opened 2025-08-21
endpoint operator: avoid duplicate controller name

jacobbaungard:
This seems to crash otherwise with later versions of the controller-runtime.

ACM-23542


philipgough (approved):
/lgtm

PR #2184, opened 2025-08-29
ACM-18092: fix rbac-query-proxy

thibaultmg:
Fix of the issue through a full refactoring/rewrite of the rbac-query-proxy.
Highlights:

-   Reuses transport for mTLS connection to the observatorium-api across proxied requests.
-   Reuses transport for the connection to the API server across proxied requests.
-   Fixes some race conditions.
-   Fixes the rewriting of PromQL queries by using the PromQL parser, replacing regexes.
-   Improves health and readiness probes (removes the configmap script). It now becomes ready when the cache is synchronized and it manages to connect to the observatorium-api.
-   Removes constants for defining certificate paths, using flags instead.
-   Updates the Readme.
-   Fixes a race condition in integration tests.

The refactoring fixes the ACM-18092 bug by properly handling queries containing the `acm_managed_cluster_labels` metric, where the cluster name is contained in the `name` label key, while the `cluster` label key is used on all other metrics. This was not handled in the [getLabel](https://github.com/stolostron/multicluster-observability-operator/blob/40c77d8cdf936c069255c14b66ac255f2b332c9c/proxy/pkg/util/util.go#L795)  function.

This also includes fixes for certificate-related issues:

-   Mounts the proper CA certificate in the [proxy renderer](https://github.com/stolostron/multicluster-observability-operator/blob/40c77d8cdf936c069255c14b66ac255f2b332c9c/operators/multiclusterobservability/pkg/rendering/renderer_proxy.go#L109) where `observability-server-certs` was used instead of `observability-server-ca-certs`.
-   Stops relying on the `cert_controller` pod restart mechanism because of synchronization issues. In the e2e tests, we were testing the restart of the proxy, which occurred. However, following the addition of the readiness probe, the test failed because the restarted pod never reached the ready state, as it restarted with the old CA certificate and failed to connect to the Observatorium API server. Now the proxy automatically reloads the mTLS config by resolving the path symlinks every 15s. I implemented an event-based version that was more responsive, but the code was more complex, hence this simple and still lightweight solution.

One thing I haven't modified is the retrieval of the username. I don't think we need this, except for logs, which can be a valid reason to retrieve it. So I've kept it.

### DIFF
--- a/proxy/pkg/informer/informer.go
+++ b/proxy/pkg/informer/informer.go
@@ -143,11 +143,11 @@ func (i *ManagedClusterInformer) Run() {
 	klog.Info("Informer caches have successfully synced")
 
 	i.managedClustersMtx.RLock()
-	klog.Infof("Initial list of managed clusters: %v", slices.Sorted(maps.Keys(i.managedClusters)))
+	klog.V(1).Infof("Initial list of managed clusters: %v", slices.Sorted(maps.Keys(i.managedClusters)))
 	i.managedClustersMtx.RUnlock()
 
 	i.allowlistMtx.RLock()
-	klog.Infof("Initial regex label list: %v", i.inMemoryAllowlist.RegexLabelList)
+	klog.V(1).Infof("Initial regex label list: %v", i.inMemoryAllowlist.RegexLabelList)
 	i.allowlistMtx.RUnlock()
 
 	// Trigger a sync after the initial cache sync to process all existing resources.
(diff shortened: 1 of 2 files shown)


jacobbaungard (approved):
Went over the overarching changes here, and they look fine to me. I didn't do a super thorough line-by-line review though, but however tested the changes on a cluster, to check generally functionality of the rbac.

/lgtm

PR #2185, opened 2025-08-29
Add Konflux rebuild bot

jacobbaungard:
We need Konflux builds on a regular basis. However Konflux doesn't support regular/daily builds at the moment.

This commit adds a new GitHub workflow, which on a schedule, comments on the latest commit with a build command, in all of the relevant branches. This should ensure we get regular builds, even when there have been no activity for a while.

Since we migrated to the ACM common build pipeline for the konflux tasks, we will likely have various release branches with very low activity, necessitating this workflow.

### DIFF
--- a/.github/workflows/konflux-rebuild-bot.yaml
+++ b/.github/workflows/konflux-rebuild-bot.yaml
@@ -1,0 +1,15 @@
+name: Konflux periodic build bot
+on:
+  schedule:
+    - cron: '0 0 * * Sat' # Github Actions job run in UTC.
+  workflow_dispatch:
+
+jobs:
+  rebuild:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+      - name: Add comments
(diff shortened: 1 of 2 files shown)


jacobbaungard:
Example of what it will do: https://github.com/jacobbaungard/multicluster-observability-operator/commit/363741ed786ba52610404199bc0a984731d282a0#commitcomment-164842883

.github/workflows/konflux-rebuild-bot.yaml
```diff
@@ -0,0 +1,15 @@
+name: Konflux periodic build bot
+on:
+  schedule:
+    - cron: '0 0 * * Sat' # Github Actions job run in UTC.
```
coleenquadros, line 4: We can add on click too incase we need an urgent build

https://github.com/rhobs/syncbot/blob/c54b870a9ad18dc1a6223caa448d92bf2cb6a1d5/.github/workflows/acm-konflux-update-pr.yaml#L4

jacobbaungard: It's already there on line 5 below, or am I missing something?

PR #2194, opened 2025-09-11
Move rightsizing code under rightsizing controller

dvandra:
Rename Right-sizing controller to analytics controller and move right-sizing code under new controller from multiclusterobservability controller.


operators/multiclusterobservability/controllers/analytics/rightsizing/rightsizing_controller.go
```diff
@@ -2,15 +2,15 @@
 // Copyright Contributors to the Open Cluster Management project
```
rajusem, line 2: Should we move this file logic under `rs-utility` package? As we are not doing anything related to controller like we do for analytics_controller.go file.

PR #2197, opened 2025-09-11
ACM-20570 fix unbound service token

dbuchanaRH:
Will need to update here first, as Cluster role is managed by MCH -> https://github.com/stolostron/multiclusterhub-operator/blob/main/pkg/templates/charts/toggle/multicluster-observability-operator/templates/multicluster-observability-operator-clusterrole.yaml

Otherwise the kubeclient will not have permissions to successfully make API call to [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/)

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
+++ b/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
@@ -696,7 +696,6 @@ func generateAmAccessorTokenSecret(cl client.Client, kubeClient kubernetes.Inter
 		if hasExpiration && hasCreated {
 			// Check if the token is near expiration
 			expirationStr := string(expirationBytes)
-			expectedDuration := time.Duration(8640*3600) * time.Second
 			expiration, err := time.Parse(time.RFC3339, expirationStr)
 			if err != nil {
 				log.Error(err, "Failed to parse alertmanager accessor token expiration date", "expiration", expiration)
@@ -711,7 +710,7 @@ func generateAmAccessorTokenSecret(cl client.Client, kubeClient kubernetes.Inter
 				return nil, err
 			}
 
-			expectedDuration = expiration.Sub(created)
+			expectedDuration := expiration.Sub(created)
 			if expectedDuration <= 0 {
 				log.Error(nil, "Invalid duration for alertmanager accessor token", "duration", expectedDuration)
 				return nil, nil


operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -677,94 +681,59 @@ func deleteObject[T client.Object](ctx context.Context, c client.Client, obj T)
...
 // for the Alertmanager in the Hub cluster
-func generateAmAccessorTokenSecret(cl client.Client) (*corev1.Secret, error) {
-	amAccessorSA := &corev1.ServiceAccount{}
-	err := cl.Get(context.TODO(), types.NamespacedName{Name: config.AlertmanagerAccessorSAName,
-		Namespace: config.GetDefaultNamespace()}, amAccessorSA)
-	if err != nil {
-		log.Error(err, "Failed to get Alertmanager accessor serviceaccount", "name", config.AlertmanagerAccessorSAName)
-		return nil, err
-	}
+func generateAmAccessorTokenSecret(cl client.Client, kubeClient kubernetes.Interface) (*corev1.Secret, error) {
```
coleenquadros, line 685: any reason why we still want to keep cl client.Client if we are using kubernetes.Interface? We can continue using client.Client as its more generic

operators/multiclusterobservability/config/rbac/mco_role.yaml
```diff
@@ -26,6 +26,12 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
```
coleenquadros, line 29: do we have a PR for these chnages in the multicluster hub repo? I think we will need it there to be merged first

jacobbaungard: After https://github.com/stolostron/multicluster-observability-operator/pull/2210 merges MCH should automatically pull the changes.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
dbuchanaRH: We still want to keep client.Client to delete the previous token. I tried getting the deletion done by the kubernetes interface but there were problems with retrieving the service account and deleting the service token. For the method used to generate the new token, we have to make an API call to TokenRequest. Making the TokenRequest is far easier to do with kubernetes.Interface

operators/multiclusterobservability/controllers/placementrule/manifestwork_test.go
```diff
@@ -483,6 +531,25 @@ func TestManifestWork(t *testing.T) {
 		}
 	}
 
+	// Check that AlertmanagerAccessorSecret contains the token and expiration
+	for _, manifest := range found.Spec.Workload.Manifests {
+		obj := &unstructured.Unstructured{}
+		obj.UnmarshalJSON(manifest.Raw)
+		if obj.GetKind() == "Secret" && obj.GetName() == config.AlertmanagerAccessorSecretName {
+			if data, exists := obj.Object["data"]; exists {
```
thibaultmg, line 541: We could add a found flag and check it after the for loop. Just to make sure this test is effectively run.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -777,12 +746,37 @@ func generateAmAccessorTokenSecret(cl client.Client) (*corev1.Secret, error) {
 			Name:      config.AlertmanagerAccessorSecretName,
 			Namespace: spokeNameSpace,
 		},
+		Type: corev1.SecretTypeOpaque,
 		Data: map[string][]byte{
-			"token": tokenSrt.Data["token"],
+			"token":            []byte(tokenRequest.Status.Token),
+			"token-expiration": []byte(tokenRequest.Status.ExpirationTimestamp.Format(time.RFC3339)),
```
thibaultmg: I was expecting this to be in the annotation of the secret, which is more conventional. Any particular reason why you use the Data?

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -777,12 +746,37 @@ func generateAmAccessorTokenSecret(cl client.Client) (*corev1.Secret, error) {
...
 		},
 	}, nil
 }
 
+// deleteAlertmanagerAccessorTokenSecret deletes the previous unbound token secret
+func deleteAlertmanagerAccessorTokenSecret(ctx context.Context, cl client.Client) error {
```
thibaultmg, line 770: I think we should only delete it when it doesn't contain the expected expiration annotation (or data as currently implemented). Because this function is called from: PlacementRuleReconciler.Reconcile -> ensureMCOAResources -> generateAmAccessorTokenSecret -> deleteAlertmanagerAccessorTokenSecret

If MCOA is enabled, this will constantly delete and create on each reconcile loop.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
dbuchanaRH: I am assuming you are referring to the expiration date for the token? Having expiration in either annotations or data works. I'll get that changed to follow convention.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
dbuchanaRH: It may be helpful to check for the expiration before entering the delete function, but I don't think MCOA will constantly delete and recreate the token. There is a major difference between the old method of creating a service token, and creating a service token using TokenRequest where the old method we used requires the creation of a separate secret, whereas TokenRequest does not [ref](https://kubernetes.io/docs/concepts/security/service-accounts/#get-a-token). So the delete function only removes this secret storing the old token, and would have no effect on the the token created by TokenRequest. I'll need to change the name of that function and add comments because it is confusing

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -677,94 +681,59 @@ func deleteObject[T client.Object](ctx context.Context, c client.Client, obj T)
...
-				config.GetDefaultNamespace(),
-			)
-			tokenSrtName = secretName
 		}
 	}
 
-	tokenSrt := &corev1.Secret{}
-	err = cl.Get(context.TODO(), types.NamespacedName{Name: tokenSrtName,
-		Namespace: config.GetDefaultNamespace()}, tokenSrt)
+	amAccessorSA, err := kubeClient.CoreV1().ServiceAccounts(config.GetDefaultNamespace()).Get(
```
thibaultmg: We could avoid fetching this service account it seems as it's just used for logging?

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -677,94 +681,59 @@ func deleteObject[T client.Object](ctx context.Context, c client.Client, obj T)
...
-		err = fmt.Errorf("service account token not populated or empty: %s", config.AlertmanagerAccessorSAName)
-		log.Error(
-			err,
-			"no token present in Secret for Alertmanager accessor serviceaccount",
-			"service account name",
-			config.AlertmanagerAccessorSAName,
-			"secret name",
-			tokenSrtName,
-		)
+	// Starting with kube 1.24 (ocp 4.11), the k8s won't generate secrets any longer
```
thibaultmg: I would update this comment, removing the part about OCP 4.11 and just explain what we are doing to improve clarity.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
thibaultmg: oh ok, my bad. 👍

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -677,94 +681,59 @@ func deleteObject[T client.Object](ctx context.Context, c client.Client, obj T)
...
-					},
-				},
-				Type: "kubernetes.io/service-account-token",
-			}
-			err := cl.Create(context.TODO(), secret, &client.CreateOptions{})
-			if err != nil && !k8serrors.IsAlreadyExists(err) {
-				log.Error(err, "Failed to create token secret for Alertmanager accessor serviceaccount",
-					"name", config.AlertmanagerAccessorSAName)
-				return nil, err
+				percentOfExp := float64(time.Until(expiration)) / float64(expectedDuration)
```
thibaultmg: Is it working as expected?
If the apiServer delivers a token with a lifetime of 1w for example, here we will constantly be updating the token as 1w / 1y < 0.2. 
Maybe I'm missing something.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
thibaultmg: Maybe then we can add a comment to remove this function in 2.17 as this won't be needed anymore (we allow skip version upgrade, but we can only skip a single version I think)

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
dbuchanaRH: Good point so this does work as expected in my dev setup, but the situation where the api server changes the lifetime of the token is concerning. Could add an annotation to the returned secret with the token creation time ( TokenRequest doesn't have a CreationTimestamp like secrets do). Then we can just take the diff between the expiration and creation time as the expected duration of the token.

thibaultmg: Fix looks fine! 
But we are ignoring many errors. It would be better to make those checks sequential. And return or log the errors along the way. It will be difficult to debug if we have issues with it otherwise.
Nit: making the annotation keys constants would make the code more robust.

thibaultmg (approved):
/lgtm

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -677,97 +682,70 @@ func deleteObject[T client.Object](ctx context.Context, c client.Client, obj T)
...
-					Namespace: config.GetDefaultNamespace(),
-					Annotations: map[string]string{
-						"kubernetes.io/service-account.name": amAccessorSA.Name,
-					},
-				},
-				Type: "kubernetes.io/service-account-token",
+		if hasExpiration && hasCreated {
+			// Check if the token is near expiration
+			expirationStr := string(expirationBytes)
+			expectedDuration := time.Duration(8640*3600) * time.Second
```
jacobbaungard: This is overwritten below on line 714, without being used?

dbuchanaRH: Yes, was needed in my previous implementation but not anymore. Removing it should fix the lint error.

PR #2210, opened 2025-09-15
[ACM-20570] Updating rbac for amAccessorToken

dbuchanaRH:
Needed for https://github.com/stolostron/multicluster-observability-operator/pull/2197

Want to create bound tokens for the `observability-alertmanager-accessor` service account. Updating permissions to allow for API calls to [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/)

### DIFF
--- a/operators/multiclusterobservability/config/rbac/mco_role.yaml
+++ b/operators/multiclusterobservability/config/rbac/mco_role.yaml
@@ -27,6 +27,12 @@ rules:
   - update
   - watch
 - apiGroups:
+  - ""
+  resources:
+  - serviceaccounts/token
+  verbs:
+  - create
+- apiGroups:
   - apps
   resources:
   - deployments
--- a/operators/multiclusterobservability/manifests/base/alertmanager/alertmanager-accessor-serviceaccount.yaml
+++ b/operators/multiclusterobservability/manifests/base/alertmanager/alertmanager-accessor-serviceaccount.yaml
@@ -4,4 +4,3 @@ metadata:
   name: observability-alertmanager-accessor
   labels:
     alertmanager: observability
-automountServiceAccountToken: false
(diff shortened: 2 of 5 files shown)


operators/multiclusterobservability/config/rbac/mco_role.yaml
```diff
@@ -26,6 +26,12 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts/token
+  verbs:
+  - create
```
jacobbaungard, line 34: Don't we need something like `delete` or `update` maybe for when we do the rotation?  Or does the token request handle that automatically?

dbuchanaRH: So the token request does not handle the deletion of the previous token automatically. When we do the rotation, `amAccessorTokenSecret` is updated with the most recent secret, and the token used by the secret gets updated. Should probably add these permissions to remove expired tokens?

dbuchanaRH: Sorry, looked into this because I wasn't sure, and I don't think we need `delete` or `update`. There looks like there is no way to delete a TokenRequest via the kubernetes api. The only way to remove it is to delete or update the secret that it is bound to.

operators/multiclusterobservability/manifests/base/alertmanager/alertmanager-accessor-serviceaccount.yaml
```diff
@@ -4,3 +4,4 @@ metadata:
   name: observability-alertmanager-accessor
   labels:
     alertmanager: observability
+automountServiceAccountToken: false
```
jacobbaungard: I think we'd probably want this to be part of the main PR instead? I assume this would cause some problem unless merged with the rest of https://github.com/stolostron/multicluster-observability-operator/pull/2197 .

dbuchanaRH: Oh I overlooked this. Yes that would be problematic without the logic for the new service token. Will remove in next commit

PR #2217, opened 2025-10-01
Update spechash in ClusterManagementAddOn status

subbarao-meduri:
Related to [ACM-23742](https://issues.redhat.com/browse/ACM-23742)

This change implements spec hash mechanism prevents the "mca and work configs mismatch" error by ensuring ClusterManagementAddOn status accurately reflects current AddOnDeploymentConfig state. Adds unit and integration tests.

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
+++ b/operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
@@ -699,15 +699,21 @@ func createManagedClusterRes(ctx context.Context, c client.Client, mco *mcov1bet
 		addonConfig = defaultAddonDeploymentConfig
 	}
 
-	// Update the ClusterManagementAddOn status with the spec hash
-	// This prevents the "mca and work configs mismatch" error
-	if err := util.UpdateClusterManagementAddOnSpecHash(ctx, c, addonConfig); err != nil {
-		return nil, fmt.Errorf("failed to update spec hash: %w", err)
+	// Update the ClusterManagementAddOn status with the spec hash only when using the default config.
+	// CMA status only tracks default config hash, not custom configs.
+	// Only update if we have a valid default config (not empty struct).
+	if !isCustomConfig && defaultAddonDeploymentConfig != nil && defaultAddonDeploymentConfig.Name != "" {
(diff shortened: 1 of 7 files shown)


operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -683,6 +690,13 @@ func createManagedClusterRes(ctx context.Context, c client.Client, mco *mcov1bet
 		addonConfig = defaultAddonDeploymentConfig
 	}
 
+	// Update the ClusterManagementAddOn status with the spec hash
+	// This prevents the "mca and work configs mismatch" error
+	if err := util.UpdateClusterManagementAddOnSpecHash(ctx, c, addonConfig); err != nil {
```
thibaultmg: Why do we need to call the Update here? Isn't it covered by the call in setDefaultDeploymentConfigVar?

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
```diff
@@ -683,6 +690,13 @@ func createManagedClusterRes(ctx context.Context, c client.Client, mco *mcov1bet
...
+		log.Error(err, "Failed to update ClusterManagementAddOn spec hash", "namespace", namespace)
```
thibaultmg: We could avoid these error logs. Returned errors are logged at the top level when returned by the controller.

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
subbarao-meduri: setDefaultDeploymentConfigVar updates the hash only for the default AddOnDeploymentConfig. In  createManagedClusterRes, you may pick a per-addon custom AddOnDeploymentConfig via addon.Spec.Configs. That path is not covered by the default updater as far as I can see. Calling update ensures the managed cluster addon reflects the actual config used for that cluster. If the selected config ends up being the same as the default, the second call is effectively redundant but should be harmless.

operators/multiclusterobservability/controllers/placementrule/placementrule_controller.go
subbarao-meduri: Agree. Removed the extra log statement.

PR #2223, opened 2025-10-07
ACM-24791: update forward maxElapsedTime

thibaultmg:
This change stabilizes the remote-write backoff timing.

The previous maxElapsedTime calculation was dependent on the number of metric series and the interval duration, which could result in an overly short retry window under high load.
This new logic ensures a predictable and stable timeout for each chunk, improving reliability.

Concurrency is intentionally not added to keep it simple, as throughput can be scaled by enabling the existing multi-worker feature.

### DIFF
--- a/collectors/metrics/pkg/metricsclient/metricsclient.go
+++ b/collectors/metrics/pkg/metricsclient/metricsclient.go
@@ -40,8 +40,10 @@ import (
 )
 
 const (
-	nameLabelName   = "__name__"
-	maxSeriesLength = 10000
+	nameLabelName       = "__name__"
+	maxSeriesLength     = 10000
+	minRetryElapsedTime = 15 * time.Second
+	requestTimeout      = 10 * time.Second
 )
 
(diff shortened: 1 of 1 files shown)


collectors/metrics/pkg/metricsclient/metricsclient.go
```diff
@@ -532,7 +535,7 @@ func (c *Client) sendRequest(ctx context.Context, serverURL string, body []byte)
 		return backoff.Permanent(wrappedErr)
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, 15*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, requestTimeout)
```
jacobbaungard, line 538: Why do we lower this one to 10sec?

thibaultmg: I though it was enough and it was more coherent with the minRetryElapsedTime. But I can can revert as well.

PR #2224, opened 2025-10-07
ACM-24792: Add CRD resources for MCOA

thibaultmg:
Dependant CRDs for MCOA are now deployed when MCOA is enabled. 
CRDs are deployed using server-side apply to avoid conflicts with other operators like COO.
UnDeploy is modified to avoid deleting CRDs that we don't own.


operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -290,6 +290,25 @@ func (r *MultiClusterObservabilityReconciler) Reconcile(ctx context.Context, req
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to render MCO templates for %s/%s: %w", instance.GetNamespace(), instance.GetName(), err)
 	}
+
+	// Ensure dependencies are created first (like CRDs)
+	kindOrder := mcoconfig.KindOrder
+	defaultOrder := 100
+	sort.Slice(toDeploy, func(i, j int) bool {
```
jacobbaungard, line 298: Could you spell out (i.e in a comment) the logic here, was a bit hard to follow (for me anyway).

PR #2231, opened 2025-10-17
Add alertmanager url to mcoa config

thibaultmg:
This PR adds the alert manager url to the MCOA deployment config. This is needed for supporting alert forwarding on non OCP spokes.
It also removes the cluster label on rules defined for managed clusters. They are executed on the spokes' prometheus where the cluster label aggregation is not necessary. It can even result in no value when the cluster label is the only aggregate and is not defined. Like in `sum(machine_cpu_cores) by (cluster)`


thibaultmg:
Rebased, hopefully it helps with the kind test 🤞

PR #2237, opened 2025-10-31
Handle AM accessor token on redeploy/restart

jacobbaungard:
With this commit, we ensure that the AM accessor token is fetched from the existing secret if the global variable is empty, and we also make sure to clear this global variable Observability is removed.

Not doing this before caused the following sub-optimal behavior:
1) A new token is created every time MCO restarts, even if the actual token is not expired
2) If the Observability stack is re-deployed (i.e add the MCO CR, delete it again, then add it again), the MCO operator is not restarted and retains the information about the previous token. This us to try and create a new secret with the `resourceVersion` set, which led to an error, which caused the observability stack to not be created correctly.

We choose to retain the use of the global variable here, to ensure that we create as few as possible tokens. This is to handle the situation where a spoke might reconcile, and cause an a new token to be generated due to the expiry. In that case, if the hub reconciles closely after, if we didn't retain the global variable, we would once again create a new token. Further we currently do all expiry logic on the hub, so it's best to ensure the tokens are as aligned as possible.

Fixes: ACM-25717

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
+++ b/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
@@ -712,7 +712,31 @@ func generateAmAccessorTokenSecret(cl client.Client, kubeClient kubernetes.Inter
 		return nil, fmt.Errorf("kubeClient is required but was nil")
 	}
 
-	// Check expiration on amAccessorTokenSecret
+	// fetch any existing secrets from the hub-namespace if we didn't already
+	// have a secret saved in the global variable. We use a global variable
+	// because we might reconcile a spoke before the hub in which the spoke
+	// has a token which is newer than the one in the hub namespace. This is
+	// not strictly needed but we might as well create as few tokens as possible
+	if amAccessorTokenSecret == nil {
+		amAccessorTokenSecret = &corev1.Secret{}
+		err := cl.Get(context.TODO(),
+			types.NamespacedName{
+				Name:      config.AlertmanagerAccessorSecretName,
+				Namespace: config.GetDefaultNamespace(),
+			}, amAccessorTokenSecret)
+		if err != nil {
+			// If it's not found, we'll create a new one later
+			// we null the var here as it was set to an empty
+			// secret above, which is required for the Get function
+			if k8serrors.IsNotFound(err) {
+				amAccessorTokenSecret = nil
+			} else {
+				return nil, fmt.Errorf("unable to lookup existing alertmanager accessor secret. %w", err)
+			}
+		}
+	}
+
+	// If we managed to find an existing secret, check the expiration
 	if amAccessorTokenSecret != nil {
 		expirationBytes, hasExpiration := amAccessorTokenSecret.Annotations[amTokenExpiration]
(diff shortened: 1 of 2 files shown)


operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -712,8 +711,20 @@ func generateAmAccessorTokenSecret(cl client.Client, kubeClient kubernetes.Inter
 		return nil, fmt.Errorf("kubeClient is required but was nil")
 	}
 
-	// Check expiration on amAccessorTokenSecret
-	if amAccessorTokenSecret != nil {
+	// fetch any existing secrets
+	amAccessorTokenSecret := &corev1.Secret{}
+	err := cl.Get(context.TODO(),
+		types.NamespacedName{
+			Name:      config.AlertmanagerAccessorSecretName,
```
coleenquadros: are we not suffixing the hub name?

jacobbaungard: My understanding is:
- This is the "base" secret
- MCO use it to populate hubinfosecret (non mcoa mode), which the endpoint operator then use to generate the a separate accessor secret used in prometheus - this one has the hub name suffix
- MCOA uses this secret to similarly create secrets in the openshift-monitoring namespaces, also with the hub suffix.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -712,8 +711,20 @@ func generateAmAccessorTokenSecret(cl client.Client, kubeClient kubernetes.Inter
...
+			Namespace: spokeNameSpace,
```
thibaultmg: Shouldn't we use the hub namespace to fetch this secret?

jacobbaungard: hmmm. That would make sense, but we are later down the line in this function that creates the secret, we set it like this. I'm getting confused now, will double check.

jacobbaungard:
I had two different solutions in mind, the current one, and then one where the use of the global variable was removed.

I think, this one is best, because it has a better change to ensure that the token on the hub, aligns with the ones on the spoke, this especially on expiry cases. What could happen is if we didn't retain the global var state:

- spoke-2 reconciles (or this could be a new spoke)
- It sees the token is close to expiry, so creates a new one
- local-cluster might not reconcile straight away, so the actual token on the hub is not updated
- some time later local-cluster reconciles, it creates a new token, which has a later expiry date than the one on the spoke (if we use the global var, and MCO didn't restart, we would then use the same token as spoke-2 which is the solution I comitted here. This problem would still occur though, if MCO had restarted meanwhile)
- since we have to do all expiry handling hub side, if we are extremely unlucky, we might not realize that the token on spoke-2 expires, since we do the expiry handling based on the token on the hub

Now we should have enough margin for this to not be a problem, and, as noted, it still continue to be an issue if MCO restarts (since the global-var is then cleared).

Ideally we should keep the jwt token expiry information for each spoke separately, but that is likely a longer undertaking, so I suggest we keep it as is here, for now.

coleenquadros (approved):

PR #2240, opened 2025-11-11
Register observability addon registraton with framework in standard way

subbarao-meduri:
Related: [ACM-23742](https://issues.redhat.com/browse/ACM-23742)

Migrates the observability addon from manual specHash management to framework-native handling by declaring SupportedConfigGVRs in the addon agent. This enables automatic management of `AddOnDeploymentConfig` changes.

### DIFF
--- a/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
+++ b/operators/multiclusterobservability/controllers/placementrule/manifestwork.go
@@ -54,6 +54,16 @@ const (
 	amTokenCreated            = "token-created"
 )
 
+// managedManifestWorkAnnotations explicitly lists the annotations we own on ManifestWork.
+// We only update/remove annotations in this list, preserving all others to avoid
+// conflicts with addon-framework and other OCM controllers.
+//
+// When adding a new annotation to ManifestWork in createManifestWorks(), add it to this list.
+var managedManifestWorkAnnotations = []string{
+	workPostponeDeleteAnnoKey,                  // "open-cluster-management/postpone-delete"
+	workv1.ManifestConfigSpecHashAnnotationKey, // "open-cluster-management.io/config-spec-hash"
+}
+
 // intermediate resources for the manifest work.
 var (
 	hubInfoSecret             *corev1.Secret
@@ -246,7 +256,8 @@ func createManifestwork(ctx context.Context, c client.Client, work *workv1.Manif
 
 	log.Info("Updating manifestwork", "namespace", namespace, "name", name)
 	found.SetLabels(work.Labels)
-	found.SetAnnotations(work.Annotations)
+	// Only update annotations we manage, preserving annotations from other controllers
+	updateManagedAnnotations(found, work)
 	found.Spec.Workload.Manifests = work.Spec.Workload.Manifests
 	err = c.Update(ctx, found)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -464,6 +527,13 @@ func createManifestWorks(
 
 	work.Spec.Workload.Manifests = manifests
 
+	// Set the config spec hash annotation if the ManagedClusterAddOn has configReferences
+	configAnnotation, err := getConfigSpecHashAnnotation(context.TODO(), c, clusterNamespace)
```
thibaultmg: we can use the real context here

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -188,6 +246,7 @@ func createManifestwork(ctx context.Context, c client.Client, work *workv1.Manif
 
 	log.Info("Updating manifestwork", "namespace", namespace, "name", name)
 	found.SetLabels(work.Labels)
+	found.SetAnnotations(work.Annotations)
```
thibaultmg: Don't we risk having controllers fight for setting annotations? Shouldn't we just make sure the ones we need are present and keep the others?

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
```diff
@@ -205,6 +264,10 @@ func shouldUpdateManifestWork(desiredWork, foundWork *workv1.ManifestWork) bool
 		return true
 	}
 
+	if !maps.Equal(foundWork.Annotations, desiredWork.Annotations) {
```
thibaultmg: Relates to the comment above. If we just need to make sure our annotations are present this check should be updated.

operators/multiclusterobservability/pkg/util/managedclusteraddon.go
```diff
@@ -100,6 +193,24 @@ func updateManagedClusterAddOnStatus(ctx context.Context, c client.Client, names
 
 		desiredStatus := existingManagedClusterAddon.Status.DeepCopy()
 
+		// Log current state before modifications on retries
+		if retryAttempt > 1 {
```
thibaultmg: I don't understand this block. This looks like debug logging. Should we remove it?

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
subbarao-meduri: Fixed in the latest commit.

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
subbarao-meduri: Implemented explicit ownership model for ManifestWork annotations. Now we only update annotations we manage, preserve others from controllers. This should prevent reconciliation fights with addon-framework and other controllers

operators/multiclusterobservability/controllers/placementrule/manifestwork.go
subbarao-meduri: Fixed. Only compare annotations we own in shouldUpdateManifestWork(), and avoid triggering updates when other controllers modify their annotations.

operators/multiclusterobservability/pkg/util/managedclusteraddon.go
subbarao-meduri: Yes, I added this while diagnosing simultaneous updates to the resource. No longer needed, and is now removed.

PR #2269, opened 2025-12-11
[OBSINTA-848] Set right-sizing enabled by default

dvandra:
Set Right-sizing by default enabled:

- As we are going with GA for rightsizing recommendation solution in the next release, we will be enabling right-sizing out of the box while creating observability instance. So PR contains changes for the same

- Default enablement set using kubebuilder is adding all other nested defaults/empty parameters under capabilities in CRD as following:
```
capabilities:
    platform:
      analytics:
        incidentDetection: {}
        namespaceRightSizingRecommendation:
          enabled: true
        virtualizationRightSizingRecommendation:
          enabled: true
      logs:
        collection: {}
      metrics:
        default: {}
        ui: {}
```

- Thus we have to go with this preferred way(in this PR) to set Default enablement for Right-sizing recommendation, it will only update right-sizing related parameters in CRD:
```
capabilities:
    platform:
      analytics:
        namespaceRightSizingRecommendation:
          enabled: true
        virtualizationRightSizingRecommendation:
          enabled: true
```

### DIFF
--- a/operators/multiclusterobservability/controllers/analytics/analytics_controller.go
+++ b/operators/multiclusterobservability/controllers/analytics/analytics_controller.go
@@ -98,14 +98,14 @@ func (r *AnalyticsReconciler) ensureRightSizingDefaults(ctx context.Context, ins
 			// Build a minimal patch that only contains the analytics fields we want to set.
 			// Use typed locals to avoid chained type assertions (which can panic if the shape changes).
 			// Set true if not present else preserve existing value
-			analytics := map[string]interface{}{
-				"namespaceRightSizingRecommendation":      map[string]interface{}{"enabled": !nsFound || nsEnabled},
-				"virtualizationRightSizingRecommendation": map[string]interface{}{"enabled": !virtFound || virtEnabled},
+			analytics := map[string]any{
+				"namespaceRightSizingRecommendation":      map[string]any{"enabled": !nsFound || nsEnabled},
+				"virtualizationRightSizingRecommendation": map[string]any{"enabled": !virtFound || virtEnabled},
 			}
-			patchData := map[string]interface{}{
-				"spec": map[string]interface{}{
-					"capabilities": map[string]interface{}{
-						"platform": map[string]interface{}{
+			patchData := map[string]any{
+				"spec": map[string]any{
+					"capabilities": map[string]any{
+						"platform": map[string]any{
 							"analytics": analytics,
 						},
 					},
--- a/operators/multiclusterobservability/controllers/analytics/analytics_controller_test.go
+++ b/operators/multiclusterobservability/controllers/analytics/analytics_controller_test.go
@@ -8,19 +8,19 @@ import (
 	"context"
 	"testing"
 
+	mcov1beta2 "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/api/v1beta2"
+	rsnamespace "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/controllers/analytics/rightsizing/rs-namespace"
+	rsutility "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/controllers/analytics/rightsizing/rs-utility"
+	rsvirtualization "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/controllers/analytics/rightsizing/rs-virtualization"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client/fake"
-	mcov1beta2 "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/api/v1beta2"
-	rsnamespace "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/controllers/analytics/rightsizing/rs-namespace"
-	rsutility "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/controllers/analytics/rightsizing/rs-utility"
-	rsvirtualization "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/controllers/analytics/rightsizing/rs-virtualization"
 	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
 	policyv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
 func setupTestScheme(t *testing.T) *runtime.Scheme {
--- a/tests/pkg/tests/observability_right_sizing_test.go
+++ b/tests/pkg/tests/observability_right_sizing_test.go
@@ -13,7 +13,6 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/stolostron/multicluster-observability-operator/tests/pkg/utils"
-
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -23,9 +22,7 @@ import (
 )
 
 var _ = Describe("Right-sizing: defaults are enabled on fresh install", Ordered, func() {
-	var (
-		mcoGVR = utils.NewMCOGVRV1BETA2()
-	)
+	mcoGVR := utils.NewMCOGVRV1BETA2()
 
 	BeforeAll(func() {
 		// initialize clients once (if not already)


rajusem:
@dvandra Can we add E2E for the observability installation without rightsizing property, as with new approach it should enable rightsizing, so lets cover that also in E2E scenario.

rajusem:
I guess e2e failure will be fixed one https://github.com/stolostron/multicluster-observability-operator/pull/2271 merged else lgtm.

operators/multiclusterobservability/controllers/analytics/analytics_controller.go
```diff
@@ -25,6 +29,12 @@ import (
 
 var log = logf.Log.WithName("controller_rightsizing")
 
+var mcoGVK = schema.GroupVersionKind{
```
rajusem: Lets use NewMCOGVRV1BETA2 function here.

operators/multiclusterobservability/controllers/analytics/analytics_controller_test.go
```diff
@@ -54,13 +58,45 @@ func newTestMCO(binding string, enabled bool, paused bool) *mcov1beta2.MultiClus
...
+
+	persisted := &mcov1beta2.MultiClusterObservability{}
+	err = c.Get(context.TODO(), types.NamespacedName{Name: mco.GetName()}, persisted)
+	require.NoError(t, err)
+	require.True(t, persisted.Spec.Capabilities.Platform.Analytics.NamespaceRightSizingRecommendation.Enabled)
+	require.True(t, persisted.Spec.Capabilities.Platform.Analytics.VirtualizationRightSizingRecommendation.Enabled)
+}
+
+func TestEnsureRightSizingDefaultsPreservesExistingFlags(t *testing.T) {
+	t.Skip("Fake client conversion can omit explicit false due to omitempty; keep tests minimal and non-flaky")
```
rajusem: What we wanted to achieve in this test case? Do we need this test?

operators/multiclusterobservability/controllers/analytics/analytics_controller.go
```diff
@@ -68,6 +84,80 @@ func (r *AnalyticsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			}
+
+			// Use MergePatch to only update the specific fields without affecting others
+			if err := r.Client.Patch(ctx, u, client.RawPatch(types.MergePatchType, patchBytes)); err != nil {
+				return instance, fmt.Errorf("failed to persist default analytics right-sizing flags: %w", err)
+			}
+			reqLogger.Info("Defaulted analytics right-sizing flags to true (fresh install)")
+
+			// refresh typed instance so downstream logic sees updated flags
+			refreshed := &mcov1beta2.MultiClusterObservability{}
```
rajusem, line 127: Probably we can add error logs for err, something like below. 
```
refreshed := &mcov1beta2.MultiClusterObservability{}
  if err := r.Client.Get(ctx, key, refreshed); err != nil {
      reqLogger.Error(err, "Failed to refresh MCO after patching defaults, using stale instance")
  } else {
      instance = refreshed.DeepCopy()
  }
  ```

File: operators/multiclusterobservability/controllers/analytics/analytics_controller.go [hunk 4]
```diff
@@ -68,6 +84,80 @@ func (r *AnalyticsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+				"spec": map[string]interface{}{
+					"capabilities": map[string]interface{}{
+						"platform": map[string]interface{}{
+							"analytics": map[string]interface{}{},
+						},
+					},
+				},
+			}
+
+			analytics := patchData["spec"].(map[string]interface{})["capabilities"].(map[string]interface{})["platform"].(map[string]interface{})["analytics"].(map[string]interface{})
```
rajusem: Can you try replacing this logic with below code, tried sorter version of current if/else logic.   
```
analytics := patchData["spec"].(map[string]interface{})["capabilities"].(map[string]interface{})["platform"].(map[string]interface{})["analytics"].(map[string]interface{})

// Set true if not present else preserve existing value 
analytics["namespaceRightSizingRecommendation"] = map[string]interface{}{"enabled": !nsFound || nsEnabled}
analytics["virtualizationRightSizingRecommendation"] = map[string]interface{}{"enabled": !virtFound || virtEnabled}
```

operators/multiclusterobservability/controllers/analytics/analytics_controller.go
```diff
@@ -68,6 +79,63 @@ func (r *AnalyticsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
```
tremes: This is very difficult to read for me. These are type assertions/conversions right? You know that it can panic right?

dvandra: Makes sense.
What do you think about going like this way:

```
analytics := map[string]interface{}{
   "namespaceRightSizingRecommendation":    map[string]interface{}{"enabled": !nsFound || nsEnabled},
   "virtualizationRightSizingRecommendation": map[string]interface{}{"enabled": !virtFound || virtEnabled},
 }
 
 patchData := map[string]interface{}{
   "spec": map[string]interface{}{
     "capabilities": map[string]interface{}{
       "platform": map[string]interface{}{
         "analytics": analytics,
       },
     },
   },
 }
```

tremes: Would something like following work?

```
			analytics := mcov1beta2.PlatformAnalyticsSpec{}
			patchData := &mcov1beta2.MultiClusterObservabilitySpec{
				Capabilities: &mcov1beta2.CapabilitiesSpec{
					Platform: &mcov1beta2.PlatformCapabilitiesSpec{
						Analytics: analytics,
					},
				},
			}
			analytics.NamespaceRightSizingRecommendation.Enabled = !nsFound || nsEnabled
			analytics.VirtualizationRightSizingRecommendation.Enabled = !virtFound || virtEnabled
```

dvandra: Not working.

tremes: OK probably because of `RawPatch` (as we discussed in Slack) :thinking:

operators/multiclusterobservability/controllers/analytics/analytics_controller.go
```diff
@@ -67,6 +78,63 @@ func (r *AnalyticsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			// Use MergePatch to only update the specific fields without affecting others
+			if err := r.Client.Patch(ctx, u, client.RawPatch(types.MergePatchType, patchBytes)); err != nil {
+				return instance, fmt.Errorf("failed to persist default analytics right-sizing flags: %w", err)
+			}
+			reqLogger.Info("Defaulted analytics right-sizing flags to true (fresh install)")
+
+			// refresh typed instance so downstream logic sees updated flags
+			refreshed := &mcov1beta2.MultiClusterObservability{}
+			if err := r.Client.Get(ctx, key, refreshed); err != nil {
+				reqLogger.Error(err, "Failed to refresh MCO after patching defaults, using stale instance")
```
thibaultmg, line 129: You don't return the error here? No need to retry patching?

operators/multiclusterobservability/controllers/analytics/analytics_controller.go
```diff
@@ -67,6 +78,63 @@ func (r *AnalyticsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+func (r *AnalyticsReconciler) ensureRightSizingDefaults(ctx context.Context, instance *mcov1beta2.MultiClusterObservability, reqLogger logr.Logger) (*mcov1beta2.MultiClusterObservability, error) {
+	// Default-enable analytics right-sizing flags ONLY when absent on fresh installs.
+	// Persist defaults back to the MCO spec so users can later override to true/false explicitly.
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(mcoGVK)
+	key := types.NamespacedName{Name: instance.GetName()}
+
+	if err := r.Client.Get(ctx, key, u); err == nil {
+		// Check if the fields already exist
+		nsEnabled, nsFound, _ := unstructured.NestedBool(u.Object,
```
thibaultmg, line 91: why are you using unstructured here? It's also about not setting the default fields in capabilities?

dvandra: unstructured is used to detect whether the enabled fields are missing from the original YAML (so they can be defaulted to true on fresh installs) versus explicitly set to false (which should be preserved), since the typed struct can't distinguish between missing fields and fields set to their zero value (false).

operators/multiclusterobservability/controllers/analytics/analytics_controller.go
dvandra: The error isn't returned because the patch already succeeded (line 121), so the defaults are persisted in etcd. The Get() on line 128 is only a refresh read to update the local instance for downstream use. If the refresh fails, we continue with the stale instance; the resource in etcd is correct and will be picked up on the next reconcile. Failing the reconcile here would be unnecessary since the write already succeeded.

Let me know if it makes sense here.

operators/multiclusterobservability/controllers/analytics/analytics_controller.go
thibaultmg: Ack, makes sense thanks

PR #2283, opened 2026-01-07
[ACM-27841] Cleanup CMO  that contain stale AM secret configs post GH

coleenquadros:
https://issues.redhat.com/browse/ACM-27841

Post Global Hub changed for Alertmanager secrets. We need to get clear the CMO config that contains the AM secrets created previously. They are of the format 

- hub-alertmanager-router-ca
-  hub-alertmanager-router-ca -{clusterDomain}
- observability-alertmanager-accessor
- observability-alertmanager-accessor-{clusterDomain}

We should also delete these secrets from the openshift monitoring ns


thibaultmg (approved):
/lgtm

PR #2285, opened 2026-01-07
[ACM-27632] Reenable e2e test

dbuchanaRH:
Reenable e2e test that was commented out.
- Add delay after label added to managed cluster.
- Trigger a second reconcile.


thibaultmg (approved):
Nice!
/lgtm

PR #2295, opened 2026-01-14
Remove uiplugins from mcoa required crds

thibaultmg:
The uiplugins resource is not a configuration type needed by the ClusterManagementAddon.
It make the CI fail as the addon manager status is set as degraded because it's missing CRD.

### DIFF
--- a/operators/multiclusterobservability/pkg/rendering/renderer_mcoa.go
+++ b/operators/multiclusterobservability/pkg/rendering/renderer_mcoa.go
@@ -36,6 +36,7 @@ const (
 	nameUserWorkloadMetricsCollection = "userWorkloadMetricsCollection"
 	nameMetricsHubHostname            = "metricsHubHostname"
 	nameMetricsAlertManagerHostname   = "metricsAlertManagerHostname"
+	namePLatformMetricsUI             = "platformMetricsUI"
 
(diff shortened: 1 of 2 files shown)


operators/multiclusterobservability/pkg/rendering/renderer_mcoa.go
```diff
@@ -239,15 +239,6 @@ func (r *MCORenderer) renderAddonDeploymentConfig(
 			if cs.Platform.Metrics.Default.Enabled {
 				fqdn := mcoconfig.GetMCOASupportedCRDFQDN(mcoconfig.PrometheusAgentCRDName)
 				appendCustomVar(aodc, namePlatformMetricsCollection, fqdn)
-
-				if cs.Platform.Metrics.UI.Enabled {
-					fqdn := mcoconfig.GetMCOASupportedCRDFQDN(mcoconfig.UIPluginsCRDName)
```
coleenquadros, line 243: Are we planning to enable UI  by default when metrics collection is enabled ?

thibaultmg: No, nothing is planned there. I'm just removing the uiplugins type from the supported config type by the clusterManagementAddon resource.

coleenquadros (approved):

PR #2302, opened 2026-01-16
Add linter checks

thibaultmg:
Adding a few more linters. Notably, gocognit is added with a high threshold to cap complex functions. It lead to the refactor of a few of the biggest ones in the code base.
Reviewing commit by commit is probably the preferred approach here.

### DIFF
--- a/collectors/metrics/pkg/forwarder/forwarder.go
+++ b/collectors/metrics/pkg/forwarder/forwarder.go
@@ -45,7 +45,7 @@ type Config struct {
 	// StatusClient is a kube client used to report status to the hub.
 	StatusClient client.Client
 	Logger       log.Logger
-	Metrics      *workerMetrics
+	Metrics      *WorkerMetrics
 
 	// FromClientConfig is the config for the client used in sending /federate requests to Prometheus.
 	FromClientConfig FromClientConfig
@@ -103,9 +103,9 @@ type ToClientConfig struct {
 // CreateFromClient creates a new metrics client for the from URL.
 // Needs to be exported here so that it can be used in collectrule evaluator.
 func (cfg Config) CreateFromClient(
-	metrics *workerMetrics,
+	metrics *WorkerMetrics,
 	interval time.Duration,
-	name string,
+	_ string,
 	logger log.Logger,
 ) (*metricsclient.Client, error) {
 	fromTransport := metricsclient.DefaultTransport(logger)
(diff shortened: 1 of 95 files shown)


collectors/metrics/pkg/forwarder/forwarder.go
```diff
@@ -103,9 +103,10 @@ type ToClientConfig struct {
 // CreateFromClient creates a new metrics client for the from URL.
 // Needs to be exported here so that it can be used in collectrule evaluator.
 func (cfg Config) CreateFromClient(
-	metrics *workerMetrics,
+	ctx context.Context,
+	metrics *WorkerMetrics,
 	interval time.Duration,
-	name string,
+	_ string,
```
jacobbaungard, line 109: I guess we could remove the arg completely?

operators/multiclusterobservability/api/v1beta1/multiclusterobservability_conversion.go
```diff
@@ -29,48 +29,48 @@ ConvertTo is expected to modify its argument to contain the converted object.
 Most of the conversion is straightforward copying, except for converting our changed field.
 */
 // ConvertTo converts this MultiClusterObservability to the Hub version (v1beta2).
-func (src *MultiClusterObservability) ConvertTo(dstRaw conversion.Hub) error {
+func (m *MultiClusterObservability) ConvertTo(dstRaw conversion.Hub) error {
```
jacobbaungard, line 32: Not sure about that one, `src` seems more descriptive to me (although we should get rid v1beta1 soon).

operators/multiclusterobservability/pkg/certificates/cert_controller.go
```diff
@@ -217,25 +218,25 @@ func onUpdate(c client.Client, ingressCtlCrdExists bool) func(oldObj, newObj any
...
-				removeExpiredCA(c, newS.Name)
+				removeExpiredCA(context.Background(), c, newS.Name)
 			}
 			if needsRenew(newS) {
 				var err error
 				var hosts []string
 				switch name := newS.Name; name {
 				case serverCACerts:
-					err, _ = createCASecret(c, nil, nil, true, serverCACerts, serverCACertifcateCN)
+					_, err = createCASecret(context.Background(), c, nil, nil, true, serverCACerts, serverCACertifcateCN)
```
jacobbaungard, line 231: I think the `createCASecret` is kinda borked, in that when it returns an error object it does i.e:

```
if err != nil {
    return err
}
```

Rather than

```
return nil, err
```

So I think this change will break stuff?

PR #2375, opened 2026-03-12
[ACM-30763] - Enhance handling storageclass detection in MCO CR

coleenquadros:
## Summary by CodeRabbit

* **New Features**
  * Detects storage-class changes and automatically removes/recreates affected components so they use the new class.

* **Bug Fixes**
  * Removed implicit/default storage-class values from CRD to avoid unintended defaults.
  * Fixed certificate host handling to avoid adding empty hosts.

* **Documentation**
  * Added a warning documenting that changing storage class is destructive and listing mitigation steps.

* **Tests**
  * Added tests for storage-class selection and change-handling.


operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -77,6 +77,7 @@ var (
 	isRuleStorageSizeChanged         = false
 	isReceiveStorageSizeChanged      = false
 	isStoreStorageSizeChanged        = false
+	isStorageClassChanged            = false
```
thibaultmg: We should avoid using global variables. We can detect this change directly in HandleStorageSizeChange by checking the current state and compare with the mco value.

operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -687,9 +698,48 @@ func (r *MultiClusterObservabilityReconciler) HandleStorageSizeChange(
...
 	return nil, nil
 }
 
+func deleteStatefulSetsAndPVCs(c client.Client, matchLabels map[string]string) error {
+	stsList, err := commonutil.GetStatefulSetList(c, config.GetDefaultNamespace(), matchLabels)
+	if err != nil {
+		return err
+	}
+	for index, sts := range stsList {
+		if err := c.Delete(context.TODO(), &stsList[index]); err != nil && !apierrors.IsNotFound(err) {
```
thibaultmg: We should forward the context instead of using TODO

operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go
```diff
@@ -687,9 +698,48 @@ func (r *MultiClusterObservabilityReconciler) HandleStorageSizeChange(
 			return &reconcile.Result{}, err
 		}
 	}
+	if isStorageClassChanged {
+		isStorageClassChanged = false
+		for _, component := range []map[string]string{
+			{"observability.open-cluster-management.io/name": mco.GetName(), "alertmanager": "observability"},
```
thibaultmg, line 701: This deletion is destructive as it removes sliences. We should add a warning in the documentation.

thibaultmg (dismissed):

jacobbaungard (commented):
Looks good overall, just wondering one thing. `gp2` was added as default before. Now it is optional.

There will be a lot of systems out-there with `gp2` as set, even if they don't actually have that storage class. As far as I know, before in this case, the behaviour was, that we used the default storage class on the cluster. Just double checking that we are not changing this behavior here (I think not, but wanted to double check to avoid breaking issues).

coleenquadros:
@jacobbaungard yes there is a test case for this behavior that picks the default storage class -> https://github.com/stolostron/multicluster-observability-operator/pull/2375/changes#diff-d9263e34500c5119dec5a736dae66c906db39f90deaa5c19b3dcfeb1bfeb35f1R1148-R1149

jacobbaungard (approved):
/lgtm

jacobbaungard (approved):
/lgtm

PR #2401, opened 2026-04-08
[release-2.13] Adding proxy for Grafana tests

dbuchanaRH:
* Adding proxy to query grafana and grafana dashboards  in qe environment in 2.13
* Added proxy to alert and route tests where external requests are used


jacobbaungard (approved):
/lgtm

PR #2444, opened 2026-05-05
fix(deployer): eliminate infinite reconciliation loops caused by semantic equality failures

thibaultmg:
### What this PR does / why we need it:

This PR resolves a set of infinite reconciliation loops in the `Deployer` component that were silently consuming operator CPU and generating excessive API server traffic. The loops were caused by subtle failures in `equality.Semantic.DeepDerivative` across two distinct resource types:

**1. Workload Specs (Deployments, StatefulSets, Services)**
When parsed from Kustomize templates, missing fields in typed Go structs (e.g., `TimeoutSeconds: 0`) are zero-initialized. When compared against the runtime state, the cluster injects defaults (e.g., `timeoutSeconds: 1`). Because custom equality functions often bypass `DeepDerivative`'s zero-value-ignore logic, the operator continually detected a false drift. 
**Fix:** We now evaluate semantic equality against the `unstructured` maps for the `Spec` fields. This preserves the "missing key" state, allowing `DeepDerivative` to correctly ignore cluster-side defaults that are not explicitly managed by our manifests.

**2. Secrets with `StringData`**
Historically, secrets generated with `stringData` (like `grafana-config`) caused an infinite loop because their Go struct `Data` property evaluates to `nil`. A legacy hack attempted to bypass this by unconditionally forcing an update if `desiredSecret.Data == nil`. The Kube API accepted the update, converted the `stringData` to `data`, detected no actual byte changes, and dropped the update without bumping the `ResourceVersion`, leaving the operator in a perpetual "silent" retry loop.
**Fix:** We removed the hack and now explicitly convert `StringData` to `Data` in memory prior to the `DeepDerivative` check, accurately aligning our desired state representation with the API server's runtime representation.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable update detection by using generation-based checks.
  * Secret handling improved to avoid unnecessary update loops when server converts stringData.

* **Refactor**
  * Unified, map-based comparison and metadata-merge approach for Deployments, StatefulSets, and Services to preserve user-managed fields while applying desired changes.

* **Tests**
  * Expanded and adjusted tests covering label-merge behavior, volumeClaimTemplates, and secret update scenarios.

### DIFF
--- a/operators/multiclusterobservability/controllers/multiclusterobservability/predicate_func.go
+++ b/operators/multiclusterobservability/controllers/multiclusterobservability/predicate_func.go
@@ -27,7 +27,7 @@ func GetMCOPredicateFunc() predicate.Funcs {
 			checkStorageChanged(e.ObjectOld.(*mcov1beta2.MultiClusterObservability).Spec.StorageConfig,
 				e.ObjectNew.(*mcov1beta2.MultiClusterObservability).Spec.StorageConfig)
 			config.SetMonitoringCRName(e.ObjectNew.GetName())
-			return e.ObjectOld.GetResourceVersion() != e.ObjectNew.GetResourceVersion()
+			return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
 			return !e.DeleteStateUnknown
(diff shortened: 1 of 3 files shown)


jacobbaungard (approved):
/lgtm

PR #2404, opened 2026-04-09
[ACM-32844] Add right-sizing delegation from MCO to MCOA via MCO CR annotation

dvandra:
## Summary
                                                                                                                                                                                          
  Adds a migration path for right-sizing (RS) recommendations from MCO's
  Policy-based approach to MCOA's ManifestWork-based approach. This enables                                                                                                               
  RS to be managed by MCOA when the MCO CR is annotated with               
  `observability.open-cluster-management.io/right-sizing-capable: "v1"`.                                                                                                                  
                                                                                                                                                                                          
  **Default behavior is unchanged** — RS continues to use MCO's Policy-based                                                                                                              
  deployment. MCOA mode is opt-in and requires an admin to manually annotate                                                                                                              
  the MCO CR.                                                                                                                                                                             
                                                                                                                                                                                          
  ### Key changes                                                                                                                                                                         
                                                                                                                                                                                          
  - **Analytics controller**: New `AnalyticsReconciler` that watches the MCO CR                                                                                                           
    and handles RS lifecycle based on delegation mode:                                                                                                                                    
    - **MCO mode** (default): Creates Policy/Placement/PlacementBinding resources
      and syncs "disabled" to AddOnDeploymentConfig (ADC) so MCOA does not deploy RS                                                                                                      
    - **MCOA mode** (annotated): Cleans up Policy resources, syncs actual RS                                                                                                              
      enabled/disabled state to ADC so MCOA deploys via ManifestWork                                                                                                                      
  - **Migration gate**: `IsRightSizingDelegated()` reads the MCO CR annotation                                                                                                            
    as the authoritative signal for which component manages RS                                                                                                                            
  - **ADC sync**: Ensures `platformNamespaceRightSizing` and                                                                                                                              
    `platformVirtualizationRightSizing` variables in ADC reflect the correct                                                                                                              
    state, preventing dual deployment by both MCO and MCOA                                                                                                                                
  - **Renderer integration**: MCOA templates (ADC, Deployment, CMA) are rendered                                                                                                          
    when `MCOAEnabled || rightSizingDelegated`, ensuring MCOA infrastructure                                                                                                              
    exists even when only RS is delegated             

### How it works                                                                                                                                                                        
                                                                                                                                                                                          
  MCO CR annotation absent → MCO mode (Policy-based RS, ADC="disabled")                                                                                                                   
  MCO CR annotation present → MCOA mode (cleanup Policies, ADC="enabled")                                                                                                                 
                                                                                                                                                                                          
[...]

PR #2445, opened 2026-05-05
[ACM-32844] Add right-sizing delegation from MCO to MCOA via MCO CR a…

dvandra:
…nnotation (#2404)

* Add right-sizing delegation from MCO to MCOA via MCO CR annotation

* Reset ComponentState after delegation cleanup and add freshEnable detection

* refactor: extract RS constants, simplify delegation check, fix ADC sync and renderer

* addressed reviews and comments

* address review and comments

* fix: propagate cleanup errors and derive namespaces from MCO CR instead of stale process-local state

* fix SpecHash

* Fix stale SpecHash detection in MCA ConfigReferences

* Fix placement orphaning during MCO deletion and configmap predicate delegation check

* restore Grafana launch-link annotation on MCOA CMA

The Grafana link was removed from the MCOA ClusterManagementAddOn assuming it caused duplicate non-clickable entries in the ACM console. The actual root cause is in stolostron/console: AcmDropdown does not pass href to MenuItem, and AcmLaunchLink has a no-op onSelect handler for the multi-link dropdown path.

Restore the launch-link annotation so the MCOA CMA links to its own Grafana dashboard (UID 89eaec849a6e4837a619fb0540c22b13).

* Fix ADC overwrite race during MCO deletion with two-phase stabilization

Replace cleanupStarted bool with cleanupAt timestamp to block all reconciles during stabilization window. Phase 2 re-syncs disabled state to ADC before cleanup to handle main controller overwrite race.

* check if rs is enabled for renderer

* fix timeout issue

* fix e2e for rs-virt cleanup

* remove unrelated changes

---------


thibaultmg (approved):
/lgtm

PR #2447, opened 2026-05-06
[RFE-9238] Restrict right-sizing policies to OpenShift clusters only

dvandra:
Restrict right-sizing policy Placement defaults to OpenShift clusters only (vendor=OpenShift) so that the rs-prom-rules-policy and rs-virt-prom-rules-policy policies are not deployed to non-OpenShift clusters (e.g. AKS) that lack the openshift-monitoring namespace and Prometheus CRDs.

## Summary by CodeRabbit

* **Configuration Updates**
  * Rightsizing analytics default placement now targets OpenShift clusters exclusively.


thibaultmg (approved):
/lgtm

PR #2451, opened 2026-05-07
[main] Proxy: get user from JWT

jacobbaungard:
This commit allows the proxy to get a user ID (subject/uuid) from the
JWT auth token, if we fail to find the username from any other methods.

When external OIDC, we cannot query the OCP api for the username, since
that api type doesn't exist in this case.

The JWT token contain a few different fields that can be used, but we
use the `subject` as other fields such as `email` or
`preferred_username` are less likely to be guaranteed to exist.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved username resolution for proxied requests with a prioritized fallback (cache → cluster lookup → token extraction), reducing authentication failures.
  * Reuses connection when available for downstream project lookups, improving reliability and performance of request processing.

### DIFF
--- a/proxy/pkg/proxy/proxy.go
+++ b/proxy/pkg/proxy/proxy.go
@@ -17,6 +17,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/golang-jwt/jwt/v5"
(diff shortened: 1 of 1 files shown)


thibaultmg (approved):
/lgtm

Might be worth commenting why we use `ParseUnverified`.

PR #2455, opened 2026-05-12
[release-2.17] [ACM-30763] Handle storage class changes in MCO CR

coleenquadros:
## Summary
- Stop persisting the default storage class into the MCO CR when not explicitly set by the user. The operator resolves it in-memory for rendering instead.
- Replace global storage-change booleans with instance-level config comparison via `LastStorageConfig`.
- Handle storage class changes by deleting StatefulSets and PVCs so they are re-created with the new class.
- Skip PVC shrink operations in storage size reconciliation.
- Update e2e test to handle missing `storageClass` field in MCO CR using safe type assertions.

## Test plan
- [ ] Unit tests pass (`go test ./operators/multiclusterobservability/...`)
- [ ] E2e storage class and PVC test passes with and without explicit `storageClass` in MCO CR
- [ ] Verify PVCs are recreated when storage class is changed in MCO CR


dbuchanaRH (approved):
/lgtm

thibaultmg (approved):
/lgtm

PR #2463, opened 2026-05-13
fix(proxy): remove vestigial rbac-query-proxy-ingress

coleenquadros:
## Summary

Remove the unused `rbac-query-proxy-ingress` Ingress resource that breaks MCO reconciliation on AWS STS clusters with the AWS Load Balancer Controller (ALBO).

Fixes [ACM-33641](https://redhat.atlassian.net/browse/ACM-33641)

## Problem

On AWS STS clusters with ALBO, the admission webhook `vingress.elbv2.k8s.aws` validates that all referenced IngressClasses exist. The `rbac-query-proxy-ingress` hardcodes `ingressClassName: ingress-open-cluster-management`, but this IngressClass was never registered as a Kubernetes resource. ALBO rejects the Ingress, which halts the entire MCO reconciliation loop — preventing certificate creation, Grafana datasource setup, and leaving Grafana/RBAC Query Proxy pods stuck in ContainerCreating.

## Root Cause

- PR #1312 moved the class from annotation (`kubernetes.io/ingress.class`) to `spec.ingressClassName` to suppress the OCP `IngressWithoutClassName` alert (OCPBUGS-18537)
- Annotations were informal hints not validated by webhooks; `spec.ingressClassName` is a formal API field that ALBO strictly validates
- The `management-ingress` component that served this Ingress is deprecated (last commit July 2023, not in MCH default components)
- The proxy is already exposed via OCP Route (`GenerateProxyRoute()`)

## Fix

Delete the Ingress manifest entirely, following the same pattern used when the Grafana Ingress was removed in commit e7adf1dd (July 2022).

### Files Changed

- **Deleted**: `operators/multiclusterobservability/manifests/base/proxy/ingress.yaml`
- **Modified**: `operators/multiclusterobservability/manifests/base/proxy/kustomization.yaml` (removed `ingress.yaml` entry)

## Test Plan

- [x] `go build ./operators/multiclusterobservability/...` passes
- [x] `go test ./operators/multiclusterobservability/pkg/rendering/...` passes
- [x] `go test ./operators/pkg/deploying/...` passes
- [ ] Verify on an AWS STS cluster with ALBO that MCO reconciliation completes
- [ ] Verify Grafana pods reach Running state
- [ ] Verify proxy is accessible via OCP Route

[ACM-33641]: https://redhat.atlassian.net/browse/ACM-33641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Chores**
  * Removed an ingress resource configuration previously exposing the query proxy.
  * Updated proxy manifest references to remove the now-deleted ingress from the base resources.


operators/multiclusterobservability/manifests/base/proxy/kustomization.yaml
```diff
@@ -3,6 +3,5 @@ resources:
 - cluster-role.yaml
 - cookie-secret.yaml
 - deployment.yaml
-- ingress.yaml
```
subbarao-meduri, line 6: @coleenquadros The PR looks good.  Removing ingress.yaml from kustomization stops new deploys, but existing clusters already running MCO will  keep the old rbac-query-proxy-ingress Ingress forever, no? Should we add something like this in MCO reconciler?

```
  // Clean up vestigial Ingress removed in ACM-33641
  ingress := &networkingv1.Ingress{}
  err := client.Get(ctx, types.NamespacedName{
      Name: "rbac-query-proxy-ingress",
      Namespace: mcoNamespace,
  }, ingress)
  if err == nil {
      _ = client.Delete(ctx, ingress)
  }
```

subbarao-meduri (approved):

PR #2471, opened 2026-05-18
fix(rightsizing): add ADC delegation signal for MCO/MCOA mode disambiguation

rajusem:
## Summary

- Adds `rightSizingDelegated` key to AddOnDeploymentConfig (ADC) so MCOA can distinguish "MCO handles right-sizing" from "feature genuinely disabled in MCOA mode"
- Analytics controller syncs this key alongside existing namespace/virtualization keys
- Renderer seeds the key during initial ADC creation

## Problem

When right-sizing runs in MCO mode and the user enables any MCOA-managed feature (metrics collection, incident detection, etc.), both operators are active on the
cluster. MCOA has no way to distinguish "MCO handles RS" from "feature genuinely disabled in MCOA mode" — both cases result in ADC keys set to "disabled." This causes
MCOA to delete ConfigMaps that MCO created, triggering a create-delete loop. MCOA cannot query the MCO CR directly because it lacks RBAC permissions for
multiclusterobservabilities.

## Solution

Add a `rightSizingDelegated` key to ADC so MCOA can determine which component owns ConfigMap lifecycle. The analytics controller syncs this key alongside existing
namespace/virtualization keys.

- `rightSizingDelegated=true` → MCOA mode: MCOA owns ConfigMap lifecycle, reads namespace/virtualization keys normally
- `rightSizingDelegated=false` → MCO mode: MCO owns ConfigMap lifecycle via Policy, MCOA must skip ConfigMap cleanup

## Summary by CodeRabbit

* **New Features**
  * ADCs now include an explicit right-sizing "delegated" flag so delegation state is always available.
  * Logging now reports delegated, namespace, and virtualization right-sizing targets for clearer operational visibility.

* **Refactor**
  * Right-sizing delegation tracking enhanced to keep ADC customization consistent across scenarios.

* **Tests**
  * Updated tests to validate the new delegated right-sizing flag and existing behaviors.


thibaultmg (approved):
/lgtm

PR #2472, opened 2026-05-18
Refactor e2e hub cluster name resolution to use ManagedCluster API

subbarao-meduri:
Summary:
  - Stop overwriting testOptions.HubCluster.Name globally in initVars() — reverts to only setting default when empty 
  - MCOA test now discovers all clusters directly from ManagedCluster CRs via GetAvailableClustersFromAPI(), independent of options.yaml
  - Consolidated GetAvailableManagedClustersAsClusters and GetAllAvailableClustersAsClusterList into single GetAvailableClustersFromAPI(opt, includeHub) — API-driven discovery with optional connection info enrichment from options.yaml
  - GetHubClusterName now returns error when no hub found instead of silent "local-cluster" fallback

Fixes hub name mismatch in QE environments where options.yaml hub name (infrastructure hostname) differs from ManagedCluster name. Also fixes empty spoke list when options.yaml has no managedClusters entries.

## Summary by CodeRabbit

* **Tests**
  * Improved cluster discovery for end-to-end tests, including clearer separation of hub and managed clusters.
  * Enhanced cluster information handling and error reporting when cluster data cannot be retrieved.
  * Simplified hub cluster setup by defaulting to “local-cluster” when no name is provided.

### DIFF
--- a/tests/pkg/tests/observability-e2e-test_suite_test.go
+++ b/tests/pkg/tests/observability-e2e-test_suite_test.go
@@ -272,12 +272,9 @@ func initVars() {
 	if testOptions.HubCluster.Password != "" {
 		kubeadminCredential = testOptions.HubCluster.Password
 	}
-	hubName, err := utils.GetHubClusterName(testOptions)
-	if err != nil {
-		klog.Warningf("Failed to resolve hub managed cluster name, using default: %v", err)
-		hubName = "local-cluster"
+	if testOptions.HubCluster.Name == "" {
+		testOptions.HubCluster.Name = "local-cluster"
 	}
-	testOptions.HubCluster.Name = hubName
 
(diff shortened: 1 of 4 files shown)


jacobbaungard (changes requested):
This breaks tests when not all managed clusters are added to the resources.yaml. This is almost always the case when executing the tests in the QE env.

There will need to be two lists:
1) A list of managed clusters from resource.yaml needs to be used when we need a kubeclient to login to the managed cluster (it happens at least when calling i.e `CheckDeploymentAvailabilityOnClusters` but maybe other places)
2) A list of managed clusters from the API, when we are not required to do things on the managed cluster side, such as check if metrics from each managed cluster has arrived.

PR #2476, opened 2026-05-26
fix(rightsizing): eliminate ConfigMap predicate reconciliation storm

rajusem:
The analytics controller's ConfigMap predicates returned true for non-matching ConfigMaps and executed side-effects (Policy/Placement creation) inside the informer goroutine, causing many spurious reconciles per operator restart.

Fix 1: Predicate now returns false for non-matching ConfigMaps, adds explicit DeleteFunc/GenericFunc returning false, and skips reconcile when ConfigMap data hasn't changed.

Fix 2: Removed processConfigMap side-effect closure from predicate. Apply logic now runs unconditionally in the reconcile loop via HandleComponentRightSizing, which is safe since all apply functions are idempotent (Get-then-Create/Update).

## Summary by CodeRabbit

* **Refactor**
  * Simplified ConfigMap event predicates to be pure filters (no side-effectful processing) and reduced coupling to controller context.
  * Updated right-sizing logic to always load/apply config when enabled and to perform cleanup only when namespace bindings change.
  * Adjusted analytics reconcile flow to requeue immediately after adding finalizers for faster follow-up processing.

* **Tests**
  * Updated unit assertions to reflect the revised right-sizing behavior.

### DIFF
--- a/operators/multiclusterobservability/controllers/analytics/analytics_controller.go
+++ b/operators/multiclusterobservability/controllers/analytics/analytics_controller.go
@@ -140,7 +140,7 @@ func (r *AnalyticsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			return ctrl.Result{}, fmt.Errorf("rs - failed to add analytics finalizer: %w", err)
 		}
 		reqLogger.Info("rs - Analytics finalizer added to MCO CR")
-		return ctrl.Result{}, nil // watch-triggered reconcile picks up updated finalizers
+		return ctrl.Result{Requeue: true}, nil
 	}
 
 	// Do not reconcile objects if this instance of mch is labeled "paused"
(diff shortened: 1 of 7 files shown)


dvandra (approved):
lgtm

coleenquadros (approved):

subbarao-meduri:
@rajusem should this be cherry picked to release-2.16 too? we are seeing random e2e failures. Could this be related?

```  
Summarizing 1 Failure:
    [FAIL] RHACM4K-58751: Enable and teardown virtualization right-sizing
  recommendation (rightsizing/g1) [It] Should find the PrometheusRule
  'acm-rs-virt-prometheus-rules' in namespace 'openshift-monitoring'
    /go/src/github.com/stolostron/multicluster-observability-operator/tests/pkg/
  tests/observability_right_sizing_test.go:410
```

PR #2493, opened 2026-05-29
fix(endpoint-operator): Integration test missing hubInfoSecret key

guidonguido:
Integration test TestCMOConfigWatching was not being tested because the pipeline is running `go test -tags integration -run=Integration ./operators/...`

## Changes
- Added Integration in the test name
- Added missing `HubInfoSecret` key `hub-cluster-id` which was causing the test to fail
- Added `SkipNameValidation: true` on the controller manager so subsequent tests running on the same go process does not cause a controller names collision error

## Summary by CodeRabbit

* **Tests**
  * Enhanced integration test configuration and setup for observability endpoint metrics, providing improved test reliability and comprehensive validation coverage for endpoint monitoring functionality.


coleenquadros (approved):

thibaultmg:
Merging. The e2e "[It] should configure hub Thanos components with the correct CLI arguments" test is flaky. The observatorium operator fails to update quickly Thanos. It is unrelated to this change.

PR #2509, opened 2026-06-10
fix(tests): POST alerts via obs-api with mTLS in e2e

coleenquadros:
## Summary
- Refactors the route test to POST alerts via observatorium-api with mTLS (hub collector cert, writeOnly RBAC) and GET alerts via the alertmanager route with bearer token (readOnly path)
- Adds `GetObsAPIServerCA`, `GetObsAPIClientCert`, and `GetHubClusterID` test utilities to fetch mTLS certs from `openshift-monitoring` for obs-api authentication

## Summary by CodeRabbit

* **Tests**
  * Updated observability integration tests to validate alert creation and retrieval using the Observatorium alerting API and improved TLS authentication handling.
  * Expanded test utilities to fetch per-cluster monitoring TLS materials and hub cluster identifiers to support mTLS-based test flows.


tests/pkg/tests/observability_alert_test.go
```diff
@@ -535,32 +519,16 @@ var _ = Describe("", func() {
...
+			Scheme: "https",
+			Host:   "observatorium-api-open-cluster-management-observability.apps." + testOptions.HubCluster.BaseDomain,
+			Path:   "/api/alertmanager/v2/default/api/v2/alerts",
 		}
 		q := amURL.Query()
 		q.Set("filter", "alertname=Watchdog")
 		amURL.RawQuery = q.Encode()
 
-		caCrt, err := utils.GetRouterCA(hubClient)
+		caCrt, err := utils.GetObsAPIServerCA(hubClient)
```
dbuchanaRH: Because obsAPI used mTLS, we also need to have the cert here. Will follow up with fix
https://github.com/stolostron/multicluster-observability-operator/blob/4dda4b64988e6594d161b872cf1d6fd1ebe7d1a4/operators/endpointmetrics/controllers/observabilityendpoint/ocp_monitoring_config.go#L277-L301

tests/pkg/tests/observability_alert_test.go
```diff
@@ -381,56 +380,38 @@ var _ = Describe("", func() {
...
-					TLSClientConfig: &tls.Config{RootCAs: pool},
+					Proxy: http.ProxyFromEnvironment,
+					TLSClientConfig: &tls.Config{
+						RootCAs:      pool,
+						Certificates: []tls.Certificate{clientCert},
+					},
 				},
 			}
 
 			alertGetReq, err := http.NewRequest("GET", amURL.String(), nil)
```
dbuchanaRH, line 423: This request should not go through observatorium-api. We would expect alerts to be POSTed from prometheus in managed clusters. Then we should be able to GET the alert from hub alertmanager. So I think the original logic was right here

tests/pkg/tests/observability_route_test.go
```diff
@@ -149,21 +140,24 @@ var _ = Describe("", func() {
...
+					Proxy: http.ProxyFromEnvironment,
+					TLSClientConfig: &tls.Config{
+						RootCAs:      pool,
+						Certificates: []tls.Certificate{clientCert},
+					},
+				},
 			}
+
 			if !alertCreated {
 				resp, err := client.Do(alertPostReq)
```
dbuchanaRH: This test change is valid. Would suggest reverting changes in tests/pkg/tests/observability_alert_test.go but keeping this

tests/pkg/tests/observability_route_test.go
```diff
@@ -189,11 +183,6 @@ var _ = Describe("", func() {
 				return err
 			}
 
-			if os.Getenv("IS_KIND_ENV") != trueStr {
-				BearerToken, err = utils.FetchBearerToken(testOptions)
-				alertGetReq.Header.Set("Authorization", "Bearer "+BearerToken)
-			}
-
 			resp, err := client.Do(alertGetReq)
```
dbuchanaRH: Certs and route should be different on this get request

tests/pkg/tests/observability_route_test.go
```diff
@@ -140,32 +134,41 @@ var _ = Describe("", func() {
...
+			// POST via obs-api with mTLS (hub collector cert — writeOnly RBAC)
+			obsCaCrt, err := utils.GetObsAPIServerCA(hubClient)
+			if err != nil {
+				return fmt.Errorf("GetObsAPIServerCA: %w", err)
 			}
+			obsPool := x509.NewCertPool()
+			obsPool.AppendCertsFromPEM(obsCaCrt)
 
-			client := &http.Client{}
-			if os.Getenv("IS_KIND_ENV") != trueStr {
```
dbuchanaRH, line 162: BearerToken still needs to be used with POST of alert to observatorium-api

dbuchanaRH (approved):
/lgtm

PR #2514, opened 2026-06-17
Test: Ensure MCH pullsecret is created in right ns

jacobbaungard:
Sometimes MCH might be installed in a different namespace (commonly `ocm`) in regression tests. This caused issues creating the pull secret, since the namespace was hardcoded.

This correctly detects which namespace MCH is installed, and uses that one on the secret.

## Bug Fixes
* Corrected pull secret creation to use the MultiClusterHub instance’s own namespace instead of a hardcoded default, improving resource isolation and multi-instance compatibility. The secret name is unchanged for backward compatibility, and failure messages now include the target namespace to simplify troubleshooting.


jacobbaungard:
/cherry-pick release-2.13 release-2.14 release-2.15 release-2.16 release-2.17

thibaultmg (approved):
/lgtm

PR #2528, opened 2026-06-25
ci: query latest-published snapshot tags to avoid race during publish

guidonguido:
The snapshot images take a ~7 minute publish window. `acm-custom-registry` is the last image published, so any tag found there is guaranteed to be available across all stolostron repositories.

In addition, validation webhooks verification are disabled during delete operations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the ACM snapshot lookup to retrieve tags from the correct Quay registry source.
  * Improves reliability of selecting the latest matching ACM snapshot tags to reduce missed or incorrect matches.
  * Fixed the multicluster observability update admission webhook to skip update validation when the object is being deleted.

### DIFF
--- a/scripts/test-utils.sh
+++ b/scripts/test-utils.sh
@@ -21,7 +21,7 @@ get_latest_acm_snapshot() {
   # z version 1-2 digits
   # -SNAPSHOT
   MATCH=$SNAPSHOT_RELEASE".\d{1,2}-SNAPSHOT"
-  LATEST_SNAPSHOT=$(curl -s "https://quay.io/api/v1/repository/stolostron/multicluster-observability-operator/tag/?filter_tag_name=like:$SNAPSHOT_RELEASE&limit=100" | jq --arg MATCH "$MATCH" '.tags[] | select(.name | match($MATCH; "i")  ).name' | sort -r --version-sort | head -n 1)
+  LATEST_SNAPSHOT=$(curl -s "https://quay.io/api/v1/repository/stolostron/acm-custom-registry/tag/?filter_tag_name=like:$SNAPSHOT_RELEASE&limit=100" | jq --arg MATCH "$MATCH" '.tags[] | select(.name | match($MATCH; "i")  ).name' | sort -r --version-sort | head -n 1)
 
   # trim the leading and tailing quotes
   LATEST_SNAPSHOT="${LATEST_SNAPSHOT#\"}"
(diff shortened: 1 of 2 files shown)


scripts/test-utils.sh
```diff
@@ -50,7 +50,7 @@ get_latest_mce_snapshot() {
   # z version 1-2 digits
   # -SNAPSHOT
   MATCH=$SNAPSHOT_RELEASE".\d{1,2}-SNAPSHOT"
-  LATEST_SNAPSHOT=$(curl -s "https://quay.io/api/v1/repository/stolostron/registration/tag/?filter_tag_name=like:$SNAPSHOT_RELEASE&limit=100" | jq --arg MATCH "$MATCH" '.tags[] | select(.name | match($MATCH; "i")  ).name' | sort -r --version-sort | head -n 1)
+  LATEST_SNAPSHOT=$(curl -s "https://quay.io/api/v1/repository/stolostron/acm-custom-registry/tag/?filter_tag_name=like:$SNAPSHOT_RELEASE&limit=100" | jq --arg MATCH "$MATCH" '.tags[] | select(.name | match($MATCH; "i")  ).name' | sort -r --version-sort | head -n 1)
```
jacobbaungard: if we are using the same repo, I think `get_latest_mce_snapshot` is useless?

I was under the impression that the MCE snapshots were different, but seems they are the same unless you look for the BACKPLANE ones like ` 2.8.3-BACKPLANE-2026-06-25-12-21-08 `.

iirc there was some reason we had different ones. @thibaultmg do you remember?

I think we either should:
- in the `mce_snapshot` function use `BACKPLANE` instead of `SNAPSHOT` to filter the images (and keep the regiration operator or similar backplane image)
- Or we get rid of this function all together.

guidonguido: yes, looking at the registry, it looks like the `registration` SNAPSHOT image tags are created together with other components.

We either use BACKPLANE for MCE or remove the function. If we are sure SNAPSHOTs are always included for the MCE components, and generated when needed, let's go for the second option.

thibaultmg: Looking at my PR description at the time: https://github.com/stolostron/multicluster-observability-operator/pull/1779

The snapshots differ from time to time.
Maybe it's not the case anymore 🤷

guidonguido: Thank you @thibaultmg, let's just revert the `get_latest_mce_snapshot()` to fetch the tag from `registration` then

jacobbaungard (approved):
/lgtm

PR #2541, opened 2026-07-06
[release-2.14] Prevent downgrade of thanos receive PVC to 1Gi in e2e

dbuchanaRH:
In jenkins QE regression tests, RHACM4K-43019 will apply MCO CR that lowers thanos receive PVC from 10Gi to 1Gi. The observability-thanos-receive pods will then fail to start as no storage is left on the device. No observability-thanos-receive pods will start causing a cascading failure in CI


jacobbaungard (approved):
/lgtm

PR #2549, opened 2026-07-07
[release-2.14] [ACM-30479] <MCO>: <Fix Issue Where ViolatedPolicyGuidelines rule would break when additional labels are added to managed cluster that is violating policy guidelines>

GeorgeFrankAeillo:
This is a cherry-pick of https://github.com/stolostron/multicluster-observability-operator/pull/2486 to release-2.14. This fixes the issue where ViolatedPolicyGuidelines rule would break when additional labels are added to managed cluster that is violating policy guidelines.


dbuchanaRH (approved):
/lgtm

PR #2558, opened 2026-07-08
Prevent downgrade of thanos receive PVC to 1Gi in jenkins e2e

dbuchanaRH:
e2e tests are deployed with thanos receive storage size at 10Gi. In jenkins QE regressions the test [RHACM4K-43019](https://github.com/stolostron/multicluster-observability-operator/blob/7f469b761faeb6cceeb188427c11a9490c1fef31/tests/pkg/tests/observability_config_test.go#L371) can run which downgrades the PVC size for thanos receive to 1Gi. If the test suite was running for an extended period of time this will push the thanos receive pods into an error state.

As mentioned in https://github.com/stolostron/multicluster-observability-operator/pull/2257 increasing the PV storage will delay the problem for now. There may come a need for another solution.

After these changes thanos receive storage size should stay at 10Gi through the duration of the e2e suite. receiveStorageSize is not changed for unit tests or for kind e2e tests i.e.
```
/examples/mco/e2e/v1beta2/custom-certs-kind/observability.yaml
```

## Summary by CodeRabbit

* **Documentation / Examples**
  * Updated observability example configurations to increase receive storage capacity to `10Gi` (from `1Gi`).
  * Applied the update consistently across initial and advanced setup examples, including custom-certificate variants.

### DIFF
--- a/examples/updatemcocr/advancedmcoconfig/custom-certs/v1beta2-observability-custom-mco.yaml
+++ b/examples/updatemcocr/advancedmcoconfig/custom-certs/v1beta2-observability-custom-mco.yaml
@@ -155,6 +155,6 @@ spec:
       name: thanos-object-storage
       tlsSecretMountPath: /etc/minio/certs
       tlsSecretName: minio-tls-secret
-    receiveStorageSize: 1Gi
+    receiveStorageSize: 10Gi
     ruleStorageSize: 1Gi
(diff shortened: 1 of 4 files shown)


thibaultmg (approved):
/lgtm

coleenquadros (approved):

coleenquadros (approved):

PR #2559, opened 2026-07-09
[release-2.17] e2e: use latest konflux build of MCOA/obo

jacobbaungard:
For the relevant release. This ensures we are getting the right image for the specific stream instead of hardcoding the latest version.


thibaultmg (approved):
/lgtm

PR #2567, opened 2026-07-14
[ACM-36434] feat(analytics): migrate right-sizing from Policy-based (MCO) to always-MCOA for ACM 5.0

rajusem:
## Summary

- Migrate right-sizing recommendation from MCO Policy-based delivery to always-MCOA 
  (ManifestWork-based) for ACM 5.0 GA
- Remove the entire Policy-based right-sizing subsystem (rs-namespace, rs-virtualization, 
  rs-utility policy/placement/placementbinding code)
- Simplify analytics controller — remove delegation detection, cleanup stabilization, 
  and dual-mode switching logic
- Remove dead RightSizingDelegated field and strengthen cleanup tests
- Add error-discard comments at GetComponentConfig call sites

## Changes

- Deleted: Policy, Placement, PlacementBinding, PrometheusRule builders for both 
  namespace and virtualization right-sizing
- Deleted: ConfigMap-based right-sizing resource management (rs-utility/configmap.go)
- Simplified: analytics_controller.go, rightsizing_controller.go, component.go
- Removed: IsRightSizingDelegated utility and tests
- Updated: RBAC (mco_role.yaml), CSV, predicate functions

## Summary

- **New Features**
  - Right-sizing now consistently drives AddOnDeploymentConfig (ADC) customized variables from the CR’s state, including keeping delegation enabled.
  - Deletion now uses a safer two-phase flow with a stabilization window.
- **Bug Fixes**
  - Improved error handling when retrieving the MCO instance (non-“not found” failures now surface).
  - Simplified right-sizing reconciliation to better align with MCOA/CR spec behavior.
- **Security**
  - Tightened least-privilege RBAC permissions for policy and placement resources.
- **Tests**
  - Updated and expanded analytics and right-sizing test coverage, including ADC validation.


dvandra (approved):
LGTM

operators/multiclusterobservability/main.go
```diff
@@ -327,8 +327,6 @@ func main() {
 
 	if err = (&analytics.AnalyticsReconciler{
 		Client: mgr.GetClient(),
```
thibaultmg, line 329: Why do you remove the log prefix?

rajusem: Added back the `Log` field to maintain consistency with other controllers in `main.go` (controllers.Analytics prefix). 

The original removal was because after removing the Policy-based RS code, the reconciler was simplified — Scheme was no longer needed (no typed resource creation, only client operations and unstructured patches), and `Log` was replaced with a package-level logger. Restored `Log` to follow the `main.go` convention. `Scheme` remains removed as it's genuinely unused.

thibaultmg (commented):
Looks good. 

We'll have to test the interactions with MCOA for metrics collection. Like today, we deploy a endpoint controller that modifies the CMO configmaps in the manage clusters. When the addon is uninstalled, we deploy a cleanup job to reset them. Here, the addon will remain when mcoa for metrics is disabled, and the resources won't be cleaned. To fix this, we may have to permanently deploy this endpoint controller while today, it is only deployed when:

`{{- if or .Values.platformEnabled .Values.userWorkloadsEnabled }}`

Also, when MCOA for metrics is disabled, we wait that all manifest works are deleted so that the cleanup job can be triggered by the addon manager before deleting it. But now this won't happen. And the legacy collector will not start. We'll have to fix that too. I don't know how.

I'll let the rest of the team see how they want to manage that. Like wether we merge and fix after.

rajusem:
@jacobbaungard @thibaultmg 
**Issue 1 (legacy collector blocked)**: When MCOA metrics are disabled but RS stays enabled, hasMCOAManifestWorks in the placement controller can't distinguish RS ManifestWorks from metrics ManifestWorks (since they carry the same addon label). This causes the gate to permanently block the legacy collector.

**Issue 2 (CMO cleanup)**: The pre-delete cleanup job only triggers on CMA deletion. Since RS keeps the CMA alive, the endpoint controller's CMO config modifications persist on the spokes even after disabling metrics.

For Issue 1, I have a proposed fix — replace hasMCOAManifestWorks with a content-aware hasMCOAMetricsManifestWorks that checks the ManifestWork spec for the PrometheusAgent kind (the definitive metrics marker: always present when MCOA  metrics are active, never produced by RS). This preserves the transition safety guard while allowing the legacy addon to deploy when the ManifestWorks only contain RS content. No additional API calls — same List call as before, only adds in-memory partial unmarshal of the manifest kind field.
Gist with the proposed change: https://gist.github.com/rajusem/35dcf756d145d3fa980b6d5a818497b1

For Issue 2, I explored two approaches — endpoint controller self-cleanup on state transition vs. always deploying the endpoint controller. The self-cleanup approach has a timing issue (ManifestWork applies deployment removal and state update atomically, so the controller can't observe the transition before being killed). The "always deploy" approach works without any Go code changes — the existing reconcile logic already handles empty AlertmanagerEndpoint by  reverting CMO config automatically.
Gist with both options and analysis: https://gist.github.com/rajusem/3f9baed1e0060cfd2446ebeb48e0713a

I can include the fix for Issue 1 in this PR or raise it as a separate follow-up — let me know what you prefer!

> Looks good.
> 
> We'll have to test the interactions with MCOA for metrics collection. Like today, we deploy a endpoint controller that modifies the CMO configmaps in the manage clusters. When the addon is uninstalled, we deploy a cleanup job to reset them. Here, the addon will remain when mcoa for metrics is disabled, and the resources won't be cleaned. To fix this, we may have to permanently deploy this endpoint controller while today, it is only deployed when:
> 
> `{{- if or .Values.platformEnabled .Values.userWorkloadsEnabled }}`
> 
> Also, when MCOA for metrics is disabled, we wait that all manifest works are deleted so that the cleanup job can be triggered by the addon manager before deleting it. But now this won't happen. And the legacy collector will not start. We'll have to fix that too. I don't know how.
> 
> I'll let the rest of the team see how they want to manage that. Like wether we merge and fix after.

jacobbaungard:
@rajusem if you have the fix ready, I purpose we add it here. For issue 2 we'd need to ensure it's fixed at least before the bug cutoff - could I get you to create a jira for it?

rajusem:
@jacobbaungard Fixed issue1 with https://github.com/stolostron/multicluster-observability-operator/pull/2567/commits/61b968cb3265e65d12ae361085ef85a0fb2af1de with current PR and created [ACM-38875](https://redhat.atlassian.net/browse/ACM-38875) for issue2.   

[ACM-38875]: https://redhat.atlassian.net/browse/ACM-38875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jacobbaungard (approved):
/lgtm

PR #2604, opened 2026-07-29
[release-2.16] fix(rightsizing): eliminate ConfigMap predicate reconciliation storm

rajusem:
## Summary

Manual cherry-pick of #2476 (fc99ea8b) to release-2.16. Automated cherry-pick was not possible due to merge conflicts (4 files conflicted, all resolved manually).

- **Predicate is now a pure filter** — removed `processConfigMap` side-effect closure that ran Policy/Placement/PlacementBinding creation inside the informer goroutine
- **Apply runs unconditionally in reconcile loop** — `HandleComponentRightSizing` now always reads the ConfigMap and calls `ApplyChangesFunc`, not just on namespace binding change
- **Eliminates reconciliation storm** — predicate returns `false` for non-matching ConfigMaps (was `true`, causing ~80 spurious reconciles per restart)
- **Explicit requeue after finalizer** — `Requeue: true` instead of relying on watch-triggered reconcile

## Root Cause

The virt right-sizing e2e test (`RHACM4K-58751`) fails intermittently because:
1. `HandleComponentRightSizing` only called `ApplyChangesFunc` when `namespaceBindingUpdated == true`
2. Initial Policy/Placement creation relied on a predicate side-effect (race condition)
3. On operator restart or timing mismatch, the predicate didn't fire → PrometheusRule never deployed

## Conflict Resolution

4 files had merge conflicts during cherry-pick, all trivial:
- `analytics_controller.go` — predicate function signature `(ctx, c)` → `()`
- `component.go` — moved apply block out of `if namespaceBindingUpdated`
- `configmap.go` — removed `processConfigMap` closure
- `component_test.go` — inserted new test functions

Additionally, `mockGetDefaultConfigFunc` in tests was updated to return valid YAML since `ApplyChangesFunc` now runs unconditionally (previously unreachable code path).

### DIFF
--- a/operators/multiclusterobservability/controllers/analytics/analytics_controller.go
+++ b/operators/multiclusterobservability/controllers/analytics/analytics_controller.go
@@ -87,7 +87,7 @@ func (r *AnalyticsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			return ctrl.Result{}, fmt.Errorf("rs - failed to add analytics finalizer: %w", err)
 		}
 		reqLogger.Info("rs - Analytics finalizer added to MCO CR")
-		return ctrl.Result{}, nil // watch-triggered reconcile picks up updated finalizers
+		return ctrl.Result{Requeue: true}, nil
 	}
 
 	// Do not reconcile objects if this instance of mch is labeled "paused"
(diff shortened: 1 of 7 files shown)


jacobbaungard (approved):
/lgtm

PR #2678, opened 2026-08-20
fix(analytics): use APIReader for legacy policy and placement cleanup

thibaultmg:
Partially addresses https://redhat.atlassian.net/browse/ACM-38875. A follow-up PR will address remaining watchers if applicable.

### Problem
When `AnalyticsReconciler` performs legacy migration cleanup (`CleanupLegacyPolicyResources`) or deletion cleanup (`CleanupRightSizingResources`), it invokes `.List(...)` on `Policy`, `PlacementBinding`, and `Placement` resources using `r.Client`.

Because `r.Client` is a cached `controller-runtime` client, calling `.List(...)` dynamically starts background Informers that issue cluster-scoped `WATCH` requests to the Kubernetes API server. However, `mco_role.yaml` intentionally grants only `list` and `delete` permissions (no `watch`) for these legacy governance resources, causing the API server to reject the requests with `403 Forbidden` (`User ... cannot watch resource "policies" ... at the cluster scope`).

### Fix
- Inject `APIReader` (`mgr.GetAPIReader()`) into `AnalyticsReconciler`.
- Pass `apiReader client.Reader` to `CleanupRSResourcesByLabel` and `CleanupLegacyPolicyResourcesByLabel` to perform direct, uncached `List` queries without starting Informer watches.
- Added unit test coverage in `analytics_controller_test.go` and `component_test.go`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup during analytics migration by ensuring resources are reliably discovered and removed.
  * Enhanced cleanup of legacy policies and right-sizing resources, including labeled resources.

* **Tests**
  * Added coverage verifying successful migration completion and removal of legacy policies.

### DIFF
--- a/operators/multiclusterobservability/controllers/analytics/analytics_controller.go
+++ b/operators/multiclusterobservability/controllers/analytics/analytics_controller.go
@@ -41,6 +41,7 @@ const analyticsStabilizationWindow = 10 * time.Second
 // that would not be consistent across replicas.
 type AnalyticsReconciler struct {
 	Client        client.Client
+	APIReader     client.Reader
 	Log           logr.Logger
 	migrationDone bool
 	cleanupAt     time.Time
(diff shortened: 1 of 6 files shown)


thibaultmg:
Closing as it fixes the same bug as https://github.com/stolostron/multicluster-observability-operator/pull/2674
The approach is different, but adding watchers on the Policy, Placement and PlacementBindings shouldn't have a noticeable impact on memory as they usually have a limited number of instances.

PR #2690, opened 2026-08-26
fix(grafana): maintain time range when navigating between VM right-si…

dvandra:
fix(grafana): preserve time range across VM right-sizing dashboard navigation

Enable keepTime and includeVars on the "Back to Main Dashboard" links in the virtualization overestimation and underestimation dashboards so users don't lose their selected time range and variable context when navigating between dashboards.

## Summary by CodeRabbit

- **Bug Fixes**
  - Grafana dashboard links now preserve selected variable values and the current time range when navigating between related dashboards.
  - Updated both right-sizing virtualization overestimation and underestimation dashboards for consistent navigation behavior.


thibaultmg (approved):
/lgtm

PR #2701, opened 2026-08-31
fix(mcoa): synchronize addon teardown and harden e2e install/config tests

thibaultmg:
## Description

Fixes MCOA teardown synchronization and hardens E2E test suites against timeouts and flakes:

* **MCOA Teardown Synchronization:** Waits for all spoke-side `ManifestWork`s and `ManagedClusterAddOn`s across available clusters before undeploying the MCOA manager (`HasRemainingMCOAResources`), preventing orphaned spoke resources and stuck finalizers.
* **Install Timeout & Crash-Loop Fast-Fail:** Extends MCO/OBA readiness timeouts to 30 minutes for slow environments (e.g. IBM Power) while adding a `stopIfCrashLooping` guard (>5 container restarts) to abort immediately via Gomega `StopTrying` on real regressions.
* **MCO CR Conflict Retries:** Wraps MCO CR updates in config tests and test helpers in `retry.RetryOnConflict` to resolve optimistic locking conflicts from concurrent reconciliations.
* **Capability Teardown Hardening:** Explicitly sets all MCOA capabilities to `false` on teardown to prevent the analytics controller from re-enabling them, and synchronizes CMA/MCA deletion before removing CRDs.
* **Log Noise Reduction:** Trims `spec.workload` from ManifestWorks in debug log printers.

### DIFF
--- a/operators/multiclusterobservability/pkg/util/mcoa_util.go
+++ b/operators/multiclusterobservability/pkg/util/mcoa_util.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"slices"
 
 	mcov1beta2 "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/api/v1beta2"
@@ -60,13 +61,7 @@ func HasMCOAManifestWorks(ctx context.Context, c client.Client) ([]string, error
 		return nil, fmt.Errorf("failed to list ManagedClusters: %w", err)
 	}
 
-	ignoredNamespaces := make(map[string]struct{})
-	for _, mc := range clusterList.Items {
-		isAvailable := meta.IsStatusConditionTrue(mc.Status.Conditions, clusterv1.ManagedClusterConditionAvailable)
-		if !isAvailable {
-			ignoredNamespaces[mc.Name] = struct{}{}
-		}
-	}
+	ignoredNamespaces := getUnavailableClusterNamespaces(clusterList)
 
 	workList := &workv1.ManifestWorkList{}
 	opts := []client.ListOption{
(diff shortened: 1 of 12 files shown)


operators/multiclusterobservability/pkg/util/mcoa_util.go
```diff
@@ -116,3 +108,78 @@ func containsPrometheusAgent(work workv1.ManifestWork) bool {
...
+
+// HasRemainingMCOAResources checks for ANY remaining MCOA ManifestWorks or ManagedClusterAddOns
+// on available ManagedClusters.
+//
+// Unlike HasMCOAManifestWorks (which only checks for PrometheusAgent to coordinate legacy
+// metrics collector transitions), this function checks for ANY remaining MCOA ManifestWorks
+// or ManagedClusterAddOns. It is used during MCOA teardown to ensure the MCOA manager
+// deployment is not undeployed prematurely while spokes are still cleaning up resources.
+//
+// Resources on unavailable ManagedClusters are ignored to prevent disconnected spokes
```
jacobbaungardrh, line 133: What happens if disconnected spokes connect again and they have MCOA resources left?

thibaultmg: It should be cleaned automatically when it comes back online. The corresponding manifeswork will have a deletion timestamp and a cleanup finalizer. Once the managed cluster reconnects, the klusterlet will delete the resources listed in the manifestwork. 
However it will not get the endpoint-operator cleanup job for the cmo configmaps. 
It's a tradeoff...

jacobbaungardrh: ack ok. I guess that is acceptable.

jacobbaungardrh (approved):
/lgtm

PR #2709, opened 2026-09-02
First pass of adding Loki Operator Installation before MCOA Pod is deployed

GeorgeFrankAeillo:
This is the first pass of adding support for Loki Operator Installation in the MCO. This occurs if logging is enabled and the MCOA will get deployed but the hub, currently does not have the LokiStack CRD. It checks this, and if not, it installs the Loki Operator resources into the cluster, which install the LokiStack CRD.

### DIFF
--- a/operators/multiclusterobservability/pkg/dependencies/loki_operator.go
+++ b/operators/multiclusterobservability/pkg/dependencies/loki_operator.go
@@ -55,7 +55,11 @@ func BuildLokiOperatorResources() []client.Object {
 	sub.SetName(mcoconfig.LokiOperatorPackageName)
 	sub.SetNamespace(mcoconfig.LokiOperatorNamespace)
 	sub.Object["spec"] = map[string]any{
-		"channel":             mcoconfig.LokiOperatorChannel,
+		// channel is deliberately omitted: it's an optional field on the Subscription CRD, and
+		// leaving it unset tells OLM to track the package's current default channel. Loki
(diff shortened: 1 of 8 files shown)


coleenquadros:
COO is also installed via OLM Subscription. We can follow installation similar to how we do it here? 
https://github.com/stolostron/multicluster-observability-addon/pull/592/changes#diff-3b6fb46e93cc67867945cbcae358cd82ba05cdafe012c514c78d42de986cc101

cc @saswatamcode

GeorgeFrankAeillo:
@coleenquadros I originally had it in the MCOA, but since MCOA watches LokiStack resources, it is needed before the MCOA is started so that LokiStack resources can be created

operators/multiclusterobservability/pkg/dependencies/loki_operator.go
```diff
@@ -0,0 +1,82 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// BuildLokiOperatorResources returns the Namespace, OperatorGroup and Subscription objects
+// required to install Loki Operator via OLM on the hub cluster.
+//
+// OperatorGroup/Subscription are built as unstructured objects (rather than importing
+// github.com/operator-framework/api) to avoid adding a new dependency to this operator for
+// three object kinds.
+func BuildLokiOperatorResources() []client.Object {
```
saswatamcode, line 29: So this generally looks good, only thing I'm wondering is if we should use the controller that @coleenquadros mentioned here directly in MCOA

saswatamcode (approved):

PR #2711, opened 2026-09-04
[ACM-43197] Split right-sizing Grafana profile into CPU profile and memory profile

dvandra:
- Split the single `$profile` Grafana variable into `$cpu_profile` and `$memory_profile` so CPU and memory percentile recommendations can be selected independently on the namespace and virtualization right-sizing dashboards.
- Add e2e coverage that those dashboards expose the new variables and no longer reference `$profile`.


jacobbaungard (approved):
/lgtm

jacobbaungard (approved):
/lgtm

PR #2712, opened 2026-09-04
First pass of adding Cert Manager Operator installation before MCOA Pod is deployed

GeorgeFrankAeillo:
This follows the same pattern as https://github.com/stolostron/multicluster-observability-operator/pull/2709. This is the first pass of adding support for Cert Manager Installation in the MCO. This occurs if logging is enabled and the MCOA will get deployed but the hub, currently does not have the Certificates CRD. It checks this, and if not, it installs the Cert Manager Operator resources into the cluster, which install the CertManager CRD. It also includes adding RBAC permissions to the MCOA Pod for certificates.


dbuchanaRH:
Tested on a cluster, installs crds
cert_mr_mcoa_logging

dbuchanaRH (approved):
/lgtm

PR #2713, opened 2026-09-04
First pass of adding Cert Manager Operator installation before MCOA Pod is deployed

GeorgeFrankAeillo:
This follows the same pattern as https://github.com/stolostron/multicluster-observability-operator/pull/2709. This is the first pass of adding support for Cert Manager Installation in the MCO. This occurs if logging is enabled and the MCOA will get deployed but the hub, currently does not have the Certificates CRD. It checks this, and if not, it installs the Cert Manager Operator resources into the cluster, which install the CertManager CRD. It also includes adding RBAC permissions to the MCOA Pod for certificates.

## Summary by CodeRabbit

- **New Features**
  - Platform log collection now automatically verifies and installs the required Loki and cert-manager operators when enabled.
  - The system waits for required operator resources to become available before continuing configuration.

- **Bug Fixes**
  - Added the permissions needed to manage and monitor Loki and cert-manager resources, improving setup reliability.

- **Tests**
  - Added coverage for operator installation, resource creation, repeatable reconciliation, and error handling.

### DIFF
--- a/operators/multiclusterobservability/pkg/dependencies/cert_manager_operator.go
+++ b/operators/multiclusterobservability/pkg/dependencies/cert_manager_operator.go
@@ -1,0 +1,81 @@
+// Copyright (c) Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+// Licensed under the Apache License 2.0
+
+package dependencies
+
+import (
+	"context"
+	"fmt"
+
+	mcoconfig "github.com/stolostron/multicluster-observability-operator/operators/multiclusterobservability/pkg/config"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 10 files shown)


operators/multiclusterobservability/pkg/dependencies/cert_manager_operator.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	return []client.Object{ns, og, sub}
+}
+
+// EnsureCertManagerOperatorInstalled creates the cert-manager Operator's Namespace,
+// OperatorGroup and Subscription if they don't already exist. It never updates an existing
+// object, so it won't fight a manual install or an in-progress OLM upgrade.
+func EnsureCertManagerOperatorInstalled(ctx context.Context, c client.Client) error {
+	for _, obj := range BuildCertManagerOperatorResources() {
```
coleenquadros, line 75: How is the cleanup handled when capabilities are removed?

GeorgeFrankAeillo: I don't think we're covering that right now as part of the managed_logstore POC work, I just made the MR into main to get the prow build kicked off. Do you think we should uninstall the operator?

GeorgeFrankAeillo:
Closing due to MR only being created off of main for build purposes

PR #2727, opened 2026-09-09
[ACM-27838] fix misleading namespace log

dbuchanaRH:
**Problem:**
The logs of the multicluster-observability-operator would mention the wrong cluster namespace
**Fix:**
Change from using request.Namespace to using namespace


GeorgeFrankAeillo (approved):
/lgtm

saswatamcode (approved):

PR #2728, opened 2026-09-09
First pass of adding RootClusterIssuer,RootIssuer,RootCertificate

GeorgeFrankAeillo:
In this MR, I add support for installing a self-signed Issuer, root CA Certificate, and ClusterIssuer when logging is enabled and the certificate CRD is installed. These resources are currently needed by the managed_logstore branch in the MCOA for its self-signed certificates that get distributed to the managed/MCOA created Cluster Log Forwarder. I tested this locally and verified that the resources were created. I also updated the mco_role.yaml file to add these RBAC permissions and regenerated the bundle. I verified that this works end to end on a hub cluster.

` oc get clusterIssuer -A
NAME               READY   AGE
mcoa-root-issuer   True    47s
oc get Issuer -A
NAMESPACE      NAME           READY   AGE
cert-manager   mcoa-root-ca   True    53s
oc get Certificate -A
NAMESPACE      NAME             READY   SECRET           AGE
cert-manager   mcoa-root-cert   True    mcoa-root-cert   57s`

### DIFF
--- a/operators/multiclusterobservability/bundle/manifests/multicluster-observability-operator.clusterserviceversion.yaml
+++ b/operators/multiclusterobservability/bundle/manifests/multicluster-observability-operator.clusterserviceversion.yaml
@@ -49,7 +49,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2026-09-02T13:10:40Z"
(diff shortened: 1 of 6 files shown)


dbuchanaRH (approved):
/lgtm

PR #2730, opened 2026-09-10
fix: report COO status via ClusterClaims

coleenquadros:
## Summary

The endpoint-monitoring-operator now writes COO install status as a single ClusterClaim `coo.observability.open-cluster-management.io` with values: `not-installed`, `mcoa`, `external`. ClusterClaims are automatically synced to ManagedCluster.Status on the hub by the OCM registration agent, so MCOA can read them directly without ManifestWork feedback.

Handles multiple concurrent COO Subscriptions by treating any non-MCOA subscription as external. Skips COO detection on non-OCP clusters where OLM is unavailable.

**Depends on:** stolostron/multicluster-observability-addon#654 (hub reads the ClusterClaim + grants RBAC)

## Changes
- Added `WriteCOOStatus()` method on `MCOAAgentReconciler` — creates/updates a single ClusterClaim
- Added OLM Subscription watch to trigger COO status updates on install/uninstall
- Added `OLMAvailable` field set at instantiation, gating COO detection on non-OCP clusters
- Used `controllerutil.CreateOrUpdate` for ClusterClaim management
- Registered `clusterv1alpha1` scheme for ClusterClaim types
- Added unit tests for `WriteCOOStatus` and `getCOOSubscriptionStatus`
- Added ClusterClaim and Subscription CRD stubs for envtest

## Test plan
- [ ] Deploy both images (addon + endpoint operator)
- [ ] Verify single ClusterClaim created on spoke after endpoint operator starts
- [ ] Install COO manually → claim value `external`
- [ ] Install COO via MCOA → claim value `mcoa`
- [ ] Uninstall COO → claim value `not-installed`
- [ ] Non-OCP spoke → no ClusterClaim created, no crash

### DIFF
--- a/operators/endpointmetrics/controllers/mcoa/coo_status.go
+++ b/operators/endpointmetrics/controllers/mcoa/coo_status.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -33,24 +32,29 @@ const (
 // WriteCOOStatus checks whether a COO Subscription exists on the spoke and
 // writes the result as ClusterClaims. The OCM registration agent automatically
 // syncs these to ManagedCluster.Status.ClusterClaims on the hub.
-func WriteCOOStatus(ctx context.Context, c client.Client, log logr.Logger) error {
-	installed, managedBy, err := getCOOSubscriptionStatus(ctx, c)
+// On non-OCP clusters (OLMAvailable=false), this is a no-op.
+func (r *MCOAAgentReconciler) WriteCOOStatus(ctx context.Context) error {
+	if !r.OLMAvailable {
+		return nil
+	}
+
+	installed, managedBy, err := getCOOSubscriptionStatus(ctx, r.Client)
 	if err != nil {
 		return fmt.Errorf("failed to check COO subscription: %w", err)
 	}
 
-	if err := ensureClusterClaim(ctx, c, CooInstalledClaimName, installed); err != nil {
+	if err := ensureClusterClaim(ctx, r.Client, CooInstalledClaimName, installed); err != nil {
 		return fmt.Errorf("failed to ensure ClusterClaim %s: %w", CooInstalledClaimName, err)
 	}
 
 	if managedBy == "" {
 		managedBy = "none"
 	}
-	if err := ensureClusterClaim(ctx, c, CooManagedByClaimName, managedBy); err != nil {
+	if err := ensureClusterClaim(ctx, r.Client, CooManagedByClaimName, managedBy); err != nil {
 		return fmt.Errorf("failed to ensure ClusterClaim %s: %w", CooManagedByClaimName, err)
 	}
 
-	log.V(1).Info("COO status ClusterClaims updated", "installed", installed, "managedBy", managedBy)
+	r.Log.V(1).Info("COO status ClusterClaims updated", "installed", installed, "managedBy", managedBy)
 	return nil
 }
 
--- a/operators/endpointmetrics/controllers/mcoa/coo_status.go
+++ b/operators/endpointmetrics/controllers/mcoa/coo_status.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	clusterv1alpha1 "open-cluster-management.io/api/cluster/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 const (
@@ -54,23 +55,14 @@ func WriteCOOStatus(ctx context.Context, c client.Client, log logr.Logger) error
 }
 
 func ensureClusterClaim(ctx context.Context, c client.Client, name, value string) error {
-	claim := &clusterv1alpha1.ClusterClaim{}
-	err := c.Get(ctx, client.ObjectKey{Name: name}, claim)
-	if errors.IsNotFound(err) {
-		claim = &clusterv1alpha1.ClusterClaim{
-			ObjectMeta: metav1.ObjectMeta{Name: name},
-			Spec:       clusterv1alpha1.ClusterClaimSpec{Value: value},
-		}
-		return c.Create(ctx, claim)
+	claim := &clusterv1alpha1.ClusterClaim{
+		ObjectMeta: metav1.ObjectMeta{Name: name},
 	}
-	if err != nil {
-		return err
-	}
-	if claim.Spec.Value != value {
+	_, err := controllerutil.CreateOrUpdate(ctx, c, claim, func() error {
 		claim.Spec.Value = value
-		return c.Update(ctx, claim)
-	}
-	return nil
+		return nil
+	})
+	return err
 }
 
 func getCOOSubscriptionStatus(ctx context.Context, c client.Client) (installed string, managedBy string, err error) {
--- a/operators/endpointmetrics/controllers/mcoa/coo_status.go
+++ b/operators/endpointmetrics/controllers/mcoa/coo_status.go
@@ -22,54 +22,44 @@ const (
 	cooSubscriptionName = "cluster-observability-operator"
 	mcoaReleaseLabel    = "release"
 	mcoaReleaseName     = "multicluster-observability-addon"
-	managedByMCOA       = "mcoa"
-	managedByExternal   = "external"
 
-	CooInstalledClaimName = "coo-installed.observability.open-cluster-management.io"
-	CooManagedByClaimName = "coo-managed-by.observability.open-cluster-management.io"
+	CooStatusClaimName = "coo.observability.open-cluster-management.io"
+
+	CooStatusNotInstalled = "not-installed"
+	CooStatusMCOA         = "mcoa"
+	CooStatusExternal     = "external"
 )
 
 // WriteCOOStatus checks whether a COO Subscription exists on the spoke and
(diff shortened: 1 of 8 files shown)


operators/endpointmetrics/controllers/mcoa/coo_status.go
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) Red Hat, Inc.
```
thibaultmg, line 1: Could we add unit tests covering these functions?

operators/endpointmetrics/controllers/mcoa/mcoa_agent_controller.go
```diff
@@ -183,3 +218,25 @@ func (r *MCOAAgentReconciler) mapComponentLabelToRequests(
...
+	obj := &unstructured.Unstructured{}
+	obj.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "operators.coreos.com",
+		Version: "v1alpha1",
+		Kind:    "Subscription",
+	})
+	return obj
+}
+
+func isCOOSubscriptionObject(obj client.Object) bool {
```
thibaultmg, line 222: Could we add unit tests for this function?

operators/endpointmetrics/controllers/mcoa/mcoa_agent_controller.go
```diff
@@ -40,6 +42,7 @@ type MCOAAgentReconciler struct {
 	AccessorSecret                string
 	EnablePlatformAlertForwarding bool
 	EnableUWLAlertForwarding      bool
+	olmAvailable                  bool
```
thibaultmg: Why is this field private? We can't test this easily. It's better to set it externally at instantiation time. We avoid an untestable internal state.

operators/endpointmetrics/controllers/mcoa/mcoa_agent_controller.go
```diff
@@ -144,8 +153,34 @@ func (r *MCOAAgentReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				}}}
+			}),
+			ctrlbuilder.WithPredicates(predicate.Funcs{
+				CreateFunc: func(e event.CreateEvent) bool {
+					return isCOOSubscriptionObject(e.Object)
+				},
+				DeleteFunc: func(e event.DeleteEvent) bool {
+					return isCOOSubscriptionObject(e.Object)
+				},
+				UpdateFunc:  func(_ event.UpdateEvent) bool { return false },
```
thibaultmg, line 165: Isn't a bit risky to ignore all update events? Like what happens if the subscription ownership is changed. For example, mcoa first deployed coo and then some automation from the customer decides to override it and own it. We no longer manage COO. But we fail to detect it. Am I missing something?
We could react on labels update and filter the rest out.

operators/endpointmetrics/controllers/mcoa/mcoa_agent_controller.go
```diff
@@ -78,44 +81,50 @@ func NewMCOAAgentReconciler(
 func (r *MCOAAgentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	r.Log.V(1).Info("Reconciling MCOA Agent", "name", req.Name, "namespace", req.Namespace)
 
+	var cooStatusErr error
+	if r.olmAvailable {
```
thibaultmg: Why don't we move this at the end of the function where we return an error if cooStatusErr is not nil?
Or even better, we add a switch case to handle the WriteCOOStatus, otherwise, whenever the subscription is updated, it will log the default case "Ignoring event for unmanaged resource", which is confusing.

operators/endpointmetrics/controllers/mcoa/coo_status.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	if err := ensureClusterClaim(ctx, c, CooManagedByClaimName, managedBy); err != nil {
+		return fmt.Errorf("failed to ensure ClusterClaim %s: %w", CooManagedByClaimName, err)
+	}
+
+	log.V(1).Info("COO status ClusterClaims updated", "installed", installed, "managedBy", managedBy)
+	return nil
+}
+
+func ensureClusterClaim(ctx context.Context, c client.Client, name, value string) error {
```
thibaultmg: It would be cleaner and simpler to use controllerutil.CreateOrUpdate in this function.

operators/endpointmetrics/controllers/mcoa/coo_status.go
```diff
@@ -0,0 +1,104 @@
...
+	})
+
+	if err := c.List(ctx, subList); err != nil {
+		if errors.IsNotFound(err) || meta.IsNoMatchError(err) {
+			return "false", "", nil
+		}
+		return "", "", fmt.Errorf("failed to list subscriptions: %w", err)
+	}
+
+	for _, sub := range subList.Items {
```
thibaultmg, line 79: What happens if several subscriptions coexist at the same time? Like during a migration to a new namespace while one is terminating and the other starting. It will loop through them in random order. It's a bit of an edge case. But we may check the subscription status maybe and ignore inactive ones. And if there are more than one active, I would set it is externally managed in the claim.

operators/endpointmetrics/controllers/mcoa/coo_status.go
```diff
@@ -0,0 +1,104 @@
...
+	managedByExternal   = "external"
+
+	CooInstalledClaimName = "coo-installed.observability.open-cluster-management.io"
+	CooManagedByClaimName = "coo-managed-by.observability.open-cluster-management.io"
+)
+
+// WriteCOOStatus checks whether a COO Subscription exists on the spoke and
+// writes the result as ClusterClaims. The OCM registration agent automatically
+// syncs these to ManagedCluster.Status.ClusterClaims on the hub.
+func WriteCOOStatus(ctx context.Context, c client.Client, log logr.Logger) error {
```
thibaultmg: I would make this function a method of the MCOAAgentReconciler. This would follow the pattern of other functions. And you could encapsulate all the logic like `if r.olmAvailable` in it, simplifying the Reconcile function.

thibaultmg (changes requested):
Awesome that it works with claims! I've left a few comments.

Another point: I feel we could use a single claim instead of two, reducing the load when there are many managed clusters. We could keep only `coo-managed-by.observability.open-cluster-management.io` and add a third value like `none` when it is not installed. WDYT? This would also simplify the code.

coleenquadros:
> Another point: I feel we could use a single claim instead of two, reducing the load when there are many managed clusters. We could keep only `coo-managed-by.observability.open-cluster-management.io` and add a third value like `none` when it is not installed. WDYT? This would also simplify the code.


spec.value can only hold a single string and then that would mean additional parsing work
```
           spec:
              description: Spec defines the attributes of the ClusterClaim.
              type: object
              properties:
                value:
                  description: value is a claim-dependent string
                  type: string
                  maxLength: 1024
                  minLength: 1
```

operators/endpointmetrics/controllers/mcoa/mcoa_agent_controller.go
coleenquadros: I intentionally avoided the update events. I don't think we should allow ownership override for COO. We make the decision based on if MCOA installs or an admin installs and leave the ownership as is. Override could also mean they end up taking up the Monitoring UI  plugin resource which we rely on if incident detection is enabled. Then that would make things complicated.

thibaultmg:
> > Another point: I feel we could use a single claim instead of two, reducing the load when there are many managed clusters. We could keep only `coo-managed-by.observability.open-cluster-management.io` and add a third value like `none` when it is not installed. WDYT? This would also simplify the code.
> 
> spec.value can only hold a single string and then that would mean additional parsing work
> 
> ```
>            spec:
>               description: Spec defines the attributes of the ClusterClaim.
>               type: object
>               properties:
>                 value:
>                   description: value is a claim-dependent string
>                   type: string
>                   maxLength: 1024
>                   minLength: 1
> ```

You would have the value "none", "mcoa" or "external". It's a single one each time.

coleenquadros:
> You would have the value "none", "mcoa" or "external". It's a single one each time.

Ahh i misunderstood. I thought you meant combine the claims to one. I refactored to one claim

operators/endpointmetrics/controllers/mcoa/coo_status.go
```diff
@@ -0,0 +1,98 @@
...
+	CooStatusNotInstalled = "not-installed"
+	CooStatusMCOA         = "mcoa"
+	CooStatusExternal     = "external"
+)
+
+// WriteCOOStatus checks whether a COO Subscription exists on the spoke and
+// writes the result as a single ClusterClaim. The OCM registration agent
+// automatically syncs it to ManagedCluster.Status.ClusterClaims on the hub.
+// On non-OCP clusters (OLMAvailable=false), this is a no-op.
+func (r *MCOAAgentReconciler) WriteCOOStatus(ctx context.Context) error {
```
thibaultmg, line 37: We need to make sure this runs at least once on startup. Right now it depends on getting a `CooStatusClaimName` request in the Reconcile function. While we don't get it, we won't generate the claim. 
We can either call this function from `runMCOA` in main as we call `mcoa.DeployCRDs(crdCtx, cl)` there.
Or better, we can call it from `SetupWithManager`, generating a synthetic initial event with the controller runtime source channel that supports this kind of use case:

```go
            initCh := make(chan event.GenericEvent, 1)
            initCh <- event.GenericEvent{
                Object: &clusterv1alpha1.ClusterClaim{
                    ObjectMeta: metav1.ObjectMeta{Name: CooStatusClaimName},
                },
            }
    
            b = b.WatchesRawSource(
                source.Channel(initCh, &handler.EnqueueRequestForObject{}),
            )
```

thibaultmg (changes requested):
Looks almost good, thanks for the updates! Just left one last comment.

thibaultmg (approved):
/lgtm

